PR #1091, opened 2023-12-18
Add the LICENSES directory for install and use the corrected name in …

jensschmeink:
…the about dailog

## Checklist:

/label ~InProgress


rainman110 (changes requested):
As much as I'd like to approve, I don't think that installing unconditionally to /LICENSES is a good idea.

This is a non-standard director, in particular on Linux. Assume, you install GTlab to /usr, then all GTlab licenses are installed to /usr/LICENSES. This is definitely not the right place. We already use non-standard directory layouts but we should not make it worse.

Instead, I propose to install to `share/gtlab/LICENSES` or a similar location.

rainman110:
Oh no! Please don't differentiate between linux and windows. Also never install to a absolute location (/usr).

Just install to CMAKE_INSTALL_PREFIX/share/gtlab/licenses, i.e. 

```
install(DIRECTORY LICENSES/ share/gtlab/licenses)
```

You just have done so much work 😞 . I am sorry

src/app/dialogs/gt_aboutdialog.cpp
```diff
@@ -371,30 +371,24 @@ GtAboutDialog::openModuleDetails(const QModelIndex& indexOfModule) const
...
-    QDir appLicenseDir(appDir.absolutePath() + QDir::separator() + "Licenses");
+    QString licenseDirPath = gtApp->licenseFolder();
 
-    if (appLicenseDir.exists())
+    if (licenseDirPath.isEmpty())
     {
-        QDesktopServices::openUrl(appLicenseDir.absolutePath());
+        gtError() << tr("No licenses information could be found.");
+        gtError() << tr("Your installation might be broken.");
+        gtError() << tr("Please contact the developers for further information.");
```
jensschmeink, line 380:
```suggestion
```

rainman110 (approved):
Nice work :)

PR #1134, opened 2024-02-22
Enable console application to use collections

jensschmeink:
## Description
The console application now uses the GtApplication and not only the GtCoreApplication. This was required as the collections had been read from the interfaces in the mdiLauncher which is not accessible in the CoreApplication.
This is a work around as the core application should normally depend on GUI elements

## How Has This Been Tested?
The colleague who asked for the change used it successfully.

## Checklist:

### DIFF
--- a/src/batch/CMakeLists.txt
+++ b/src/batch/CMakeLists.txt
@@ -36,6 +36,7 @@ target_link_libraries(GTlabConsole
     PRIVATE
     GTlabCore
     GTlabDataProcessor
+    GTlabGui
     Qt5::Widgets
     Qt5::Xml
 )
--- a/src/batch/batch.cpp
+++ b/src/batch/batch.cpp
@@ -21,6 +21,8 @@
 #include "internal/gt_commandlinefunctionhandler.h"
 #include "batchremote.h"
 
+#include "gt_application.h"
+
 #include "gt_coreapplication.h"
 #include "gt_coredatamodel.h"
 #include "gt_coreprocessexecutor.h"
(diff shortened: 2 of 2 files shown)


jensschmeink:
There will be some registration functions for widgets which might be started. Therefore the starting process might be a bit longer. But JHä did not report about any other problematic issues.

real-ct-ac:
Hast anyone tested it on Linux?

rainman110:
> Hast anyone tested it on Linux?

I did not. But good point. I suppose, we anyhow need an x server etc (we already did before)

jensschmeink:
> Hast anyone tested it on Linux?

Me neither.

jensschmeink:
It needs to be discussed if this temporary fix should be part of GTlab 2-0-7

OliverKunc:
We would appreciate it because we need this feature to run our tests

jensschmeink:
@OliverKunc do you use it on linux? We could give you the executable of this PR and you could test the specific usecase. That would be perfect

OliverKunc:
@jensschmeink We don't run the tests on Linux yet but we hope to do so soon. Executable would be great.

jensschmeink:
@rainman110 @real-ct-ac initial tests on linux did not show any problems (e.g. no appearing widgets). But the logging should be cleaned up as there is some GUI specific logging:  
"### Registered executor 'ProcessRunner'"  
"No shortcuts registered!"  
"No shortcut registered for openProject"  
...

jensschmeink:
@rainman110 as we discussed this PR should be merged. An approval is needed

rainman110 (approved):

PR #1153, opened 2024-03-27
Add type information to property container elements as tooltip

jensschmeink:
## Checklist:

### DIFF
--- a/src/gui/dock_widgets/properties/gt_propertymodel.cpp
+++ b/src/gui/dock_widgets/properties/gt_propertymodel.cpp
@@ -84,13 +84,28 @@ GtPropertyModel::data(const QModelIndex& index, int role) const
         return QVariant();
     }
 
-    if (item->data(index.column(), GtPropertyModel::ContainerRole).toBool() &&
-            role == Qt::DisplayRole)
+    if (item->data(index.column(), GtPropertyModel::ContainerRole).toBool())
     {
-        return item->data(index.column(), role).toString() + + " [" +
-                QString::number(index.row()) + "]";
-    }
+        if (role == Qt::DisplayRole)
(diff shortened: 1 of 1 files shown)


jensschmeink:
This is a first solution for the problem. But as the tool tip might not be seen by the user a more convenient solution might be added in 2-1.
Currently the model view structure of 2-0-6 here would have to be manipulated for another solution (Maybe by a delegate ...)

jensschmeink:
@rainman110 Do you understand why the pipeline fails?

rainman110:
> @rainman110 Do you understand why the pipeline fails?

Please rebase on current master.

jensschmeink:
@rainman110 please check for approval

src/gui/dock_widgets/properties/gt_propertymodel.cpp
```diff
@@ -84,13 +84,27 @@ GtPropertyModel::data(const QModelIndex& index, int role) const
...
-        return item->data(index.column(), role).toString() + + " [" +
-                QString::number(index.row()) + "]";
-    }
+        if (role == Qt::DisplayRole)
+        {
+            return item->data(index.column(), role).toString() +
+                    + " [" + QString::number(index.row()) + "]";
+        }
 
+        if (role == Qt::ToolTipRole)
```
rainman110, line 95: Can you please change the code to avoid the deep nesting? There are many if statements nested, which makes the code hard to read. Use early returns, if possible.

rainman110 (changes requested):

jensschmeink:
@rainman110 i reduced the nested loops

rainman110 (approved):
Much better, thanks :)

PR #1159, opened 2024-04-04
implementation to add object to list of linkable objects

jensschmeink:
For the usage of calculators in the graph system (outside the original process management) it is necessary to fill the list of linkable objects to be able to work with object links.
This is done by an additional function which enables to add objects to this list

## How Has This Been Tested?
Usage in the graph system 

## Checklist:


rainman110 (approved):
Looks good to me :)

PR #1167, opened 2024-04-09
Activate warning for unknown classes only in dev mode

jensschmeink:
## Description
The widget entries for unknown classes is only shown in dev mode

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Fixed unicode characters not correctly saved to tasks, in particular python tasks and file properties - #1125, #45
 
+### Changed
+- Warning for unknown classes is only triggers Error dialog in developer mode of GTlab - #1160
 
 ## [2.0.6] - 2023-11-07 
 ### Fixed
--- a/src/core/gt_projectanalyzer.cpp
+++ b/src/core/gt_projectanalyzer.cpp
@@ -9,6 +9,7 @@
  *  Tel.: +49 2203 601 2907
  */
 
+#include "gt_coreapplication.h"
 #include "gt_footprint.h"
 #include "gt_objectmemento.h"
 
@@ -55,7 +56,7 @@ GtProjectAnalyzer::hasIrregularities()
         return true;
     }
 
-    if (!m_pimpl->m_unknownClasses.isEmpty())
+    if (!m_pimpl->m_unknownClasses.isEmpty() && gtApp->devMode())
     {
         return true;
     }


jensschmeink:
Could be reviewed. Is a simple change to start show the dialog for unknown classes only in dev mode

rainman110:
@jensschmeink How can I even show the unkown classes dialog (before this change). I tried to open a project using predesign and removed  the predesign module first. Then I don't get this dialog but only the "unknown module" dialog.

jensschmeink:
> @jensschmeink How can I even show the unkown classes dialog (before this change)? I tried to open a project using predesign and removed the predesign module first. Then I don't get this dialog but only the "unknown module" dialog.

It is the same dialog but another page

rainman110:
> > @jensschmeink How can I even show the unkown classes dialog (before this change)? I tried to open a project using predesign and removed the predesign module first. Then I don't get this dialog but only the "unknown module" dialog.
> 
> It is the same dialog but another page

I just have the page unknown modules:
image

jensschmeink:
> > > @jensschmeink How can I even show the unkown classes dialog (before this change)? I tried to open a project using predesign and removed the predesign module first. Then I don't get this dialog but only the "unknown module" dialog.
> > 
> > 
> > It is the same dialog but another page
> 
> I just have the page unknown modules: image

I think a module package is not shown if it is unknown and so non of your predesign objects are tried to be constructed. To reproduce add a predesign calculator to one process.

rainman110:
Okay, I could reproduce it.

rainman110 (changes requested):
Please add to changelog

rainman110:
@jensschmeink With your code changes, I still get the error dialog ?! I am not in dev mode.

image

jensschmeink:
> @jensschmeink With your code changes, I still get the error dialog ?! I am not in dev mode.
> 
> image

Yes, you get it as long there is also found an unknown module. If the unknown classes would be the only reason to show the warning it is not shown anymore

rainman110:
I think the problem is, that the popup opens up, because it also has unknown modules. What are the cases, when you have all modules but still unknwn classes?

jensschmeink:
> I think the problem is, that the popup opens up, because it also has unknown modules. What are the cases, when you have all modules but still unknwn classes?

If you have unkown modules but you save your project. Then the usage of the module is no longer registered in the project and there is no comparison of the footprints failing. But the classes use din the project are till unknown

rainman110:
> If you have unkown modules but you save your project.

rainman110 (approved):

jensschmeink:
> > If you have unkown modules but you save your project.
> 
> Okay, this sounds dangerous to me. Just by saving, I loose the connection to the module? Should we fix this (in a separate issue)? Or is this intentional?

I'm not completely sure about that behaviour, so we should check it, discuss it and repair it in another issue

PR #1168, opened 2024-04-09
Move multiple elements in the process dock widget

jensschmeink:
Multiple elements of the same parent can be moved (to another task or a different position in the same task)

## Description
the process view re implemented the drop event which emits a signal. This is connected to a slot in the process dock.

There the list of selected items and the target are checked for validity and the move is performed

## How Has This Been Tested?
By usage in the GUI 

## Checklist:

### DIFF
--- a/src/app/dock_widgets/process/gt_processdock.cpp
+++ b/src/app/dock_widgets/process/gt_processdock.cpp
@@ -1433,7 +1433,7 @@ GtProcessDock::moveElements(const QList<QModelIndex>& source,
 
             if (parent != p->parentObject())
             {
-                gtWarning() << tr("It is only allowed to move elements of the"
+                gtWarning() << tr("It is only allowed to move elements of the "
                                   "same task");
                 return;
             }
--- a/src/app/dock_widgets/process/gt_processdock.cpp
+++ b/src/app/dock_widgets/process/gt_processdock.cpp
@@ -190,6 +190,8 @@ GtProcessDock::GtProcessDock() :
             SLOT(skipComponent(QModelIndex,bool)));
     connect(m_view, SIGNAL(renameProcessElement(QModelIndex)),
             SLOT(renameElement()));
+    connect(m_view, SIGNAL(moveProcessElements(QList<QModelIndex>, QModelIndex)),
+            SLOT(moveElements(QList<QModelIndex>, QModelIndex)));
 
     connect(m_runButton, SIGNAL(clicked(bool)), SLOT(runProcess()));
     connect(m_addElementButton, SIGNAL(clicked(bool)), SLOT(addElement()));
@@ -1397,6 +1399,93 @@ GtProcessDock::renameElement()
     {
         m_view->edit(m_view->currentIndex());
     }
+}
+
+void
+GtProcessDock::moveElements(const QList<QModelIndex>& source,
+                            const QModelIndex& target)
+{
+    if (!target.isValid()) return;
+
+    if (source.isEmpty()) return;
+
+    QList<QModelIndex> mapped;
+
+    for (auto i : source)
+    {
+        if (i.isValid()) mapped.append(mapToSource(i));
+    }
+
+    if (mapped.isEmpty()) return;
+
+    QList<GtObject*> objToMove;
+
+    /// check if all elements to move have the same parent
+    GtObject* parent = nullptr;
+
+    for (auto j : mapped)
+    {
+        if (j.model() != gtDataModel) return;
+
+        if (auto p = gtDataModel->objectFromIndex(j))
+        {
+            if (!parent) parent = p->parentObject();
+
+            if (parent != p->parentObject())
+            {
+                gtWarning() << tr("It is only allowed to move elements of the"
+                                  "same task");
+                return;
+            }
+
+            objToMove.append(p);
+        }
+    }
+
+    if (objToMove.isEmpty()) return;
+
+    QModelIndex mappedTarget = mapToSource(target);
+
+    GtObject* targetObject = gtDataModel->objectFromIndex(mappedTarget);
+
+    auto targetComp = qobject_cast<GtProcessComponent*>(targetObject);
+
+    if (!targetComp) return;
+
+    auto moveCmd = gtApp->makeCommand(gtApp->currentProject(),
+                                       tr("move process element"));
+
+    if (auto taskParent = qobject_cast<GtTask*>(targetComp))
+    {
+        for (auto o: objToMove)
+        {
+            gtDataModel->appendChild(o, taskParent);
+        }
+    }
+    else
+    {
+        GtObject* targetparent = targetComp->parentObject();
+
+        if (auto tp = qobject_cast<GtTask*>(targetparent))
+        {
+            /// to keep the order the swap is neede if the new parent is
+            ///  not the current parent
+            if (objToMove.first()->parentObject() != tp)
+            {
+                std::reverse(objToMove.begin(), objToMove.end());
+            }
+
+            for (auto o: objToMove)
+            {
+                /// if parent is not reset before the insert function
+                /// does not work
+                o->setParent(nullptr);
+                gtDataModel->insertChild(o, tp, mappedTarget.row());
+            }
+        }
+    }
+
+    moveCmd.finalize();
 }
 
 void
--- a/src/app/dock_widgets/process/gt_processdock.cpp
+++ b/src/app/dock_widgets/process/gt_processdock.cpp
@@ -1405,9 +1405,7 @@ void
 GtProcessDock::moveElements(const QList<QModelIndex>& source,
                             const QModelIndex& target)
 {
-    if (!target.isValid()) return;
-
-    if (source.isEmpty()) return;
+    if (!target.isValid() || source.isEmpty()) return;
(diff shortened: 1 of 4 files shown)


src/app/dock_widgets/process/gt_processview.cpp
```diff
@@ -11,6 +11,7 @@
 #include <QScrollBar>
 #include <QKeyEvent>
 
+#include "gt_logging.h"
```
jensschmeink, line 14:
```suggestion
```

src/app/dock_widgets/process/gt_processdock.cpp
```diff
@@ -1399,6 +1401,93 @@ GtProcessDock::renameElement()
...
+    {
+        if (j.model() != gtDataModel) return;
+
+        if (auto p = gtDataModel->objectFromIndex(j))
+        {
+            if (!parent) parent = p->parentObject();
+
+            if (parent != p->parentObject())
+            {
+                gtWarning() << tr("It is only allowed to move elements of the"
```
mariusalexander, line 1436: Missing space
```suggestion
                gtWarning() << tr("It is only allowed to move elements of the "
```

mariusalexander:
When I drag and drop a task that is already a root task to an empty space in the process view I get the error `FATAL [10:49:49] [Gui] move rows failed!` I can't rember ever seeing that before. Also when I drag tw subtasks and want to make them root taks only one is moved. I would propse that either both are moved (and thus both are made root tasks) or none is moved

src/app/dock_widgets/process/gt_processdock.cpp
```diff
@@ -1399,6 +1401,93 @@ GtProcessDock::renameElement()
     }
 }
 
+void
+GtProcessDock::moveElements(const QList<QModelIndex>& source,
+                            const QModelIndex& target)
+{
+    if (!target.isValid()) return;
+
+    if (source.isEmpty()) return;
```
mariusalexander, line 1410: I would prefer merging such simple conditions
```suggestion
    if (!target.isValid() || source.isEmpty()) return;
```

jensschmeink:
> When I drag and drop a task that is already a root task to an empty space in the process view I get the error `FATAL [10:49:49] [Gui] move rows failed!` I can't rember ever seeing that before. Also when I drag two subtasks and want to make them root tasks only one is moved. I would propose that either both are moved (and thus both are made root tasks) or none is moved

The first point is a bit confusing me as I have not touched the function which calls that warning.

The second point sound a good request. I'll add that

src/app/dock_widgets/process/gt_processdock.cpp
```diff
@@ -1399,6 +1401,91 @@ GtProcessDock::renameElement()
...
 
+void
+GtProcessDock::moveElements(const QList<QModelIndex>& source,
+                            const QModelIndex& target)
+{
+    if (!target.isValid() || source.isEmpty()) return;
+
+    QList<QModelIndex> mapped;
+
+    for (auto i : source)
```
mariusalexander, line 1412:
```suggestion
    for (QModelIndex i : source)
```

src/app/dock_widgets/process/gt_processdock.cpp
```diff
@@ -1399,6 +1401,91 @@ GtProcessDock::renameElement()
...
+    }
+
+    if (mapped.isEmpty()) return;
+
+    QList<GtObject*> objToMove;
+
+    /// check if all elements to move have the same parent
+    GtObject* parent = nullptr;
+
+    for (auto j : mapped)
```
mariusalexander, line 1424:
```suggestion
    for (QModelndex j : qAsConst(mapped))
```

Is there a specific reason you have used `j` here? Maybe use `idx, but I dont want to be to nitpicky.

src/app/dock_widgets/process/gt_processdock.cpp
```diff
@@ -1399,6 +1401,91 @@ GtProcessDock::renameElement()
...
+    {
+        if (j.model() != gtDataModel) return;
```
mariusalexander, line 1426: Maybe an assert is a better fit? There should be no reason that the model is not the same, or could it be?
```suggestion
        assert (j.model() == gtDataModel);
```

src/app/dock_widgets/process/gt_processdock.cpp
```diff
@@ -1399,6 +1401,91 @@ GtProcessDock::renameElement()
...
+
+        if (auto p = gtDataModel->objectFromIndex(j))
```
mariusalexander, line 1428: Consider using `obj` or `comp` here, because `p` implies to me parent, but then you access the parent of `p`. Or is `p` referring to `process` here?

```suggestion
        if (auto* p = gtDataModel->objectFromIndex(j))
```

src/app/dock_widgets/process/gt_processdock.cpp
```diff
@@ -1399,6 +1401,91 @@ GtProcessDock::renameElement()
...
+    QList<QModelIndex> mapped;
+
+    for (auto i : source)
+    {
+        if (i.isValid()) mapped.append(mapToSource(i));
+    }
+
+    if (mapped.isEmpty()) return;
+
+    QList<GtObject*> objToMove;
```
mariusalexander, line 1419:
```suggestion
    QList<GtObject*> objsToMove;
```
OR
```suggestion
    QList<GtObject*> objectsToMove;
```

src/app/dock_widgets/process/gt_processdock.cpp
```diff
@@ -1399,6 +1401,91 @@ GtProcessDock::renameElement()
...
+            {
+                /// if parent is not reset before the insert function
+                /// does not work
+                o->setParent(nullptr);
+                gtDataModel->insertChild(o, tp, mappedTarget.row());
+            }
+        }
+    }
+
+    moveCmd.finalize();
```
mariusalexander, line 1486: You dont have to call `finalize` explicitly. It does that in the destructor. It is only useful if you want to finalize the command early
```suggestion
```

src/app/dock_widgets/process/gt_processdock.cpp
```diff
@@ -1399,6 +1401,91 @@ GtProcessDock::renameElement()
...
+
+    if (auto taskParent = qobject_cast<GtTask*>(targetComp))
+    {
+        for (auto o: objToMove)
+        {
+            gtDataModel->appendChild(o, taskParent);
+        }
+    }
+    else
+    {
```
mariusalexander, line 1464: Since you dont need to call `moveCmd.finalize()` you can exit early here and do not require an else case
```suggestion
        return;
    }
```

src/app/dock_widgets/process/gt_processdock.cpp
```diff
@@ -1399,6 +1401,91 @@ GtProcessDock::renameElement()
...
+                std::reverse(objToMove.begin(), objToMove.end());
+            }
+
+            for (auto o: objToMove)
+            {
+                /// if parent is not reset before the insert function
+                /// does not work
+                o->setParent(nullptr);
+                gtDataModel->insertChild(o, tp, mappedTarget.row());
+            }
```
mariusalexander, line 1482: What if object could not be inserted? Then we have a memory leak. ```suggestion
            for (auto o : objToMove)
            {
                /// if parent is not reset before the insert function
                /// does not work
                o->setParent(nullptr);
                if (!gtDataModel->insertChild(o, tp, mappedTarget.row()).isValid())
                {
                    // error handling
                }
            }
```

mariusalexander:
I am now finished with my review. Just minor things

jensschmeink:
> I am now finished with my review. Just minor things

It is part of a different function so I would not like to have it a part of this issue-

mariusalexander:
> > I am now finished with my review. Just minor things
> 
> It is part of a different function so I would not like to have it a part of this issue-

Opened issue #1173 for this purpose

jensschmeink:
@mariusalexander please check again after I added the changes as suggested in your review

mariusalexander:
> > When I drag and drop a task that is already a root task to an empty space in the process view I get the error `FATAL [10:49:49] [Gui] move rows failed!` I can't rember ever seeing that before. Also when I drag two subtasks and want to make them root tasks only one is moved. I would propose that either both are moved (and thus both are made root tasks) or none is moved
> 
> The first point is a bit confusing me as I have not touched the function which calls that warning.
> 
> The second point sound a good request. I'll add that

Have you implemented the second suggestion, to either make all tasks root tasks or none? I tested it locally and the behavior has not changed.

jensschmeink:
> > > When I drag and drop a task that is already a root task to an empty space in the process view I get the error `FATAL [10:49:49] [Gui] move rows failed!` I can't rember ever seeing that before. Also when I drag two subtasks and want to make them root tasks only one is moved. I would propose that either both are moved (and thus both are made root tasks) or none is moved
> > 
> > 
> > The first point is a bit confusing me as I have not touched the function which calls that warning.
> > The second point sound a good request. I'll add that
> 
> Have you implemented the second suggestion, to either make all tasks root tasks or none? I tested it locally and the behavior has not changed.

I added the requested feature. Please check again

mariusalexander (approved):

PR #1169, opened 2024-04-09
Fixed encoding of tasks

### LINKED ISSUE #1125: Special characters within PythonTask scripts
If there are special charakters as umlauts or "°" within python-tasks, they are overwritten with "?" or  "�" after project re-opening. The problems occures during the script reading (GtStringProperty).

image

rainman110:
## Description

GTlab assumes a UTF-8 encoding. However, tasks have been saved using the default ANSI encoding. Thus, when reading in a task, the unicode characters were misinterpreted.

Now, tasks and other xml files are forced to be written as UTF-8.

Fixes #1125

## How Has This Been Tested?

Manually using the following steps:
 - Create a python tasks
 - Insert `print("°")`
 - Save & close the project 
 - Reopen the project and open the task

The `°` character should be still the same and not replace by `?`.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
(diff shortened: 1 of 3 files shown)


jensschmeink:
I can reproduce your nice results

jensschmeink (approved):

PR #1170, opened 2024-04-09
Fixed error "Empty folder if" on docker

rainman110:
## Description

This fixed the aforementioned error message. The reason is, that the username is determined
to create a task group folder. On docker, the username is not set and thus the username is empty,
thus an "empty" folder cannot be created.

This PR simply uses an "undefined" folder, whenever the username could not be determined.

## How Has This Been Tested?

On linux, I simply set the `USER` and the `USERNAME` environment variable to an empty string.

## Checklist:


real-ct-ac (approved):
nice!

rainman110:
The GUI tests are failing because of the changes from master, that changed the context menus when inserting a task.

PR #1177, opened 2024-04-12
Implemented proper style switching and memory management

### LINKED ISSUE #1174: Switching from themes does not update the style correctly
### Summary

On windows I have noticed that when switching from dark to bright mode the style is still the "Fusion" style instead of the expected "windowsvista" style. A restart fixes this

### Expected Behaviour

The styles should update correctly.

### Reproduction steps

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Screenshots

From dark to bright (on windows!) 

grafik

### Logs

_No response_

### GTlab environment

```bash
GTlab Version: 2.0.6-dev
```

### OS

Windows

rainman110:
## Description

The original implementation (GTlab 2.0.6) created a new GtStyle Object, whenever the style has been applied to a widget. All these objected leaked memory.

The first fix was, to simply create a singleton style and apply this to the Widgets. This was not sufficient though, as the style can change and style switching was not correct anymore.

This PR adds a StyleManager under the hood that creates and stores styles on the fly and returns references to the current style.

Closes #1174

## How Has This Been Tested?

In the GUI obviously. Switching styles now works without problems, memory should also not leak.

## Checklist:

### DIFF
--- a/src/gui/style/gt_palette.cpp
+++ b/src/gui/style/gt_palette.cpp
@@ -145,15 +145,6 @@ gt::gui::standardTheme()
     return p;
 }
 
-namespace
-{
-    GtStyle& getStyle()
-    {
-        static GtStyle s;
-        return s;
-    }
-}
-
 
 template <typename Widget>
 void setStyle(Widget& w);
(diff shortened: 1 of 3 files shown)


jensschmeink (approved):
Nice work

PR #1179, opened 2024-04-15
Make tasks groups executable from console

jensschmeink:
If a task is organized in a task group which might not be the default group (normally the users group) the task cannot be executed in the console application.

This is fixed in this PR.
Call the run function now as following:
GTlabConsole run <PROJECT_ID> <TASK_ID> <TASK_GROUP_ID> (last one is optional)

## Description
If the given <TASK_GROUP_ID> is not empty the task group is switched to the given group and the task is searched and if found executed. 
For a cleanup of the batch the functions related to process execution are extracted to new files in a new namespace.

## How Has This Been Tested?
Manual tests of execution of a task which is part of another task group

## Checklist:

### DIFF
--- a/src/batch/batch.cpp
+++ b/src/batch/batch.cpp
@@ -24,10 +24,7 @@
 
 #include "gt_coreapplication.h"
 #include "gt_coredatamodel.h"
-#include "gt_coreprocessexecutor.h"
 #include "gt_project.h"
-#include "gt_projectprovider.h"
-#include "gt_task.h"
 #include "gt_footprint.h"
 #include "gt_utilities.h"
 #include "gt_consoleparser.h"
@@ -64,20 +61,20 @@ checkMetaInput(const QStringList& args)
 
     if (p.positionalArguments().size() != 1)
     {
-        qWarning() << QStringLiteral("ERROR: ")
-                   << QObject::tr("Invalid size of arguments for check meta!"
-                                  "Exactly one argument is required");
+        gtWarning() << QStringLiteral("ERROR: ")
+                    << QObject::tr("Invalid size of arguments for check meta!"
+                                   "Exactly one argument is required");
         return -1;
     }
 
     QString fileName = p.positionalArguments().constFirst();
 
-    qDebug() << "meta input check...";
+    gtDebug() << QObject::tr("meta input check...");
 
     if (fileName.isEmpty())
     {
-        qWarning() << QStringLiteral("ERROR: ")
-                   << QObject::tr("file name is empty!");
+        gtWarning() << QStringLiteral("ERROR: ")
+                    << QObject::tr("file name is empty!");
         return -1;
     }
 
@@ -85,12 +82,12 @@ checkMetaInput(const QStringList& args)
                                GtCoreApplication::version().toString(),
                                false, false))
     {
-        qWarning() << QStringLiteral("ERROR: ")
-                   << QObject::tr("meta input invalid!");
+        gtWarning() << QStringLiteral("ERROR: ")
+                    << QObject::tr("meta input invalid!");
         return -1;
     }
 
-    qDebug() << "meta input ok!";
+    gtDebug() << QObject::tr("meta input ok!");
     return 0;
 }
 
@@ -102,8 +99,8 @@ runMetaInput(const QStringList& args)
 
     if (p.positionalArguments().size() < 2)
     {
-        qWarning() << "Invalid arguments for runMetaInput.";
-        qWarning() << "Two arguments needed as <Input> <Output>";
+        gtWarning() << QObject::tr("Invalid arguments for runMetaInput.");
+        gtWarning() << QObject::tr("Two arguments needed as <Input> <Output>");
 
         return -1;
     }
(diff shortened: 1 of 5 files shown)


jensschmeink:
Most changed lines are the results of the switch from qDebug to gtLogging and the move of the process execution related functions to new files

src/batch/batch.cpp
```diff
@@ -64,33 +61,33 @@ checkMetaInput(const QStringList& args)
 
     if (p.positionalArguments().size() != 1)
     {
-        qWarning() << QStringLiteral("ERROR: ")
-                   << QObject::tr("Invalid size of arguments for check meta!"
-                                  "Exactly one argument is required");
+        gtWarning() << QStringLiteral("ERROR: ")
```
rainman110, line 64: If this is a warning, why do you add "Error"? I think you simply need to remove the first part of the error message.

 Imho, this needs to be an error, as the command is failing!

src/batch/batch.cpp
```diff
@@ -64,33 +61,33 @@ checkMetaInput(const QStringList& args)
...
     QString fileName = p.positionalArguments().constFirst();
 
-    qDebug() << "meta input check...";
+    gtDebug() << QObject::tr("meta input check...");
 
     if (fileName.isEmpty())
     {
-        qWarning() << QStringLiteral("ERROR: ")
-                   << QObject::tr("file name is empty!");
+        gtWarning() << QStringLiteral("ERROR: ")
```
rainman110, line 76: same as above.  Imho, this needs to be an error, as the command is failing!

src/batch/batch.cpp
```diff
@@ -64,33 +61,33 @@ checkMetaInput(const QStringList& args)
...
         return -1;
     }
 
     if (!gt::batch::checkInput(fileName,
                                GtCoreApplication::version().toString(),
                                false, false))
     {
-        qWarning() << QStringLiteral("ERROR: ")
-                   << QObject::tr("meta input invalid!");
+        gtWarning() << QStringLiteral("ERROR: ")
```
rainman110, line 85: same as above.  Imho, this needs to be an error, as the command is failing!

src/batch/batch.cpp
```diff
@@ -112,45 +109,45 @@ runMetaInput(const QStringList& args)
...
+                    << QObject::tr("input file name is empty!");
         return -1;
     }
 
     if (outputFileName.isEmpty())
     {
-        qWarning() << QStringLiteral("ERROR: ") <<
-                   QObject::tr("output file name is empty!");
+        gtWarning() << QStringLiteral("ERROR: ")
```
rainman110, line 123: same as above. Imho, this needs to be an error, as the command is failing!

src/batch/gt_consolerunprocess.cpp
```diff
@@ -16,26 +16,41 @@
...
+
+    for (const auto& o : options())
+    {
+        p.addOption(o.names.first(), o);
+    }
 
     if (!p.parse(args))
     {
-        qWarning() << "Run method without arguments is invalid";
+        gtWarning() << QObject::tr("Run method without arguments is invalid");
```
rainman110, line 54: Why is this a warning?

src/batch/gt_consolerunprocess.cpp
```diff
@@ -132,32 +173,40 @@ gt::console::runProcess(const QString& projectId,
...
+    GtTask* process = nullptr;
+
+    if (taskGroupId.isEmpty())
+    {
+        process = project->findProcess(processId);
+    }
+    else
+    {
+        process = getTask(project, taskGroupId, processId);
+    }
```
rainman110, line 204: Why don't you change getTask in a way. that it can handle an empty taskGroupId instead of needing here an if statement?

src/batch/gt_consolerunprocess.cpp
```diff
@@ -264,23 +314,62 @@ gt::console::runProcessByFile(const QString& projectFile,
...
         }
     }
 
     return 0;
 }
+
+
+
+GtTask*
+gt::console::getTask(GtProject* project, const QString& groupid,
```
rainman110, line 342: Please extend for the case, that the groupId could be empty, such that it will search in the default user folders.

src/batch/gt_consolerunprocess.cpp
```diff
@@ -264,23 +314,62 @@ gt::console::runProcessByFile(const QString& projectFile,
...
+                                                     project->path());
+
+    if (!check)
+    {
+        processData->switchCurrentTaskGroup(groupid, GtTaskGroup::USER,
+                                            project->path());
+    }
+
+    if (!check)
+    {
```
rainman110, line 369: This does not seem to make any sense. Either the variable `check` changes in between, then you need to reassign check to the result of line 363.

Otherwise you can implement everything in one `if(!check)` block.

src/batch/gt_consolerunprocess.cpp
```diff
@@ -264,23 +314,62 @@ gt::console::runProcessByFile(const QString& projectFile,
...
+        gtError() << QStringLiteral("ERROR:")
```
rainman110, line 370: Remove "ERROR: " string

rainman110 (changes requested):
There seems to be some confusion regarding when to use gtError and gtWarning.

IMHO, an error should be logged, when the applications stops and returns a non-zero error code. Please check all occurences, I did not annotate all of the changed lines, as there are many of them.

Also, since gtError etc append [ERROR], you don't need to add it manually.

Finally, the function getTask could be changed, that it also works without a groupId. Then, you can simplify the calling code.

rainman110:
Please append the changelog!

rainman110 (approved):
Thank you, this looks better.

I am not sure, whether it is now actually better to use the logging here instead of plain cerr / cout, when the console application is reporting an error to the user.

But lets postpone this discussion in a separate issue.

The implementation itself is now fine. Thank you!

PR #1190, opened 2024-04-18
Resolve "Add app setting to change logging level at runtime"

### LINKED ISSUE #1086: Add setting to change logging level at runtime
The user should be able to change the logging level - similar to how the verbosity level can be changed on the fly. 

There should also be "Default" option, that uses the logging level as defined by the application

mariusalexander:
## Description

Closes #1086.

Added setting to change logging level at runtime. Will remember selection, however, arguments like `--debug` or `--dev` will override this setting but only for the duration of the session. If the logging level is overridden and the preferences dialog is opened and saved the logging level is changed permanently to the overridden logging level.

grafik

## How Has This Been Tested?

Manually.

## Checklist:

### DIFF
--- a/src/app/dock_widgets/output/gt_outputdock.cpp
+++ b/src/app/dock_widgets/output/gt_outputdock.cpp
@@ -23,11 +23,14 @@
 #include "gt_icons.h"
 #include "gt_statehandler.h"
 #include "gt_state.h"
+#include "gt_settings.h"
 
 #include <QHeaderView>
-#include <QVBoxLayout>
+#include <QGridLayout>
 #include <QHBoxLayout>
+#include <QLabel>
 #include <QTabWidget>
+#include <QComboBox>
 #include <QPushButton>
 #include <QScrollBar>
 #include <QFileDialog>
@@ -42,16 +45,38 @@
 
 #include <algorithm>
 
+struct LoggingLevel
+{
+    gt::log::Level level;
+    QString name;
+};
+
+std::array<LoggingLevel, 4> const s_loggingLevels{
+    LoggingLevel{gt::log::TraceLevel,   QStringLiteral("Trace")},
+    LoggingLevel{gt::log::DebugLevel,   QStringLiteral("Debug")},
+    LoggingLevel{gt::log::InfoLevel,    QStringLiteral("Default")},
+    LoggingLevel{gt::log::WarningLevel, QStringLiteral("Warnings Only")}
+};
+
 GtOutputDock::GtOutputDock()
 {
+    // order verbosity levels depending on their value
+    QStringList loggingLevels;
+    std::transform(std::begin(s_loggingLevels), std::end(s_loggingLevels),
+                   std::back_inserter(loggingLevels), [](auto const& entry){
+        return entry.name;
+    });
+
     setObjectName(tr("Output"));
 
-    QWidget* widget = new QWidget(this);
+    auto* widget = new QWidget(this);
     widget->setObjectName("centralWidget");
     setWidget(widget);
 
-    QVBoxLayout* layout = new QVBoxLayout;
+    auto* layout = new QGridLayout;
     layout->setContentsMargins(0, 0, 0, 0);
+
+    widget->setLayout(layout);
 
     // run model test only in debug mode
 #ifdef QT_DEBUG
@@ -69,6 +94,8 @@ GtOutputDock::GtOutputDock()
 
     QTabWidget* tab = new QTabWidget;
     tab->setObjectName("tabWidget");
+
+    layout->addWidget(tab, 0, 0);
 
     // default page
     QWidget* defaultPage = new QWidget(this);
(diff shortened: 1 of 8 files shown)


jensschmeink:
I like the new function but I have some comments:
- It is a bit missleading that the buttons to display messages in the output dock stay in the widget. I know they can be used to filter old existing messages but it feels a bit strange to see the buttons now even in non-dev mode.
- The names of the logging levels are not so intuitive for me

mariusalexander:
>     * It is a bit missleading that the buttons to display messages in the output dock stay in the widget. I know they can be used to filter old existing messages but it feels a bit strange to see the buttons now even in non-dev mode.

Agree, it might be misleading. The reason I have not hidden these entries is because there is no easy way of updating the visibility of these buttons. For example, say you start GTlab in user mode (so only info and higher is logged), then there will be filter buttons for info, warning and error. If you now change the logging level in the settings to say Trace level, there is (currently) no way of telling the output dockwidget, that the logging level has changes, thus that the trace and debug filter buttons (which should appear) cannot be shown (and vice versa).

I did not want to add a signal, say to the `GtCoreApplication`, but maybe we have to? It is your call

And now that you mention it: What if trace messages are currently in the logging history, but trace message are no longer logged after changing the logging level. What should happen to the trace filter button?


>     * The names of the logging levels are not so intuitive for me

I could add an tooltip. What would you suggest?

rainman110:
Maybe instead of adjusting the logging in the settings, it could be a combo box in the logging / output widget itself.

mariusalexander:
> Maybe instead of adjusting the logging in the settings, it could be a combo box in the logging / output widget itself, e.g.:
> 
> image

Is the screenshot edited or how did you place it there? I like the idea to place, but I cannot imagine how I could achieve to place it there @rainman110. If we place it in the output dock I would like to make it take up less space, since the output dock is already quite noisy. 

> Also, I was wondering, what "user-level" is. So it seems to print the "information" messages. Why not rename it "Addition Info"

"Additional Info" does not sit right with me as well. The Info-Level is the main logging level to be used by the user. Maybe I'll go with `Default` or `Informative`

rainman110:
> Is the screenshot edited or how did you place it there? I like the idea to place, but I cannot imagine how I could achieve to place it there @rainman110. If we place it in the output dock I would like to make it take up less space, since the output dock is already quite noisy.

It was just photoshopped. More like a general idea. It does not need to be exactly there though. I could imagine, that this is more user friendly. I could definitely take less space there.

mariusalexander:
I could do it like this: 
grafik

grafik

rainman110:
Looks good for me 👍

mariusalexander:
Okay wait... Would you look at that, I found a way! @rainman110 

grafik

rainman110:
> Okay wait... Would you look at that, I found a way! @rainman110
> 
> grafik

Wow cool 😄  I like it 🥳

rainman110:
How did you do it?

mariusalexander:
So, obviously you can add a widget to a layout of the "main" widget. The layout will take care of resize events and general placement as far as I know. However you can also simply set the parent of the child widget to the main widget. This will make it appear at (0, 0) which is the top left. Since I needed the widget to be on the far right side I

mariusalexander:
Sorry I miss-clicked and closed the PR 😂

mariusalexander:
Oh I have noticed that I still need to make the buttons update when changing the logging level. What should happen e.g. if I have Trace messages in the output but then I switch the logging level to Debug? Should the trace messages be deleted and the trace filter button be hidden? @rainman110 @jensschmeink

jensschmeink:
> Oh I have noticed that I still need to make the buttons update when changing the logging level. What should happen e.g. if I have Trace messages in the output but then I switch the logging level to Debug? Should the trace messages be deleted and the trace filter button be hidden?

Good question. Is it an option to check the existance of those messages and hide/show the button depending on this?

mariusalexander:
Sure, that is possible too

rainman110:
I would probably just hide the buttons but keep the state of the buttons. Thus, the old messages stay as they are but no new trace messages will appear.

But I also like you idea to decide based on the current history, whether to show the buttons or not.

mariusalexander:
@rainman110 Finished this PR, please review.

I have implemented it so that each time the logmodel deletes a row it is checked if each filter button (only trace, debug and info are affected) are visible and should be hidden depending on the logging level. Only if the buttons are visible and if they should be hidden, it is checked whether the log model contains the corresponding logging level. If not, the button is hidden else it won't.

src/gui/style/gt_palette.cpp
```diff
@@ -58,16 +58,18 @@ generateTheme(ColorConfig const& config)
 
     // tones
     /// QPalette::Dark: Apperantly affects the lower part of a raised frame
+    /// the bottom part of sunken frame
     palette.setColor(QPalette::Dark,
                      gt::gui::color::darken(config.main, 40));
 
-    /// QPalette::Mid:
+    /// QPalette::Mid: ???
```
rainman110, line 65: ???

src/core/gt_logmodel.cpp
```diff
@@ -118,6 +118,15 @@ GtLogModel::setEnabled(bool enable)
         << tr("Logmodel is now enabled, recieving incomming messages!");
 }
 
+bool
+GtLogModel::containsLogLevel(gt::log::Level level) const
+{
+    return std::any_of(m_entries.begin(), m_entries.end(),
```
rainman110, line 124: 💋

src/app/preferences/pages/gt_preferencesapp.h
```diff
@@ -70,7 +70,7 @@ private slots:
      * @brief onAutoSaveTriggered
      * @param val
      */
-    void onAutoSaveTriggered(bool val);
+    [[deprecated]] void onAutoSaveTriggered(bool val);
```
rainman110, line 73: Good catch

src/app/dock_widgets/output/gt_outputdock.cpp
```diff
@@ -124,27 +151,70 @@ GtOutputDock::GtOutputDock()
...
-            searchWidget, &GtSearchWidget::enableSearch);
-
     defaultLayout->addWidget(m_logView);
 
     connect(searchWidget, &GtSearchWidget::textChanged,
             m_model, &GtFilteredLogModel::filterData);
 
     filterLayout->addStretch(1);
 
+    // logging level
```
rainman110, line 161: Could you please extract some functions to reduce the overall length of this constructor? The code is overall fine, but too long to read.

src/app/preferences/pages/gt_preferencesapp.cpp
```diff
@@ -201,19 +135,22 @@ GtPreferencesApp::GtPreferencesApp() :
...
+        auto iter = std::find_if(std::begin(s_verbosityLevels),
+                                 std::end(s_verbosityLevels),
+                                 [=](auto const& entry){
+            return verbosityLevel <= entry.verbosity;
+        });
+
+        if (iter != std::end(s_verbosityLevels))
+        {
+            m_verbositySelection->setCurrentText(iter->name);
+        }
```
rainman110, line 153: Why was this change required? I just see a scope around the old code...

rainman110 (changes requested):
Overall a very good PR. I really like, how it works now and how it is implemented.

Just have minor comments. Please have a look

src/app/preferences/pages/gt_preferencesapp.cpp
mariusalexander: It was changed in a previous commit to separate logging level and verbosity, I will move this into a namespace

mariusalexander:
@rainman110 I have integrated your suggestions. Both the constructor of the app settings page and the output dock are now a lot cleaner 👍🏽 But the diff is now a bit messy... Can you have another look?

rainman110 (approved):
Looks good now. Thank you, great work!

rainman110:
@Please avoid merging, before the CI has passed. It is currently running for you two latest commits.

mariusalexander:
> @mariusalexander Please avoid merging, before the CI has passed. It is currently running for you two latest commits.

I see, sorry 😇

PR #1195, opened 2024-04-23
Reduced scope of the import menu command

### LINKED ISSUE #1188: [CHANGE] - Recude the undo/redo command scope for the import menu
### Summary

It seems to me that the import menu is using a project wide scope for undo/redo command when a object wide scope **may** suffice. This could reduce lag spikes for larger projects. But we a´have to check if this is indeed a mistake in the import menu.

See below

https://github.com/dlr-gtlab/gtlab-core/blob/bca377172d6df52e1ece73d743c7ac3067735f30/src/gui/importer/gt_importmenu.cpp#L123-L141

### Impact on the codebase

Minimal.

### Current workarounds

None.

rainman110:
Closes #1188

## How Has This Been Tested?

 - I first checked, that the command is actually used by commenting it out: Undo was not available, great!
 - Then you changed the code to use the enclosing object, Undo is working again 😄 

## Checklist:

### DIFF
--- a/src/gui/importer/gt_importmenu.cpp
+++ b/src/gui/importer/gt_importmenu.cpp
@@ -135,7 +135,7 @@ GtImportMenu::onActionTrigger(QObject* obj)
                              QStringLiteral(" ") +
                              m_obj->objectName();
 
-            auto command = gtApp->makeCommand(project, msg);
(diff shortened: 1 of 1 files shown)


jensschmeink (approved):
Nice and short

PR #1198, opened 2024-04-24
Resolve "fix GtEventLoop running indefintely if timeout was set to 0"

### LINKED ISSUE #1154: GtEventLoop does not exit if the timeout is set to 0
When using `GtEventLoop` with a timeout of `0` the event loop does not exit at all, it runs indefinitely. This goes against the expected behaviour (IMO) and the doxygen comments also says otherwise.

```cpp
// this should do a single iteration of the event loop but it runs indefinitely
GtEventLoop eventLoop(std::chrono::milliseconds(0)); 
eventLoop.exec();
```

I had a quick look into [./src/dataprocessor/gt_eventloop.cpp#59](https://github.com/dlr-gtlab/gtlab-core/blob/fb3bc88f62556b92c73e31bdde748735eaa0ba80/src/dataprocessor/gt_eventloop.cpp#L59) and I think the problem is the `>` operator. Changing this to `>=` could fix this.

mariusalexander:
## Description

Closes #1154 

Fixed `GtEventLoop` not returning if timeout was set to 0.

## How Has This Been Tested?

- Checked usages of `GtEventLoop`
- Added Unittests

## Checklist:

### DIFF
--- a/src/dataprocessor/gt_eventloop.cpp
+++ b/src/dataprocessor/gt_eventloop.cpp
@@ -56,7 +56,7 @@ GtEventLoop::exec()
 {
     // trigger eventloop and timeout
     QTimer timer;
-    if (m_mstimeout > 0)
+    if (m_mstimeout >= 0)
     {
         timer.setSingleShot(true);
         connectFailed(&timer, &QTimer::timeout);
--- a/tests/unittests/datamodel/test_gt_eventloop.cpp
+++ b/tests/unittests/datamodel/test_gt_eventloop.cpp
@@ -15,7 +15,7 @@
 #include <QTimer>
 
 static const int timeout = 1 * 1000;
-static const int no_timeout = 0 * 1000;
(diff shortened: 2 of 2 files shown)


mariusalexander:
@rainman110 May you have a look? :)

mariusalexander:
>    * What does a timeout of zero even mean?

It should do at least one iteration. The documentation says https://doc.qt.io/qt-6/qtimer.html#interval-prop: (which is exactly what I need/expected)

> A [QTimer](https://doc.qt.io/qt-6/qtimer.html) with a timeout interval of 0 will time out as soon as all the events in the window system's event queue have been processed.

Regarding negative values for the timer. I could not find an answer to that in the documentation. But since the eventloop should run idefinetly we do not need a timer for that, therefore I would propose to keep the if-statement as is.

rainman110 (approved):

PR #1200, opened 2024-04-25
Reset scope for undo_redo commands in process dock

jensschmeink:
The scope for the change commands in the process dock are changed to not use the whole currentProject but 

## Description
The commands to determine the changes to keep in the undo/redo system got a more specific scope: the current selected task group

## How Has This Been Tested?
I startet GTlab with a project with multiple task groups and tasks in it. I did several actions like moving, copy, paste, ... and no problem occured but undo was possible.

## Checklist:

### DIFF
--- a/src/app/dock_widgets/process/gt_processdock.cpp
+++ b/src/app/dock_widgets/process/gt_processdock.cpp
@@ -1456,7 +1456,7 @@ GtProcessDock::moveElements(const QList<QModelIndex>& source,
             }
         }
 
-        auto moveCmd = gtApp->makeCommand(gtApp->currentProject(),
+        auto moveCmd = gtApp->makeCommand(m_taskGroup,
                                           tr("move tasks element"));
         for (auto o : objectsToMove)
         {
@@ -1693,7 +1693,7 @@ GtProcessDock::cutElement(const QModelIndex& index)
     toDelete.append(pComp);
 
     auto command =
-        gtApp->makeCommand(gtApp->currentProject(),
+        gtApp->makeCommand(m_taskGroup,
                            tr("Cut Process Element") +
                            QStringLiteral(" (") + obj->objectName() +
                            QStringLiteral(")"));
@@ -1773,7 +1773,7 @@ GtProcessDock::deleteProcessElements(const QList<QModelIndex>& indexList)
         {
             gtDataModel->reduceToParents(objects);
 
-            auto cmd = gtApp->makeCommand(gtApp->currentProject(),
+            auto cmd = gtApp->makeCommand(m_taskGroup,
                                           tr("Delete Process Elements"));
             Q_UNUSED(cmd)
 
@@ -1863,7 +1863,7 @@ GtProcessDock::pasteElement(GtObject* obj, GtObject* parent)
 
     m_view->setFocus();
 
-    auto pasteCmd = gtApp->makeCommand(gtApp->currentProject(),
+    auto pasteCmd = gtApp->makeCommand(m_taskGroup,
                                        tr("Paste process element"));
 
     auto const propCons = obj->findChildren<GtPropertyConnection*>();
@@ -1943,7 +1943,7 @@ GtProcessDock::skipComponent(const QModelIndex& index, bool skip)
     QString msg = tr("%1 selected Process Elements")
                   .arg(skip ? tr("Skip") : tr("Unskip"));
 
-    auto cmd = gtApp->makeCommand(gtApp->currentProject(), msg);
+    auto cmd = gtApp->makeCommand(m_taskGroup, msg);
     Q_UNUSED(cmd)
 
     skipComponent(componentByModelIndex(index), skip);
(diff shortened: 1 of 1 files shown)


src/app/dock_widgets/process/gt_processdock.cpp
```diff
@@ -1773,7 +1773,7 @@ GtProcessDock::deleteProcessElements(const QList<QModelIndex>& indexList)
         {
             gtDataModel->reduceToParents(objects);
 
-            auto cmd = gtApp->makeCommand(gtApp->currentProject(),
+            auto cmd = gtApp->makeCommand(m_taskGroup,
```
mariusalexander, line 1776: Consider this: You have many many tasks, and somewhere deeply nested you delete a process component. The code still check all other tasks, you could however use the the `find_lowest_ancestor` function to find the "commonParent". It may be the task group or it may be the an object that has a lot less children, thus could perform better still.
```suggestion
            auto* commonParent = gt::find_lowest_ancestor(objects, gt::get_parent_object);
            assert(commonParent);
            auto cmd = gtApp->makeCommand(commonParent,
```

src/app/dock_widgets/process/gt_processdock.cpp
```diff
@@ -1693,7 +1693,7 @@ GtProcessDock::cutElement(const QModelIndex& index)
     toDelete.append(pComp);
 
     auto command =
-        gtApp->makeCommand(gtApp->currentProject(),
+        gtApp->makeCommand(m_taskGroup,
```
mariusalexander, line 1696: Can you narrow it down to the parent of the object/objects that is cut? Maybe use `find_lowest_ancestor` as well (see the other comment)

src/app/dock_widgets/process/gt_processdock.cpp
```diff
@@ -1863,7 +1863,7 @@ GtProcessDock::pasteElement(GtObject* obj, GtObject* parent)
 
     m_view->setFocus();
 
-    auto pasteCmd = gtApp->makeCommand(gtApp->currentProject(),
+    auto pasteCmd = gtApp->makeCommand(m_taskGroup,
```
mariusalexander, line 1866: You may be able to use parent here
```suggestion
    auto pasteCmd = gtApp->makeCommand(parent,
```

src/app/dock_widgets/process/gt_processdock.cpp
```diff
@@ -1943,7 +1943,7 @@ GtProcessDock::skipComponent(const QModelIndex& index, bool skip)
     QString msg = tr("%1 selected Process Elements")
                   .arg(skip ? tr("Skip") : tr("Unskip"));
 
-    auto cmd = gtApp->makeCommand(gtApp->currentProject(), msg);
+    auto cmd = gtApp->makeCommand(m_taskGroup, msg);
     Q_UNUSED(cmd)
 
     skipComponent(componentByModelIndex(index), skip);
```
mariusalexander, line 1949: Can you narrow it down to the object that is skipped? Maybe access `componentByModelIndex` and then use this object for the command

mariusalexander:
Can you have a look at [porcessdock.cpp#L1477](https://github.com/dlr-gtlab/gtlab-core/blob/9cbc3628c158c5c36e1bf7eee95ccabf234cb46d/src/app/dock_widgets/process/gt_processdock.cpp#L1477-L1478): I think you can use the `find_lowest_ancestor` function here as well:
```cpp
auto* commonParent = gt::find_lowest_ancestor(objectsToMove << targetComp, gt::get_parent_object);
assert (commonParent);
auto cmd = gtApp->makeCommand(commonParent,
```

mariusalexander:
@jensschmeink  btw: have you already noticed any improvements for larger projects?

jensschmeink:
> @jensschmeink btw: have you already noticed any improvements for larger projects?

Currently not but my projects are not that huge.

jensschmeink:
@mariusalexander I applied your suggestions and everything still seems to work fine

src/app/dock_widgets/process/gt_processdock.cpp
```diff
@@ -1474,7 +1475,13 @@ GtProcessDock::moveElements(const QList<QModelIndex>& source,
 
     if (!targetComp) return;
 
-    auto moveCmd = gtApp->makeCommand(m_taskGroup,
+
+    auto* commonParent = gt::find_lowest_ancestor(objectsToMove,
```
mariusalexander, line 1479: I think you may need target component here as well (You should be able to use the << operator to create temporary list). This is the object where the tasks are moved to, correct?
```suggestion
    auto* commonParent = gt::find_lowest_ancestor(objectsToMove << targetComp,
```

src/app/dock_widgets/process/gt_processdock.cpp
```diff
@@ -1943,7 +1923,7 @@ GtProcessDock::skipComponent(const QModelIndex& index, bool skip)
     QString msg = tr("%1 selected Process Elements")
                   .arg(skip ? tr("Skip") : tr("Unskip"));
 
-    auto cmd = gtApp->makeCommand(gtApp->currentProject(), msg);
+    auto cmd = gtApp->makeCommand(m_taskGroup, msg);
     Q_UNUSED(cmd)
 
     skipComponent(componentByModelIndex(index), skip);
```
mariusalexander, line 1929: I would suggest to narrow it down to the object that is skipped
```suggestion
    auto* comp = componentByModelIndex(index);
    if (!comp) return;
    
    auto cmd = gtApp->makeCommand(comp, msg);
    Q_UNUSED(cmd);
    
    skipComponent(comp, skip);
```

src/app/dock_widgets/process/gt_processdock.cpp
mariusalexander: May check if this is indeed needed?

mariusalexander:
Rest looks good to me

mariusalexander (approved):

Issue #1199, opened 2024-04-25
[CHANGE] - Refactor GtMdiLauncher and resolve ownership issues of MDI Items and their widgets


mariusalexander:
### Summary

The `GtMdiLauncher` is quite messy IMHO. Both from an API point of view and from looking at the code it is not clear who ownes which objects. 

For example, there are connections that ....
- delete an MdiItem if the widget is closed
- delete the widget if the MdiItem was deleted
and in addition within the MdiItem desctructor the widget is deleted again (if it still exists). 

Maybe we could work with `unqiue_ptr` or the "volatile" pointer from the IntelliGraph module to make this system *and the API*  more clear.

Detaching a widget became quite messy due to these relation ships (see #603)

### Impact on the codebase

Ideally this should both result in a ABI and API change

PR #1210, opened 2024-05-02
2nd Attempt: "Allow to detach MDI Tabs to make them their own standalone widget"

### LINKED ISSUE #604: Allow to detach MDI Tabs to make them their own standalone widget
In GitLab by @mariusalexander on Oct 4, 2023, 15:59

Similar to brwoser tabs they can be moved to a standalone window. This would enhance the user experience quite a bit IMO and would allow for a more productive and less clunky user experience. Obviously when closing GTlab all detached tabs should be closed as well (and maybe the user should be informed that there are open tabs?)

mariusalexander:
## Description

Closes #604.

*See previous PR #1201 for first attempt and more details.* 

I just could not live without this feature so I gave it another shot and I think I finally could make it work. It not the nicest piece of code I have ever written but I hope it is sufficient. With the redesign of the MDI launcher (see #1199) this code * hopefully* does not live long anyways.

The widget can be redocked by minimizing the window (using the window manager). This feels somewhat intuitive, but in the we may want to enhance the UX by allowing drag and drop.

## How Has This Been Tested?

Manually. I verified that the buttons of the tab widget are deleted properly, that the MDI Item and widget are deleted properly and could not found any issue.

There was an bug where GTlab would crash if I tried to dock the undocked widget a second time, this was reslved by using `Qt::QueuedConnection` for the redock slot-function. 

## Screenshots

Before undocking: 

grafik

After undocking:

grafik

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - MDI Items can now be closed using middle mouse click (similarly to most common browsers). - #603
+- Allowed MDI Items to be detached and redocked by minimizing the window. Finally enabling multi-window workflows with GTlab (*Note:* This feature is still WIP). - #604
 
 ### Removed
 - Removed the error message dialog. - #611
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - MDI Items can now be closed using middle mouse click (similarly to most common browsers). - #603
 - Allowed MDI Items to be detached and redocked by minimizing the window. Finally enabling multi-window workflows with GTlab (*Note:* This feature is still WIP). - #604
 
+    *Known Issues:*
+	- Detaching 3D viewers may interfer with other MDI Items and cause strange side effects.
+
 ### Removed
 - Removed the error message dialog. - #611
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Allowed MDI Items to be detached and redocked by minimizing the window. Finally enabling multi-window workflows with GTlab (*Note:* This feature is still WIP). - #604
 
     *Known Issues:*
-	- Detaching 3D viewers may interfer with other MDI Items and cause strange side effects.
+	- Dropping objects into detached MDI windows stops working after opening a 3D PreDesign CAD viewer.
 
 ### Removed
 - Removed the error message dialog. - #611
--- a/src/dataprocessor/gt_utilities.h
+++ b/src/dataprocessor/gt_utilities.h
@@ -157,13 +157,13 @@ public:
         using pointer           = void;
         using reference         = T;
 
-        explicit iterator(T _num) : num(_num) {}
+        constexpr explicit iterator(T _num) noexcept : num(_num) {}
 
-        iterator& operator++() { ++num; return *this; }
-        iterator operator++(int) { iterator tmp = *this; ++(*this); return tmp; }
-        bool operator==(iterator other) const { return num == other.num; }
-        bool operator!=(iterator other) const { return !(*this == other); }
-        reference operator*() const { return num; }
+        constexpr iterator& operator++() noexcept { ++num; return *this; }
+        constexpr iterator operator++(int) noexcept { iterator tmp = *this; ++(*this); return tmp; }
+        constexpr bool operator==(iterator other) const noexcept { return num == other.num; }
+        constexpr bool operator!=(iterator other) const noexcept { return !(*this == other); }
+        constexpr reference operator*() const noexcept { return num; }
 
     private:
 
@@ -173,12 +173,16 @@ public:
     using const_iterator = iterator;
 
     /// constructor
-    explicit DynamicRange(T from, T to) : m_begin(from), m_end(to) { }
+    constexpr explicit DynamicRange(T from, T to) noexcept :
+        m_begin(from), m_end(to)
+    {
+        assert(to >= from || !"only positive ranges are supported!");
+    }
 
-    iterator begin() const { return iterator(m_begin); }
-    iterator end() const { return iterator(m_end); }
+    constexpr iterator begin() const noexcept { return iterator(m_begin); }
+    constexpr iterator end() const noexcept { return iterator(m_end); }
 
-    size_t size() const { return m_end - m_begin; }
+    constexpr size_t size() const noexcept { return m_end - m_begin; }
 
 private:
 
@@ -196,7 +200,7 @@ private:
  * @return Range
  */
 template<typename T>
-inline DynamicRange<T> range(T from, T to)
+constexpr inline DynamicRange<T> range(T from, T to)
 {
     return DynamicRange<T>(from, to);
 }
@@ -206,7 +210,7 @@ inline DynamicRange<T> range(T from, T to)
  * @return Range
  */
 template<typename T>
-inline DynamicRange<T> range(T to)
+constexpr inline DynamicRange<T> range(T to)
 {
     return DynamicRange<T>(T(0), to >= T(0) ? to : T(0));
 }
--- a/src/dataprocessor/gt_utilities.h
+++ b/src/dataprocessor/gt_utilities.h
@@ -135,6 +135,82 @@ T const& clamp(T const& value, T const& low, T const& high)
     return std::max(low, std::min(high, value));
 }
 
+/**
+ * @brief The DynamicRange class. Implements a simple way to iterator from a
+ * a starting number to an end number using iterators and without the need to
+ * create a temporary list.
+ *
+ * Only supports positive ranges.
+ */
+template<typename T>
+class DynamicRange
+{
+public:
+
+    /// iterator class
+    class iterator
+    {
+    public:
+        using iterator_category = std::input_iterator_tag;
+        using difference_type   = T;
+        using value_type        = T;
+        using pointer           = void;
+        using reference         = T;
+
+        explicit iterator(T _num) : num(_num) {}
+
+        iterator& operator++() { ++num; return *this; }
+        iterator operator++(int) { iterator tmp = *this; ++(*this); return tmp; }
+        bool operator==(iterator other) const { return num == other.num; }
+        bool operator!=(iterator other) const { return !(*this == other); }
+        reference operator*() const { return num; }
+
+    private:
+
+        T num = 0;
+    };
+
+    using const_iterator = iterator;
+
+    /// constructor
+    explicit DynamicRange(T from, T to) : m_begin(from), m_end(to) { }
+
+    iterator begin() const { return iterator(m_begin); }
+    iterator end() const { return iterator(m_end); }
+
+    size_t size() const { return m_end - m_begin; }
+
+private:
+
+    T m_begin = 0, m_end = 0;
+};
+
+/**
+ * @brief Creates a dynamic range that can be iterated over using c++ iterators.
+ * Can be used to replace classic for loops with range based for loops
+ *
+ * for (auto idx : range(0, 10)) { ... }
+ *
+ * @param from Begin of range
+ * @param to End of range
+ * @return Range
+ */
+template<typename T>
+inline DynamicRange<T> range(T from, T to)
+{
+    return DynamicRange<T>(from, to);
+}
+/**
+ * @brief Overload. Starts range at 0.
+ * @param to End of range
+ * @return Range
+ */
+template<typename T>
+inline DynamicRange<T> range(T to)
+{
+    return DynamicRange<T>(T(0), to >= T(0) ? to : T(0));
+}
+
 } // namespace gt
 
 #endif // GT_UTILITIES_H
--- a/src/gui/gt_mdiitem.cpp
+++ b/src/gui/gt_mdiitem.cpp
@@ -19,7 +19,6 @@
 #include "gt_objectchangedevent.h"
 #include "gt_statehandler.h"
 
-#include <gt_logging.h>
 
 GtMdiItem::GtMdiItem() :
     m_frame(new GtDockableFrame),
@@ -40,7 +39,7 @@ GtMdiItem::GtMdiItem() :
 QWidget*
 GtMdiItem::widget()
 {
-    return static_cast<QWidget*>(m_frame.data());
+    return m_frame;
 }
 
 QIcon
--- a/src/gui/gt_mdiitem.cpp
+++ b/src/gui/gt_mdiitem.cpp
@@ -19,7 +19,6 @@
 #include "gt_objectchangedevent.h"
 #include "gt_statehandler.h"
 
-
 #include <gt_logging.h>
 
 GtMdiItem::GtMdiItem() :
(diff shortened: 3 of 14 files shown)


src/gui/gt_mdiitem.cpp
```diff
@@ -8,21 +8,24 @@
...
 #include "gt_mdiitem.h"
 #include "gt_application.h"
 #include "gt_datamodel.h"
 #include "gt_projectchangedevent.h"
 #include "gt_objectchangedevent.h"
 #include "gt_statehandler.h"
 
-GtMdiItem::GtMdiItem() : m_frame(new QFrame), m_d(nullptr),
-    m_subWin(nullptr), m_queueEvents(false)
+#include <gt_logging.h>
```
mariusalexander, line 22:
```suggestion
```

src/gui/gt_mdiitem.cpp
```diff
@@ -37,7 +40,7 @@ GtMdiItem::GtMdiItem() : m_frame(new QFrame), m_d(nullptr),
 QWidget*
 GtMdiItem::widget()
 {
-    return m_frame;
+    return static_cast<QWidget*>(m_frame.data());
```
mariusalexander, line 43:
```suggestion
    return m_frame;
```

mariusalexander:
@jensschmeink @rainman110 I would like both of your reviews. I know I moved the issue to 2-1-0, but I could not live without the feature once implemented. I hope this PR can still make it into 2-0-7 even tough it is not a fix but a new feature. Since 2-1-0 will probably takes us quite a while to finish, I would like to have access to this system now (and I think everybody else would agree :joy:).

src/gui/widgets/gt_tabwidget.cpp
```diff
@@ -22,6 +23,47 @@ GtTabWidget::GtTabWidget(QWidget* parent) :
 
 }
 
+int
+GtTabWidget::tabIndex(QWidget const* widget) const
+{
+    if (!widget) return -1;
+
+    int n = count();
+    for (int i = 0; i < n; ++i)
```
jensschmeink, line 32: I think there should be a std algorithm to replace this for loop (maybe find_if)

src/gui/gt_mdilauncher.cpp
```diff
@@ -438,19 +454,88 @@ GtMdiLauncher::open(const QString& id, GtObject* data, const QString& customId)
...
+    QPointer<GtDockableFrame> dockableWidget{qobject_cast<GtDockableFrame*>(wid)};
+    auto* tabWidget = qobject_cast<GtTabWidget*>(m_area.data());
+
+    // helper function to make mdi buttons
+    auto const makeMdiButton = [tabWidget, icon, dockableWidget,
+                                mdiItem = QPointer<GtMdiItem>(mdiItem)](){
+        assert(tabWidget);
+
+        // set custom close button
+        QPushButton* closeBtn = new QPushButton;
```
jensschmeink, line 467:
```suggestion
        auto* closeBtn = new QPushButton;
```

In most parts of the code we use auto in this case

jensschmeink (commented):
I added 2 comments but overall I like the feature

src/gui/widgets/gt_tabwidget.cpp
mariusalexander: I have implemented a `DynamicRange` class to remove the classic for loop. I think this is a good addition

jensschmeink:
@mariusalexander IMHO its ok like this

mariusalexander:
@rainman110 May you have a alook at the `DybamicRange` class? One could also add a step size parameter or support negative ranges (e.g. 10...0)

rainman110:
💋 This is really cool! I just tested it. Great work. Now I just need to look at the code ;)

src/gui/gt_mdilauncher.cpp
```diff
@@ -438,19 +451,88 @@ GtMdiLauncher::open(const QString& id, GtObject* data, const QString& customId)
         icon = gt::gui::icon::layers();
     }
 
+    // custom widget to handle redocking
+    QPointer<GtDockableFrame> dockableWidget{qobject_cast<GtDockableFrame*>(wid)};
+    auto* tabWidget = qobject_cast<GtTabWidget*>(m_area.data());
+
+    // helper function to make mdi buttons
+    auto const makeMdiButton = [tabWidget, icon, dockableWidget,
```
rainman110, line 459: Please extract to a free function in an unnamed namespace

src/gui/widgets/gt_tabwidget.cpp
```diff
@@ -22,6 +24,42 @@ GtTabWidget::GtTabWidget(QWidget* parent) :
...
+GtTabWidget::contains(QWidget const* widget) const
+{
+    return tabIndex(widget) >= 0;
+}
+
+QList<QWidget*>
+GtTabWidget::widgets()
+{
+    QList<QWidget*> list;
+    for (int i : gt::range(count()))
```
rainman110, line 50: Here, it would have been cool to use C++ 20 ranges funtionality to create a range of widgets first and the use std::copy_if. But so what ;)

rainman110 (commented):
Nice work, I just have some minor change request and some questions.

rainman110:
@mariusalexander How does redocking work? I can't figure it out...

rainman110:
@mariusalexander I unfortunately found a bug. Drag and Drop is not yet working anymore with undocked MDI widgets. Please fix this before a merge can be approved.

src/gui/gt_mdilauncher.cpp
```diff
@@ -146,12 +153,26 @@ GtMdiLauncher::setFocus(const QString& mdiId)
...
 }
 
+void
+GtMdiLauncher::close()
+{
+    // the ownership is quite a mess, deleting the widget
+    for (GtMdiItem* item : m_openItems)
+    {
+        assert(item);
+        assert(item->widget());
```
rainman110, line 163: A warning from CPP check:

```
severity="warning" msg="Assert statement calls a function which may have desired side effects: &apos;widget&apos;." verbose="Non-pure function: &apos;widget&apos; is called inside assert statement. Assert statements are removed from release builds so the code inside assert statement is not executed. If the code is needed also in release builds, this is a bug." cwe="398" file0="src/gui/gt_mdilauncher.cpp"
```

rainman110:
One GUI test is failing that seems to be related directly to the new frame code:
https://gtlab.pages.gitlab.dlr.de/-/internal/github-mirrors/gtlab-core-mirror/-/jobs/3846663/artifacts/gui_tests_web/index.html

src/dataprocessor/gt_utilities.h
```diff
@@ -135,6 +135,86 @@ T const& clamp(T const& value, T const& low, T const& high)
...
+    using const_iterator = iterator;
+
+    /// constructor
+    constexpr explicit DynamicRange(T from, T to) noexcept :
+        m_begin(from), m_end(to)
+    {
+        assert(to >= from || !"only positive ranges are supported!");
+    }
+
+    constexpr iterator begin() const noexcept { return iterator(m_begin); }
```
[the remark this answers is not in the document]
mariusalexander: @rainman110 I am not sure what you mean, isn't this the behavior for all "end" iterators? 

Say I iterate over a vector with 3 elements, "begin" points to the first, "end" points to the 4th (which does not exist). I can increment the "begin" iterator three times until I have reached the end. If I then increment the iterator further it has passed "end", hasn't it? In this case `iter == end` would not return true, thus fail. As far as I can see this is the same behavior for the `DynamicRange` class, or do I miss something? In that case, I would argue that this behavior is fine.

src/gui/gt_mdiitem.cpp
```diff
@@ -8,21 +8,23 @@
...
 #include "gt_datamodel.h"
 #include "gt_projectchangedevent.h"
 #include "gt_objectchangedevent.h"
 #include "gt_statehandler.h"
 
-GtMdiItem::GtMdiItem() : m_frame(new QFrame), m_d(nullptr),
-    m_subWin(nullptr), m_queueEvents(false)
+
+GtMdiItem::GtMdiItem() :
+    m_frame(new GtDockableFrame),
```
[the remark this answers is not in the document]
mariusalexander: Exactly, this I why we need an overhaul of this API. The ownership only becomes clear, when tracing the code extensively.

src/dataprocessor/gt_utilities.h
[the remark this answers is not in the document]
rainman110: > Say I iterate over a vector with 3 elements, "begin" points to the first, "end" points to the 4th (which does not exist). I can increment the "begin" iterator three times until I have reached the end. If I then increment the iterator further it has passed "end", hasn't it? In this case `iter == end` would not return true, thus fail. As far as I can see this is the same behavior for the `DynamicRange` class, or do I miss something? In that case, I would argue that this behavior is fine.

If this is really the case with standard containers, that increasing the end iterator is not end anymore, then I am fine with it. Otherwise, it could lead to problems.

mariusalexander:
> @mariusalexander How does redocking work? I can't figure it out... Maximizing does not work here.

Minimizing should do the trick

mariusalexander:
> I unfortunately found a bug. Drag and Drop is not yet working anymore with undocked MDI widgets. Please fix this before a merge can be approved.

To clarify, you mean drag and drop actions not draging the window back into the central widget area to make it redock...?

Are you sure it does not work, or can you elaborate? I can open the memento viewer and detach it. Once detached I can drag any object into the window and it will display the memento of that object. Thus I think it works as intended.

src/gui/gt_mdilauncher.cpp
mariusalexander: @rainman110 I think we can ignore this, it simply calls a getter, thus there are no side effects. Can I somehow mute the warning?

src/gui/gt_mdilauncher.cpp
```diff
@@ -438,19 +451,88 @@ GtMdiLauncher::open(const QString& id, GtObject* data, const QString& customId)
...
+        buttonLay->setContentsMargins(0, 0, 0, 0);
+        buttonLay->addWidget(undockBtn);
+        buttonLay->addWidget(closeBtn);
+
+        auto* layoutWidget = new QWidget;
+        layoutWidget->setLayout(buttonLay);
+
+        connect(closeBtn, &QPushButton::clicked, dockableWidget, &QObject::deleteLater);
+
+        if (!dockableWidget) return layoutWidget;
```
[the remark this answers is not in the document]
mariusalexander: I can remove this. `DockableWidget` should never be null IMO

src/dataprocessor/gt_utilities.h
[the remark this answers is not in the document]
mariusalexander: I'll check

rainman110:
I meant dragging GTObjects into the window. I tested it with the 3D cad kernel and predesign editor. There it did not work but it worked with the docker window. I am gonna test it again

src/dataprocessor/gt_utilities.h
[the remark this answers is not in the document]
mariusalexander: Okay you might be right. An assertion is triggered when trying to increment iterator past end for std::vector... But only in debug as far as I can tell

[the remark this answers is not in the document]
mariusalexander: Good old stackoverflow: https://stackoverflow.com/questions/1057724/what-happens-if-you-increment-an-iterator-that-is-equal-to-the-end-iterator-of-a:



> Following is the quote from Nicolai Josuttis book:
>
> >   Note that advance() does not check whether it crosses the end() of a sequence (it can't check because iterators in general do not know the containers on which they operate). Thus, calling this function might result in undefined behavior because calling operator ++ for the end of a sequence is not defined
>
> In other words, the responsibility of maintaining the iterator within the range lies totally with the caller.

src/gui/gt_mdilauncher.cpp
rainman110: I think the "problem" here is, that `widget()` is not const. Hence, cppcheck thinks it as a side-effect. You can mute the warning via
`// cppcheck-suppress assertWithSideEffect`

before this line

mariusalexander: Thank you!

src/gui/gt_mdiitem.cpp
```diff
@@ -54,12 +56,6 @@ GtMdiItem::setData(GtObject* /*obj*/)
 
 GtMdiItem::~GtMdiItem()
 {
-//    gtError() << "MDI ITEM DESTROYED!";
-    if (m_frame && !m_frame->parent())
```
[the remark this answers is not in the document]
rainman110: Have you tried it with deleteLater() ?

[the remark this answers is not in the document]
rainman110: Why is it crashing in the undocked case?

mariusalexander:
> It seems, that opening the 3D viewer makes all undocked widgets not accepting drags anymore.

Woa, that is strange!! What should we do about it?

src/dataprocessor/gt_utilities.h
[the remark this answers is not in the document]
rainman110: Okay, than lets keep it as it is.

rainman110:
> Woa, that is strange!! What should we do about it?

I would say, we need to look at the CADKernel and PreDesign and put this as a known issue into the ChangeLog.

mariusalexander:
Applied your suggestions. I have not yet looked into the issue with the GUI Tests, maybe it fixed itself (fingers crossed)

mariusalexander:
> I would say, we need to look at the CADKernel and PreDesign and put this as a known issue into the ChangeLog.

Updated the changelog

CHANGELOG.md
```diff
@@ -26,6 +26,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - MDI Items can now be closed using middle mouse click (similarly to most common browsers). - #603
+- Allowed MDI Items to be detached and redocked by minimizing the window. Finally enabling multi-window workflows with GTlab (*Note:* This feature is still WIP). - #604
+
+    *Known Issues:*
+	- Detaching 3D viewers may interfer with other MDI Items and cause strange side effects.
```
rainman110, line 32:
```suggestion
	- Dropping objects into detached MDI windows stops working after opening a 3D PreDesign CAD viewer.
```

rainman110:
> > One GUI test is failing that seems to be related directly to the new frame code: https://gtlab.pages.gitlab.dlr.de/-/internal/github-mirrors/gtlab-core-mirror/-/jobs/3846663/artifacts/gui_tests_web/index.html
> 
> Seems to have fixed itself!

No, your link leads to one failed test.

mariusalexander:
> > > One GUI test is failing that seems to be related directly to the new frame code: https://gtlab.pages.gitlab.dlr.de/-/internal/github-mirrors/gtlab-core-mirror/-/jobs/3846663/artifacts/gui_tests_web/index.html
> > 
> > 
> > Seems to have fixed itself!
> 
> No, your link leads to one failed test:
> 
> ```
>  ERROR10:49:57
> Script Error
>  LookupError: Object '{type='GtGraphicsView' unnamed='1' visible='0' window='{type=\'GtDockableFrame\' unnamed=\'1\' visible=\'0\'}'}' not ready. (Screenshot in "/home/gitlab-runner/builds/3AQtSC_d/0/gtlab/internal/github-mirrors/gtlab-core-mirror/./gui_tests_web/data/gtlab_interface_gui_tests/tst_mdi/errorImages/error_1.png")
> ```
This was the link you provided originally. Currently all Checks in GitHub pass

rainman110 (approved):
LGTM now

PR #1238, opened 2024-06-07
Removed most of the deprecated functions

### LINKED ISSUE #1237: [CHANGE] - Remove deprecated GTlab functions
### Summary

We have deprecated a variety of functions in the GTlab 2.0 series. They should be removed with 2.1.

### Impact on the codebase

API and ABI changes. GTlab itself should be not affected, modules might be, if they still use the deprecated functions.

### Current workarounds

Not required

rainman110:
Still missing:
 - Variant property without id: No real alternative, since the property is not stored to memento, if without an id
 - ObjectUI: No real alternative function accepting a method name (string) instead of a callable

Still todo:
 - [ ] Document deprecated functions replacement

## Checklist:


src/dataprocessor/property/gt_intproperty.h
```diff
@@ -68,62 +68,6 @@ class GT_DATAMODEL_EXPORT GtIntProperty : public GtProperty<int>
...
-     * @brief GtIntProperty
-     * @param ident
-     * @param name
-     * @param brief
-     * @param unitCategory
-     * @param value
-     */
-    [[deprecated("Int Properties do not support units properly and "
-                 "will be removed in the future")]]
-    GtIntProperty(const QString& ident,
```
rainman110, line 81: Note removing this can lead to hard-to-find side-effects:

Lets assume the following user code

    m_intProp("intProp", "Test Int", QString(), GtUnit::Category::Angle)
    
After removal of this constructor, the other constructor 

        GtIntProperty(const QString& ident,
                  const QString& name,
                  const QString& brief,
                  int value);

will be silently used, resulting in a value != 0.

@mariusalexander Do you have an idea, what we can do?

rainman110: One idea is to convert the Category namespace in a named enum. Then it should not be possible to use implicit conversion to int, right?

mariusalexander: I like the idea of using a class enum.

An alternative solution would be to delete the "false" constructor explicitly:

```cpp
GtIntProperty(const QString& ident,
              const QString& name,
              const QString& brief,
              GtUnit::Category cat) = delete;
 ```
 
 This should work as expected

rainman110: I had a quick test. Replacing the enum with a named enum requires many code (trivial) code changes in core and this probably also in other modules. 

So this is unfortunately a no go for a 2.1 release .

I'll try to explicitly disable the old deprecated constructor.

PR #1239, opened 2024-06-11
Remove monitoring property as separate type

jensschmeink:
The separate class GtMonitoringProperty and the specific classes GtDoubleMonitoring (and int and string) are not necessary if the abstract property has a flag if it is a monitoring one.

## Description
The class GtMonitoringProperty is removed.  
The GtAbstractProperty now has a flag for monitoring properties.  
The existing usages are replaced by "normal" properties. To reduce the need of code changes a "using" is added to read e.g. GtDoubleMonitoringProperties as gtDoubleProperties.   
The setting of the monitoring flag is done in the "registerMonitoringProperty" function.  
  
With this kind of changes the developers of other modules have to recompile but not to change their code.   
  
## How Has This Been Tested?

Manual. Please check the usage also while the review.

I started some process using monitoring properties and it worked as before the changes. Monitoring properties are set and read as before.

## Checklist:


src/core/process_management/gt_doublemonitoringproperty.h
```diff
@@ -15,39 +15,8 @@
...
-     * @param ident for the datamodel
-     * @param name shown in the GUI
-     * @param brief description text
-     */
-    Q_INVOKABLE GtDoubleMonitoringProperty(const QString& ident,
-                                           const QString& name,
-                                           const QString& brief);
-
-};
+using GtDoubleMonitoringProperty = GtDoubleProperty;
```
rainman110, line 19: This is IMHO problematic, as this alias does not yet set the monitoring flag, right? 

I would rather remove the type. Then you can at least see, where you need to add the flag.

Another option would be to defined GtDoubleMinitoringProperty derived from DoubleProperty, but set the flag inside the constructor.

src/core/process_management/gt_processcomponent.cpp
```diff
@@ -85,20 +84,15 @@ GtProcessComponent::registerMonitoringProperty(GtAbstractProperty& property)
...
-    }
-
     // register property
     if (!registerProperty(property, QStringLiteral("Monitoring")))
     {
         gtError() << tr("register monitoring property failed!");
         return;
     }
 
+    property.setMonitoring(true);
```
rainman110, line 94: Somehow, it is a bit strange, that you currently can set the monitoring flag to a property but you still need to register it to a calculator.

Would't it make sense, that a calculator register all its monitoring properties automatically, that have the flag monitoring?

src/core/process_management/gt_doublemonitoringproperty.h
jensschmeink: From discussion in GTlab meeting: Set flag in constructor and not in registration (second option)

src/core/process_management/gt_processcomponent.cpp
rainman110: Note for myself: Since we wan't backwards compatibilty, we should keep the function.

We though should get rid of the monitoringProperties entry and provide a function, that dynamically returns the monitoring properties.

src/core/process_management/gt_intmonitoringproperty.cpp
```diff
@@ -11,23 +11,16 @@
...
-
-GtIntMonitoringProperty::GtIntMonitoringProperty(const QString& ident,
-                                                 const QString& name) :
-    GtIntMonitoringProperty(ident, name, QString())
-{
-
-}
-
 gt::PropertyFactoryFunction
 gt::makeIntMonitoringProperty(int value)
```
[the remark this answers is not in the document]
jensschmeink: @rainman110 do you have a hint for how to implement something like this? I'm not that familiar with the std::function

src/core/process_management/gt_doublemonitoringproperty.h
jensschmeink: I added a new implementation for the constructors of double and int monitoring properties

src/core/process_management/gt_task.h
```diff
@@ -20,7 +20,7 @@
 
 #include "gt_processcomponent.h"
 #include "gt_objectmemento.h"
-#include "gt_intmonitoringproperty.h"
+#include "gt_intproperty.h"
```
jensschmeink, line 23:
```suggestion
#include "gt_intmonitoringproperty.h"
#include "gt_intproperty.h"
```

jensschmeink: This include is needed to avoid code changes in modules

jensschmeink:
@rainman110 I applied the points we discussed. Might you check this again?

rainman110:
@jensschmeink Before i go into the code review. Can you please tell, how (in detail) you have tested the functonality? I think, it makes sense to test it as part of the master build using some other modules and run a small process chain.

jensschmeink:
> @jensschmeink Before i go into the code review. Can you please tell, how (in detail) you have tested the functonality? I think, it makes sense to test it as part of the master build using some other modules and run a small process chain.

Good idea. Currently I tested only small parts, but I'll use a broader scope.

jensschmeink:
I started some process using monitoring properties and it worked as before the changes. Monitoring properties are set and read as before.

tests/unittests/datamodel/test_gt_doublemonitoringproperty.cpp
```diff
@@ -47,6 +47,11 @@ TEST_F(TestGtDoubleMonitoringProperty, initialization)
 
 TEST_F(TestGtDoubleMonitoringProperty, isReadOnly)
 {
+    // ToDo: This test is failing.
+    // We should discuss how the monitoring propeties should be defined
+    // currently properties become monitoring properties
+    // by the registration as monitoring properties
+    // and also becomes read only in this procedure
```
jensschmeink, line 54:
```suggestion
```

tests/unittests/datamodel/test_gt_intmonitoringproperty.cpp
```diff
@@ -34,6 +34,12 @@ TEST_F(TestGtIntMonitoringProperty, initialization)
 
     ASSERT_STREQ(prop.ident().toStdString().c_str(), "testIdent");
 
+    // ToDo: This test is failing.
+    // The GtIntProperty does behave a bit different compared to GtIntMonitoringProperty
+    // or even GtDoubleProperty and GtStringpropety:
+    // The GtIntProperty now used does set the brief value to the same as the name value
+    // if the constructor with only two arguments is used.
+    // THis difference should be discussed
```
jensschmeink, line 42:
```suggestion
```

tests/unittests/datamodel/test_gt_intmonitoringproperty.cpp
```diff
@@ -47,6 +53,11 @@ TEST_F(TestGtIntMonitoringProperty, initialization)
 
 TEST_F(TestGtIntMonitoringProperty, isReadOnly)
 {
+    // ToDo: This test is failing.
+    // We should discuss how the monitoring propeties should be defined
+    // currently properties become monitoring properties
+    // by the registration as monitoring properties
+    // and also becomes read only in this procedure
```
jensschmeink, line 60:
```suggestion
```

tests/unittests/datamodel/test_gt_stringmonitoringproperty.cpp
```diff
@@ -47,6 +47,11 @@ TEST_F(TestGtStringMonitoringProperty, initialization)
 
 TEST_F(TestGtStringMonitoringProperty, isReadOnly)
 {
+    // ToDo: This test is failing.
+    // We should discuss how the monitoring propeties should be defined
+    // currently properties become monitoring properties
+    // by the registration as monitoring properties
+    // and also becomes read only in this procedure
```
jensschmeink, line 54:
```suggestion
```

tests/unittests/core/test_gt_processtestclasses.h
```diff
@@ -17,6 +17,7 @@
 #include "gt_relativeobjectlinkproperty.h"
 #include "gt_propertystructcontainer.h"
 #include "gt_structproperty.h"
+#include "gt_intmonitoringproperty.h"
```
jensschmeink, line 20:
```suggestion
```

src/dataprocessor/property/gt_abstractproperty.h
```diff
@@ -295,6 +303,9 @@ class GT_DATAMODEL_EXPORT GtAbstractProperty : public QObject
     /// Whether property is stored in memento information or not
     bool m_storeMemento{true};
 
+    /// Monitoring indicator
+    bool m_monitoring{false};
```
rainman110, line 307: Just thinking out loud here...

The additional of this flag makes the ABI change again (which is okay), but it cries for a pimplification.
The problem is, that the flags are protected and not private, which makes it hard to pimpl it without changing client code.

jensschmeink: pimpl is used in the latest version of the branch

jensschmeink:
@rainman110 please review this PR

rainman110 (approved):
This look good to me know.

This is a pretty serious change though. I hope you have tested this extensively on some real-world examples.

PR #1240, opened 2024-06-12
Enable skip/unsik hotkey for multiple elements

jensschmeink:
## Description
The function to skip or unskip elements via a shortcuts is extended to be used for selection of more of one element.  
  
There is added a very small cleanup to reduce code. 

## How Has This Been Tested?
Manual tests

## Checklist:

### DIFF
--- a/src/app/dock_widgets/process/gt_processdock.cpp
+++ b/src/app/dock_widgets/process/gt_processdock.cpp
@@ -187,8 +187,8 @@ GtProcessDock::GtProcessDock() :
             SLOT(pasteElement(QModelIndex)));
     connect(m_view, SIGNAL(runTaskElement(QModelIndex)),
             SLOT(runProcess()));
-    connect(m_view, SIGNAL(skipCalcultorElement(QModelIndex,bool)),
-            SLOT(skipComponent(QModelIndex,bool)));
+    connect(m_view, SIGNAL(skipCalcultorElement(const QList<QModelIndex>&,bool)),
+            SLOT(skipComponents(const QList<QModelIndex>&,bool)));
     connect(m_view, SIGNAL(renameProcessElement(QModelIndex)),
             SLOT(renameElement()));
     connect(m_view, SIGNAL(moveProcessElements(QList<QModelIndex>, QModelIndex)),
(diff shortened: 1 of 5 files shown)


jensschmeink:
This PR is ok for 2.0.X as it does not affect any code available in the libraries but only the application

src/app/dock_widgets/process/gt_processdock.cpp
```diff
@@ -187,8 +187,8 @@ GtProcessDock::GtProcessDock() :
             SLOT(pasteElement(QModelIndex)));
     connect(m_view, SIGNAL(runTaskElement(QModelIndex)),
             SLOT(runProcess()));
-    connect(m_view, SIGNAL(skipCalcultorElement(QModelIndex,bool)),
-            SLOT(skipComponent(QModelIndex,bool)));
+    connect(m_view, SIGNAL(skipCalcultorElement(const QList<QModelIndex>&,bool)),
+            SLOT(skipComponents(const QList<QModelIndex>&,bool)));
```
mariusalexander, line 191: You may want to use this notation instead, that way it does not matter which arguments the functions have or if they are changing
```suggestion
    connect(m_view, &GtProcessView::skipCalcultorElement,
                 this, &GtProcessDock::skipComponents));
```

src/app/dock_widgets/process/gt_processdock.cpp
```diff
@@ -1220,26 +1220,23 @@ GtProcessDock::processContextMenu(GtProcessComponent& obj,
 void
 GtProcessDock::multiSelectionContextMenu(QList<QModelIndex> const& indexList)
 {
-    QMenu menu(this);
-
-    QAction* skipCalcs = menu.addAction(tr("Skip Selected Elements"));
-    skipCalcs->setIcon(gt::gui::icon::skip());
+//    if (indexList.isEmpty()) return;
```
mariusalexander, line 1223:
```suggestion
```

mariusalexander (commented):
I would suggest only minor code cleanup

src/app/dock_widgets/process/gt_processdock.cpp
jensschmeink: @mariusalexander I think this is only possible for static functions, isn't it?

rainman110:
@jensschmeink An this? I also think, this deserves a Changelog entry.

jensschmeink:
> @jensschmeink An this? I also think, this deserves a Changelog entry.

It is done. I can add the changelog entry as recommended.

rainman110:
@jensschmeink I would approve now, but I can't. I added some final code cleanups.

rainman110 (approved):

PR #1242, opened 2024-06-13
Remove all connections at once by button

jensschmeink:
## Description

The new button removes all graphics entities for the connections and the related connection objects.

grafik

## How Has This Been Tested?
Manual

## Checklist:

### DIFF
--- a/src/app/dock_widgets/process/gt_processconnectioneditor.cpp
+++ b/src/app/dock_widgets/process/gt_processconnectioneditor.cpp
@@ -66,15 +66,20 @@ GtProcessConnectionEditor::GtProcessConnectionEditor(GtTask* task,
 
     lay->addLayout(mainLay);
 
-    QPushButton* saveButton = new QPushButton(tr("Ok"));
+    auto* removeAllButton =
+            new QPushButton(tr("Remove all connections"));
+    removeAllButton->setIcon(gt::gui::icon::delete_());
+    auto* saveButton = new QPushButton(tr("Ok"));
     saveButton->setIcon(gt::gui::icon::check());
-    QPushButton* closeButton = new QPushButton(tr("Cancel"));
+    auto* closeButton = new QPushButton(tr("Cancel"));
     closeButton->setIcon(gt::gui::icon::cancel());
 
     connect(closeButton, SIGNAL(clicked()), SLOT(reject()));
     connect(saveButton, SIGNAL(clicked()), SLOT(accept()));
+    connect(removeAllButton, SIGNAL(clicked()), SLOT(deleteAllConnections()));
 
     QHBoxLayout* buttonsLayout = new QHBoxLayout;
+    buttonsLayout->addWidget(removeAllButton);
     buttonsLayout->setContentsMargins(4, 4, 4, 4);
     buttonsLayout->addStretch(1);
     buttonsLayout->addWidget(saveButton);
@@ -116,6 +121,12 @@ GtProcessConnectionEditor::connectionData()
     }
 
     return GtObjectMemento();
+}
+
+void
+GtProcessConnectionEditor::deleteAllConnections()
+{
+    m_connectionView->removeAllConnections();
 }
 
 void
--- a/src/app/dock_widgets/process/gt_processconnectioneditor.cpp
+++ b/src/app/dock_widgets/process/gt_processconnectioneditor.cpp
@@ -126,6 +126,8 @@ GtProcessConnectionEditor::connectionData()
 void
 GtProcessConnectionEditor::deleteAllConnections()
 {
+    if (!m_connectionView->root()) return;
+
     m_connectionView->removeAllConnections(m_connectionView->root()->uuid());
 }
 
--- a/src/app/dock_widgets/process/gt_processconnectioneditor.cpp
+++ b/src/app/dock_widgets/process/gt_processconnectioneditor.cpp
@@ -126,17 +126,14 @@ GtProcessConnectionEditor::connectionData()
 void
 GtProcessConnectionEditor::deleteAllConnections()
 {
-    m_connectionView->removeAllConnections();
+    m_connectionView->removeAllConnections(m_connectionView->root()->uuid());
 }
 
 void
 GtProcessConnectionEditor::fillData()
 {
     // check task
-    if (!m_task)
-    {
-        return;
-    }
+    if (!m_task) return;
 
     // fill output model and expand view
     m_outputModel->setRoot(m_task);
--- a/src/app/mdi_items/process_env/entities/gt_processconnectionview.cpp
+++ b/src/app/mdi_items/process_env/entities/gt_processconnectionview.cpp
@@ -11,7 +11,12 @@
 
 #include <QPainter>
 #include <QApplication>
+#include <QMenu>
 
+#include <gt_logging.h>
+#include "gt_icons.h"
+#include "gt_guiutilities.h"
+#include "gt_objectuiaction.h"
 #include "gt_processconnectionmodel.h"
 #include "gt_processconnectiongraphicsview.h"
 
@@ -22,6 +27,12 @@ GtProcessConnectionView::GtProcessConnectionView(QWidget* parent) :
     m_graphicsView(nullptr)
 {
     setIconSize({16, 16});
+    setContextMenuPolicy(Qt::CustomContextMenu);
+
+    connect(this, SIGNAL(customContextMenuRequested(QPoint)),
+            SLOT(customContextMenu(QPoint)));
+
+
 }
 
 void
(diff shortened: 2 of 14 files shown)


jensschmeink:
This PR should be merged in 2.0.X as it does not affect any exported classes

jensschmeink:
Add better colors for dark mode  
grafik

This is ofcourse only a suggestion.

mariusalexander:
@jensschmeink I like the new colors! 

I am wondering however, whether we should make it also possible to only remove connections of a specific component? After all the connection editor can only be opened on the root task, correct? 

Say we have a deeply nested process chain and you only want to disconnect a specific calculator or subtask. Maybe you could all a context menu entry to each component to address this use case? However, I do not use the connection editor that often so maybe this use case does not apply.

rainman110:
I also agree with @mariusalexander, that we probably need a more fine-grained control on the removal.

I also like the idea of a context menu entry, when you select an object to remove all connections to that objects (and its properties). By clicking on the root object, you basically remove all connections.

src/gui/gt_colors.h
```diff
@@ -229,6 +229,15 @@ GT_GUI_EXPORT QColor gridPoint();
 
 GT_GUI_EXPORT QColor gridAxis();
 
+namespace connection_editor
+{
+GT_GUI_EXPORT QColor connection();
+GT_GUI_EXPORT QColor connectionDraft();
+GT_GUI_EXPORT QColor connectionHighlight();
+GT_GUI_EXPORT QColor portBrush();
```
mariusalexander, line 237: I assumed it would return a QBrush. I would prefer a name like simply `port()` or `portBackground()`

src/gui/gt_colors.h
```diff
@@ -229,6 +229,15 @@ GT_GUI_EXPORT QColor gridPoint();
...
+GT_GUI_EXPORT QColor portHoverEnter();
```
mariusalexander, line 238: Same here, I would suggest simply `portHover()` or `portHoverBackground()`

jensschmeink:
> @jensschmeink I like the new colors!
> 
> I am wondering however, whether we should make it also possible to only remove connections of a specific component? After all the connection editor can only be opened on the root task, correct?
> 
> Say we have a deeply nested process chain and you only want to disconnect a specific calculator or subtask. Maybe you could all a context menu entry to each component to address this use case? However, I do not use the connection editor that often so maybe this use case does not apply.

Yes, you are right. The editor is always opened only for the root task.
I like your idea with the context menu and will implement a first version

mariusalexander:
@jensschmeink What causes this black line in you screenshot?

grafik

jensschmeink:
> @jensschmeink What causes this black line in you screenshot?
> 
> grafik

Good question. It seems to be a strange artifact of the connections. I have not changed anything about that. Is it also part of the existing implementation?

mariusalexander:
I don't see it in my editor

> > @jensschmeink What causes this black line in you screenshot?
> > grafik
> 
> Good question. It seems to be a strange artifact of the connections. I have not changed anything about that. Is it also part of the existing implementation?

I do not see it in my editor

jensschmeink:
grafik  

The new version has the button to remove all connections but you can also right cklick any process element to remove all connections of this process element. But it is seperated for input and output. So if you use the disconnect on the left (input) side all input nodes of this process element (and all of its children) are removed but not there connections on the output side. This behaviour was the result of a talk of @mariusalexander and @jensschmeink

rainman110:
@jensschmeink What is the status of this PR?

jensschmeink:
> @jensschmeink What is the status of this PR?

It is done but needs a final review

src/gui/gt_colors.cpp
```diff
@@ -474,5 +474,52 @@ gt::gui::color::js_highlight::marker()
...
+}
 
+QColor
+gt::gui::color::connection_editor::connectionHighlight()
+{
+    if (gtApp->inDarkMode())
+    {
+        return Qt::red;
+    }
+    return Qt::red;
```
rainman110, line 524: Why do you check for dark mode? Also, this function will crash, if gtApp is a nullptr, i.e. in unit tests.

jensschmeink: There was the plan to adapt the color depending on the test of the members of the GTlab Core meeting.

jensschmeink: Not right now so it should be ready for a review

src/app/mdi_items/process_env/entities/gt_processpropertyconnectionentity.cpp
```diff
@@ -168,12 +169,53 @@ GtProcessPropertyConnectionEntity::connection()
     return m_connection;
 }
 
+void
+GtProcessPropertyConnectionEntity::removeConnection()
+{
+    if (m_connection)
```
rainman110, line 175: Please use early return as in the  method below to improve readability.

src/app/mdi_items/process_env/entities/gt_processpropertyconnectionentity.h
```diff
@@ -112,6 +112,20 @@ class GtProcessPropertyConnectionEntity : public GtGraphicsAnimatedPathItem
...
+     */
+    void removeConnection();
+
+    /**
+     * @brief connectedToProcessComponent
+     * @param uuid of process component to check
+     * @return true if the connection is connected to the process component
+     * given by the uuid
+     */
+    bool connectedToProcessComponent(QString const& uuid, bool inPort);
```
rainman110, line 127: This should be const, shouldn't it?

src/app/mdi_items/process_env/gt_processconnectiongraphicsview.cpp
```diff
@@ -365,7 +365,47 @@ GtProcessConnectionGraphicsView::updateConnections()
...
+    }
+
+    // make sure to avoid double checks
+    allUuids.removeDuplicates();
+
+    QList<QGraphicsItem*> allItems = items();
+
+    for (auto i : qAsConst(allItems))
+    {
+        if (auto e = dynamic_cast<GtProcessPropertyConnectionEntity*>(i))
```
rainman110, line 397: Could you replace this with qobject_cast, since this should have better performance.

You could also use early continue here, if i is not an entity. Also rename i -> item, otherwise it looks like a counter.

src/app/mdi_items/process_env/gt_processconnectiongraphicsview.cpp
```diff
@@ -365,7 +365,47 @@ GtProcessConnectionGraphicsView::updateConnections()
...
+        {
+            for (auto const& u : allUuids)
```
rainman110, line 399: Use std::find_if

src/app/dock_widgets/process/gt_processconnectioneditor.cpp
```diff
@@ -118,14 +123,17 @@ GtProcessConnectionEditor::connectionData()
     return GtObjectMemento();
 }
 
+void
+GtProcessConnectionEditor::deleteAllConnections()
+{
+    m_connectionView->removeAllConnections(m_connectionView->root()->uuid());
```
rainman110, line 129: What happens, if you don't have a root object? Can this happen?

src/app/mdi_items/process_env/entities/gt_processconnectionview.cpp
```diff
@@ -194,3 +208,35 @@ GtProcessConnectionView::paintEvent(QPaintEvent* event)
...
+{
+    QModelIndex index = indexAt(p);
+
+    if (!index.isValid()) return;
+
+    GtProcessConnectionModel* connModel = connectionModel();
+
+    if (!connModel) return;
+
+    GtProcessConnectionItem* i = connModel->itemFromIndex(index);
```
rainman110, line 223: rename to item

src/app/mdi_items/process_env/entities/gt_processconnectionview.cpp
```diff
@@ -194,3 +208,35 @@ GtProcessConnectionView::paintEvent(QPaintEvent* event)
...
+
+    if (!i) return;
+
+    if (i->itemType() != GtProcessConnectionItem::PROCESS_COMPONENT) return;
```
rainman110, line 227:
```suggestion
    if (!i || i->itemType() != GtProcessConnectionItem::PROCESS_COMPONENT) return;
```

rainman110 (changes requested):

src/app/mdi_items/process_env/gt_processconnectiongraphicsview.cpp
jensschmeink: The class QGraphicsItem is not a qObject and therefore the qobject_cast is not working here

rainman110: Ok, makes sense

rainman110:
Damn, we forgot the changelog. I'll add an entry.

PR #1259, opened 2024-06-25
Added project dir to package read/save api

### LINKED ISSUE #617: Update GtPackage API to allow saving a external package data to other locations when using "SaveAs"
In GitLab by @mariusalexander on Nov 22, 2023, 07:38

The GtPackage API does not provide direct access to the project that should be saved. Since access to the project path is only possible using `gtApp->currentProject`, this results in a bugged behaviour when saving a package that stores some of its data in external files. The data will not be written to the new project but to old one instead!

rainman110:
Closes #617

## Description

__ ABI CHANGE__

- Added new package read/write functions to store miscelleneous data, like intelligraphs etc.
- The new functions are directly executed after calling the xml-based ones.
- Using the new APi, the currentProject()->path workaround can be avoided, which is error prone.
- Modules, that don't exploit this functionality require only a recompile
- Other modules (like intelligraph) should be adapted to use this new API

@mariusalexander I quickly adapted locally the intelligraph package to test this. Seems to be working.
Please still have a look!

## How Has This Been Tested?

Locally, with an adapted intelligraph.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased 2.1]
+### Changed
+ - GtPackage and all derived classes now use an additional project directory as its first argument in `GtPackage::read` and `GtPackage::save`. This will impact module code! - #617
+
 ## [2.0.8] - 2024-06-19
 ### Fixed
 - Fixed bug introduced in 2.0.7, that "GTlabConsole run" does not save projects anymore. This change reverts
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased 2.1]
-### Changed
- - GtPackage and all derived classes now use an additional project directory as its first argument in `GtPackage::read` and `GtPackage::save`. This will impact module code! - #617
+### Added
+ - New interface methods `GtPackage::readMiscData` and  `GtPackage::saveMiscData` to store package data outside of the package xml structure inside the project directory.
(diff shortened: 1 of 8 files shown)


mariusalexander:
> Update: I think I have a better idea to overcome an API change. We add different function
>
>   - saveToProjectDir
> -    readFromProjectDir
>
> These can be used, whenever non-xml files should be saved / read.

So these functions ar now `saveMiscData` and `readMiscData`? I like the new API. It is a clever yet simple fix

CHANGELOG.md
```diff
@@ -10,6 +10,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased 2.1]
+### Added
+ - New interface methods `GtPackage::readMiscData` and  `GtPackage::saveMiscData` to store package data outside of the package xml structure inside the project directory.
+   Both methods have the project directory as an argument, workarounds like currentProject()->path() can be hence avoided - #617
```
mariusalexander, line 16:
```suggestion
   Both methods have the project directory as an argument, hence workarounds like currentProject()->path() can be avoided - #617
```

mariusalexander (changes requested):

rainman110:
> So these functions ar now `saveMiscData` and `readMiscData`? I like the new API. It is a clever yet simple fix

Correct. These functions should be used, when data should be read/saved that are not part of the xml document tree.

src/core/gt_project.cpp
```diff
@@ -686,7 +676,7 @@ GtProject::saveModuleData()
 
         document.appendChild(rootElement);
 
-        if (!package->saveData(rootElement, document))
+        if (!package->saveData(rootElement, document) && !package->saveMiscData(QDir(m_path)))
```
rainman110, line 679: @mariusalexander When I now think about it, it should be an "||" instead of a "&&", right?

mariusalexander:
Actually, I found a bug. The `readMiscData` and `saveMiscData` were not called if `readData` or `saveData` returned true. I had to change the operator `&&` to `||`.

mariusalexander (approved):

rainman110:
@mariusalexander This will be our first, true merged 2.1 MR :D

mariusalexander:
> @mariusalexander This will be our first, true merged 2.1 MR :D

Pioneers!

Issue #482, opened 2023-03-14
Preferences: sort shortcuts alphabetically


rainman110:
In GitLab by @mariusalexander on Mar 14, 2023, 10:07

In preferences -> short cuts, alphabetical ordering works only for columns Function and Category but not for ShortCut.

Schould be moved from gui lib to app

see [bug report](https://gitlab.dlr.de/gtlab-bugtracking/gtlab-issuetracking/-/issues/48)

PR #1271, opened 2024-08-13
feat.: Enable Sorting for QLineEdits in Preference View

NitenIchiry:
## Description

Adding GtShortCutEdit as an Item to The Table and Implemented a special comparison method for sorting the entries according to the ShortCut-String.

Closes https://github.com/dlr-gtlab/gtlab-core/issues/482

## How Has This Been Tested?

Worked in local GTlab instance.

## Checklist:

### DIFF
--- a/src/gui/widgets/gt_shortcutedit.h
+++ b/src/gui/widgets/gt_shortcutedit.h
@@ -16,18 +16,34 @@
 
 #include "gt_lineedit.h"
 #include <QKeySequence>
+#include <QTableWidgetItem>
 
 /**
  * @brief The GtShortCutEdit class
  * Used to input a keysequence for a short cut (single key + multiple modifers).
  */
-class GT_GUI_EXPORT GtShortCutEdit : public GtLineEdit
+class GT_GUI_EXPORT GtShortCutEdit : public GtLineEdit, public QTableWidgetItem
 {
     Q_OBJECT
 
     Q_PROPERTY(QKeySequence keySequence READ keySequence WRITE setKeySequence)
 
 public:
+
+    bool operator <(const QTableWidgetItem& other) const
+    {
+        if(other.column() == 0 /* numeric cell */) {
+             return QTableWidgetItem::text().toInt() < other.text().toInt();
+        }
+        else if(other.column() == 1 /* progress bar */) {
+            const GtLineEdit *p = dynamic_cast<const GtLineEdit *>(&other);
+            if(p != 0) {
+                if(this->GtLineEdit::text() < p->text())
+                    return true;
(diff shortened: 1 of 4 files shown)


src/gui/widgets/gt_shortcutedit.h
```diff
@@ -16,19 +16,35 @@
...
 #include "gt_lineedit.h"
 #include <QKeySequence>
+#include <QTableWidgetItem>
 
 /**
  * @brief The GtShortCutEdit class
  * Used to input a keysequence for a short cut (single key + multiple modifers).
  */
-class GT_GUI_EXPORT GtShortCutEdit : public GtLineEdit
+class GT_GUI_EXPORT GtShortCutEdit : public GtLineEdit, public QTableWidgetItem
```
jensschmeink, line 25: This is an ABI change but I think this class is not used anywhere outside the GTlab Core Application.
What do you think @rainman110 ?

src/gui/widgets/gt_shortcutedit.h
```diff
@@ -16,19 +16,35 @@
...
 {
     Q_OBJECT
 
     Q_PROPERTY(QKeySequence keySequence READ keySequence WRITE setKeySequence)
 
 public:
 
+    bool operator <(const QTableWidgetItem& other) const
+    {
+        if(other.column() == 0 /* numeric cell */) {
```
jensschmeink, line 35: This part of the code is not using our style

NitenIchiry: I changed up the function and tried to use your style. Maybe we should get the stylecheck working again for these cases

src/gui/widgets/gt_shortcutedit.h
rainman110: The ABI change is not a problem, since we tagged this for 2.1

src/gui/widgets/gt_shortcutedit.h
```diff
@@ -16,19 +16,40 @@
...
+    bool operator <(const QTableWidgetItem& other) const
+    {
+        if (other.column() == 1)
+        {
+            const GtLineEdit *p = dynamic_cast<const GtLineEdit *>(&other);
+
+            if (p)
+            {
+                if (this->GtLineEdit::text() < p->text()) return true;
+            }
```
rainman110, line 47: Please combine both if statements into one

```suggestion
            if (p && this->GtLineEdit::text() < p->text()) return true;
```

rainman110 (approved):

NitenIchiry:
I hope I added it correctly. Will merge now

PR #1274, opened 2024-08-16
Add functionality to test module interface versions

### LINKED ISSUE #1273: 💡 [REQUEST] - Don't load incompatible modules
### Summary

If a module is compiled against an old (incompatible) interface, GTlab should not load it as crashes can be expected.

### Improvements

We will have the case, that old modules will be installed with GTlab 2.1. To avoid crashes, we should warn the user if he has an outdated module.

### Drawbacks

-

### Current workarounds

-

rainman110:
This PR also increases the datamodel interface version, as the GtPackage ABI was changed in a previous PR.

Closes #1273

## Description

- added function `checkInterface`  to test interface compatibility
- added macro `GT_OLD_INTERFACES` to define previous interface versions that are used to test a module against

## How Has This Been Tested?
I tried to load e.g. basictools, which is now denied. When though building basictools from scratch against this version, it will load.

Here's an example, when GTlab does not load the TestDMI module, as it was compiled against 2.0:

image

## Checklist:

### DIFF
--- a/src/core/gt_moduleloader.cpp
+++ b/src/core/gt_moduleloader.cpp
@@ -732,7 +732,7 @@ GtModuleLoader::insert(GtModuleInterface* plugin)
         gt::commandline::registerFunction(commandLineFunction);
     }
 
-    GtDatamodelInterface* dmp = dynamic_cast<GtDatamodelInterface*>(plugin);
+    GtDatamodelInterface* dmp = checkInterface<GtDatamodelInterface>(plugin->ident(), plugin);
 
     // contains dynamic linked datamodel classes
     if (dmp)
--- a/src/core/gt_moduleloader.cpp
+++ b/src/core/gt_moduleloader.cpp
@@ -611,16 +611,14 @@ GtModuleLoader::moduleLicence(const QString& id) const
 
 QString
 GtModuleLoader::getSupportedInterfaceByModule(QObject *pluginObj,
-                                              const QStringList &listOfInterfaces)
+                                              const QStringList &lOfItfs)
 {
-    QStringList supportedInterfaces;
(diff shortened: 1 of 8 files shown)


rainman110:
@jensschmeink Could you please have a look and test it with 2.0 modules and if you want also with 2.1 modules from the master build?

In theory, 2.1 modules should be loading, 2.0 modules not anymore.

src/core/gt_moduleloader.cpp
```diff
@@ -857,8 +871,8 @@ GtModuleLoader::Impl::performLoading(GtModuleLoader& moduleLoader,
                 << QObject::tr("loading ") << moduleMeta.location() << "...";
 
         // check that plugin is a GTlab module
-        auto module =
-            gt::unique_qobject_cast<GtModuleInterface>(std::move(plugin));
+        checkInterface<GtModuleInterface>(moduleMeta.location(), plugin.get());
```
mariusalexander, line 874: Is the return value not used? Why is that so?

mariusalexander:
Nicely done! If you can resolve my one comment I'll approve. Cool concept and execution!

src/core/gt_moduleloader.cpp
rainman110: This is intentional and should only be used to trigger the error message. The reason is, that the previous line already performs a check but also by converting the unique_ptr, which is not done with checkInterface.

Ideally, we get `module = checkInterface(..)` and transferring the ownership, if no nullptr was returned.

rainman110:
> Nicely done! If you can resolve my one comment I'll approve. Cool concept and execution!

Here's my proposal on the solution. I hope you like it.

mariusalexander (approved):
Cool execution!

PR #1276, opened 2024-08-19
Fix crashes when two tasks / calculators are deleted

rainman110:
## Description

This fixes the regression from 2.0.5, that deleting multiple process elements results in a crash.

The regression came from commit https://github.com/dlr-gtlab/gtlab-core/commit/37383d9c2e3b042d76ce645731d48bd6f54def0d, which had a incorrect refactoring in the end.

The change makes sure, that the single selection is not processed, when the multi-selection is handled.

## How Has This Been Tested?

1. Create two calculators
2. Delete them via context menu. 
3. Without the change, the app crashed (in debug mode at least). Now, everything is fine again.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+###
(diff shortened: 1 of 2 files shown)


rainman110:
@jensschmeink Can you please verify the bug before and after the change?

jensschmeink:
> @jensschmeink Can you please verify the bug before and after the change?

I can reproduce the error in the old code but not with the changes anymore

jensschmeink (approved):

PR #1281, opened 2024-08-29
fix.: remove warnings for creation of multiple commands. These are recorded in the command stack

NitenIchiry:
remove warnings for creation of multiple commands. These are recorded in the command stack.
Thus the warnings are unneccessary.


rainman110 (approved):

NitenIchiry:
@rainman110 I also removed the PR template here. Will try to leave it in the future. Also there is no changelog here. Do we need one ?

rainman110:
> @rainman110 I also removed the PR template here. Will try to leave it in the future. Also there is no changelog here. Do we need one ?

No, as this is not directly "visible" for the user.

PR #1282, opened 2024-08-29
fix.: disabled switching project to same project while task is running

NitenIchiry:
Projects cannot be switched while task is running. This PR removes the possibility to switch to the same project while a task is running/does not show the project switch warning in this case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased 2.1]
+### Added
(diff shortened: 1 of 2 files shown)


rainman110 (approved):
Looks good to me

rainman110:
@NitenIchiry When creating PRs, please don't remove the default PR checklist for the reviewer. Normally, you should have seen it, when initially created the PR.

Changes that are apparent to the user, should also be added to the changelog.

Could you thus please extend the changelog.

NitenIchiry:
I am very sorry. I did not like the default check-list in this case.
But I will extend the checklist just now.

NitenIchiry:
I will merge if you have no issues with this.

rainman110:
No, please don't merge yourelf. I'd like to clean the git history first

rainman110:
Please avoid merging the master into a feature branch.

NitenIchiry:
Oh that happened because this was derived from the master. I did not think about it 😢

NitenIchiry:
Just thought there would not be another 2.0.X

rainman110:
Don't worry. This can all be fixed ;)

Issue #612, opened 2023-11-06
Unknown Task not shown


rainman110:
In GitLab by @jensschmeink on Nov 6, 2023, 11:13

While unknown calculator classes are represented in GTlab by a red line in the process view (see pic) unknwon task are not read at all and therefore not displayed.   
![grafik](https://my-gitlab-bucket.s3.amazonaws.com/727352741/53fd9392750a3499e8727808e061133a9f6d5ddca443aedf363187385eb6427c/grafik.png)  
The logging says "invalid task file"

rainman110:
@NitenIchiry Please ask @jensschmeink  who wrote this issue (I just converted it to github)

jensschmeink:
[...]

PR #1286, opened 2024-09-10
fix.: add task as dummy object to taskgroup if the task-module did no…

NitenIchiry:
…t load

GtTasks recognized as dummy are now added to the datamodel.

In `GtTaskGroup` 

## Description

In `GtTaskGroup` task-objects read from momento are now additionaly checked if they are dummy objects and then added to the datamodel as background objects with the recognizable red color of dummy objects.

Previously a wrong error message was printed that the task file is invalid. This happened because the GtDummy-object could not be parsed to GtTask.

https://github.com/dlr-gtlab/gtlab-core/issues/612

## Checklist:

### DIFF
--- a/src/core/process_management/gt_taskgroup.cpp
+++ b/src/core/process_management/gt_taskgroup.cpp
@@ -495,17 +495,7 @@ GtTaskGroup::Impl::createTaskFromFile(const QString& filePath) const
 
     auto obj = memento.restore(gtProcessFactory);
 
-    if (obj)
-    {
-        retval = qobject_cast<GtObject*>(obj);
-        if (!retval)
-        {
-            gtError() << QObject::tr("Invalid task file (%1)").arg(filePath);
-            delete obj;
-        }
-    }
-
-    return std::unique_ptr<GtObject>(retval);
+    return std::unique_ptr<GtObject>(obj);
 }
 
 bool
--- a/src/core/process_management/gt_taskgroup.cpp
+++ b/src/core/process_management/gt_taskgroup.cpp
@@ -43,7 +43,7 @@ public:
     QString path(const QString& projectPath,
                  const GtTaskGroup::SCOPE scope) const;
 
-    GtObject *createTaskFromFile(const QString& filePath) const;
+    std::unique_ptr<GtObject> createTaskFromFile(const QString& filePath) const;
 
     bool saveTaskToFile(const GtTask* task, const QString& groupPath) const;
 
@@ -124,17 +124,16 @@ GtTaskGroup::read(const QString& projectPath,
 
         if (newTask)
         {
-            if (newTask->isDummy())
+            if (newTask.get()->isDummy())
             {
                 gtDebug().medium().nospace() << "dummy task created ("
                                              << newTask->uuid() << ")";
-                appendChild(newTask);
+                appendChild(newTask.release());
                 continue;
             }
+            auto taskObject = gt::unique_object_cast<GtTask>(std::move(newTask));
 
-            auto retval = qobject_cast<GtTask*>(newTask);
-
-            if (!retval)
+            if (!taskObject)
             {
                 gtError() << QObject::tr("Invalid task file (%1)").arg(e.toString());
                 newTask->deleteLater();
@@ -142,7 +141,7 @@ GtTaskGroup::read(const QString& projectPath,
 
             gtDebug().medium().nospace() << "new task created ("
                                          << newTask->uuid() << ")";
-            appendChild(retval);
+            appendChild(taskObject);
         }
     }
 
@@ -455,7 +454,7 @@ GtTaskGroup::Impl::path(const QString& projectPath,
     return _pub.get().groupPath(projectPath, scope, _pub.get().objectName());
 }
 
-GtObject*
+std::unique_ptr<GtObject>
 GtTaskGroup::Impl::createTaskFromFile(const QString& filePath) const
 {
     QFile taskFile(filePath);
@@ -493,7 +492,7 @@ GtTaskGroup::Impl::createTaskFromFile(const QString& filePath) const
 
     auto obj = memento.restore(gtProcessFactory);
 
-    return obj;
+    return std::make_unique<GtObject>(obj);
 }
 
 bool
--- a/src/core/process_management/gt_taskgroup.cpp
+++ b/src/core/process_management/gt_taskgroup.cpp
@@ -43,7 +43,7 @@ public:
     QString path(const QString& projectPath,
                  const GtTaskGroup::SCOPE scope) const;
 
-    std::unique_ptr<GtTask> createTaskFromFile(const QString& filePath) const;
+    GtObject *createTaskFromFile(const QString& filePath) const;
 
     bool saveTaskToFile(const GtTask* task, const QString& groupPath) const;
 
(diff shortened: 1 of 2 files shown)


NitenIchiry:
@rainman110 I was not entirely sure how to implement this change. A different idea of me was to add a new function which checks the task-items for dummy objects prior to the `createTaskFromFile` function.

Let me know what you would prefer in this case. The error is fixed in this case. But there is still an error scenario when the GtObject returned by createTaskFromFile returns a nullpointer. Also the name of the function should be changed with the current solution.

rainman110 (changes requested):
In principle looks good, but please keep the unique_ptr api.

In line 135, you could use the function `gt::unique_object_cast`, i.e.

auto taskObject = gt::unique_qobject_cast<GtTask>(std::move(newTask));

rainman110:
> @rainman110 I was not entirely sure how to implement this change. A different idea of me was to add a new function which checks the task-items for dummy objects prior to the `createTaskFromFile` function.

I think, the general implementation is fine, i.e. adding a dummy object to the process tree.

NitenIchiry:
> In principle looks good, but please keep the unique_ptr api.
> 
> In line 135, you could use the function `gt::unique_object_cast`, i.e.
> 
> auto taskObject = gt::unique_qobject_cast(std::move(newTask));

I cannot get the unique object cast to work as it was described here. Do you have a typo anywhere?

src/core/process_management/gt_taskgroup.cpp
```diff
@@ -439,11 +455,11 @@ GtTaskGroup::Impl::path(const QString& projectPath,
     return _pub.get().groupPath(projectPath, scope, _pub.get().objectName());
 }
 
-std::unique_ptr<GtTask>
+GtObject*
```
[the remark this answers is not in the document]
NitenIchiry: I reimplemented the unique pointer. Let me know what you think.

src/core/process_management/gt_taskgroup.cpp
```diff
@@ -120,13 +120,29 @@ GtTaskGroup::read(const QString& projectPath,
...
         {
+            if (newTask->isDummy())
+            {
+                gtDebug().medium().nospace() << "dummy task created ("
+                                             << newTask->uuid() << ")";
+                appendChild(newTask);
+                continue;
+            }
+
+            auto retval = qobject_cast<GtTask*>(newTask);
```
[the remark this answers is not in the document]
NitenIchiry: Changed the name

src/core/process_management/gt_taskgroup.cpp
```diff
@@ -120,13 +120,29 @@ GtTaskGroup::read(const QString& projectPath,
...
+
+            if (!retval)
+            {
+                gtError() << QObject::tr("Invalid task file (%1)").arg(e.toString());
+                newTask->deleteLater();
```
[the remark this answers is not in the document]
NitenIchiry: Well unique_objects cannot be deleted with delete, but with deleteLater (ownership is transfered). I changed it back to delete and release the object now. Let me know what you think.

NitenIchiry:
It is ready for a final assessment.

src/core/process_management/gt_taskgroup.cpp
```diff
@@ -120,13 +121,29 @@ GtTaskGroup::read(const QString& projectPath,
...
+            }
+
+            std::unique_ptr<GtTask> taskObject = gt::unique_qobject_cast<GtTask>(std::move(newTask));
+
+            if (!taskObject)
+            {
+                gtError() << QObject::tr("Invalid task file (%1)").arg(e.toString());
+                delete newTask.release(); // release and delete unused object newTask
+            }
+
```
rainman110, line 143: This should do much less imho

1) The manual delete should not be necessary
2) Why do you treat dummy objects similar than other objects

So here,'s what I'd do

for each task
 - newTask = createFromFile...
 - if (!newTask) continue;
 - if (task->isDummy) printWarning;
 - else assert (task type is GtTask)
 - appendChild(newTask.release)

No need for ownership transfer etc.

rainman110: Theoretically, the memento restore should already warn, that a dummy task was created, so the warning could be probably also removed

rainman110 (changes requested):

src/core/process_management/gt_taskgroup.cpp
NitenIchiry: `WARN  [11:03:11] Creating dummy object for unknown class 'GtpyTask'.
DEBUG [11:03:11] [Core] dummy task created ({dde42211-6bfa-4136-87c6-c7f2a56be5ea})
DEBUG [11:03:11] [Core] new task created ({8d9b8f01-b92f-409e-97e9-84be68878d9a})`

Currently A Warning and Debug message is printed. Should I remove the dummy task created message ?

NitenIchiry: I added an assert. Was it your idea to add it like this, or did you have a different idea ?

NitenIchiry:
Screenshot 2024-10-17 110328
You cannot press the button to execute tasks which contain dummy bjects atm. We should add a new issue for this in which we add a new type of task a `GtDummyTask` or something similar. That was what we discussed, wasnt it ?

rainman110:
> You cannot press the button to execute tasks which contain dummy bjects atm. We should add a new issue for this in which we add a new type of task a `GtDummyTask` or something similar. That was what we discussed, wasnt it ?

I think this is a good default behaviour for now. Nothing to do on your side. What we need to implement is to make it skippable (in a separate issue)

NitenIchiry:
I removed the unneeded check for GtObjects in the `createTaskFromFile` function

src/core/process_management/gt_taskgroup.cpp
```diff
@@ -439,11 +456,12 @@ GtTaskGroup::Impl::path(const QString& projectPath,
...
 }
 
-std::unique_ptr<GtTask>
+std::unique_ptr<GtObject>
 GtTaskGroup::Impl::createTaskFromFile(const QString& filePath) const
 {
     QFile taskFile(filePath);
-    GtTask* retval = nullptr;
+
+    GtObject* retval = nullptr;
```
rainman110, line 464: As far as i can see, retval is not used anymore.

src/core/process_management/gt_taskgroup.cpp
```diff
@@ -122,12 +123,28 @@ GtTaskGroup::read(const QString& projectPath,
...
+
+        if (newTask.get()->isDummy())
+        {
+            gtDebug().medium().nospace() << "dummy task created ("
+                                         << newTask->uuid() << ")";
+        }
+        else
+        {
+            GtTask* gtTask = dynamic_cast<GtTask*>(newTask.get());
+            assert(gtTask != nullptr && "newTask must be of type GtTask.");
```
rainman110, line 140:
```suggestion
            assert(qobject_cast<GtTask*>(newTask.get()) && "newTask must be of type GtTask.");
```

This way, the cast is only performed in debug mode.

NitenIchiry:
Thank you for the fast review. It should be fine now.

rainman110:
One more small thing, sorry i should have addressed this earlier: Please add a line to the changelog, that dummy tasks are now displayed in the process explorer 

(we only add to the changelog, if something user-visible has changed).

Also please remove the draft status from the PR.

rainman110 (approved):

PR #1287, opened 2024-09-11
Deprecate unused run function of GtTask

jensschmeink:
The function was not used and therefore could be removed

## Checklist:

### DIFF
--- a/src/core/process_management/gt_task.cpp
+++ b/src/core/process_management/gt_task.cpp
@@ -186,42 +186,42 @@ GtTask::exec()
     return true;
 }
 
-void
-GtTask::run(GtAbstractRunnable* r)
-{
-    if (!r)
-    {
-        setState(GtProcessComponent::FAILED);
-        gtError() << tr("%1: Failed to run task, invalid runnable!")
-                         .arg(objectName());
-        return;
-    }
-
-    setState(GtTask::RUNNING);
-    pimpl->dataToMerge.clear();
-
-    setRunnable(r);
-
-    QThreadPool* tp = QThreadPool::globalInstance();
-
-    if (!tp)
-    {
-        return;
-    }
-
-    runnable()->setAutoDelete(false);
-
-    connect(runnable().data(), &GtAbstractRunnable::runnableFinished,
-            this, &GtTask::handleRunnableFinished);
-
-    tp->start(runnable());
-
-    qDebug() << "#### exec event loop...";
-
-    pimpl->eventLoop.exec();
-
-    qDebug() << "#### exec finished";
-}
+//void
+//GtTask::run(GtAbstractRunnable* r)
+//{
+//    if (!r)
+//    {
+//        setState(GtProcessComponent::FAILED);
+//        gtError() << tr("%1: Failed to run task, invalid runnable!")
+//                         .arg(objectName());
+//        return;
+//    }
+//
+//    setState(GtTask::RUNNING);
+//    pimpl->dataToMerge.clear();
+//
+//    setRunnable(r);
+//
+//    QThreadPool* tp = QThreadPool::globalInstance();
+//
+//    if (!tp)
+//    {
+//        return;
+//    }
+//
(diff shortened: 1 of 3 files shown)


src/core/process_management/gt_task.cpp
```diff
@@ -186,42 +186,42 @@ GtTask::exec()
...
+//            this, &GtTask::handleRunnableFinished);
+//
+//    tp->start(runnable());
+//
+//    qDebug() << "#### exec event loop...";
+//
+//    pimpl->eventLoop.exec();
+//
+//    qDebug() << "#### exec finished";
+//}
```
jensschmeink, line 224:
```suggestion
```

rainman110:
@jensschmeink How do you know that the function is not used? It could be used by any module. I don't think it is a good idea to remove the function, as this will break the API.

jensschmeink:
> @jensschmeink How do you know that the function is not used? It could be used by any module. I don't think it is a good idea to remove the function, as this will break the API.
I basically agree with you. I have not found any usage after a short research and I think this function is missleading as it is not part of the designed process management.
Shall we mark it as depricated?

rainman110:
- Deprecating it is a good idea
- You could ask @OliverKunc , whether he is using this function. I known that BT are manually running calculators outside the workflow logic using the run function, hence we don't really know whether this function is also used.

src/core/process_management/gt_task.cpp
```diff
@@ -186,42 +186,6 @@ GtTask::exec()
...
-            this, &GtTask::handleRunnableFinished);
-
-    tp->start(runnable());
-
-    qDebug() << "#### exec event loop...";
-
-    pimpl->eventLoop.exec();
-
-    qDebug() << "#### exec finished";
-}
```
jensschmeink, line 189:
```suggestion
void
GtTask::run(GtAbstractRunnable* r)
{
    if (!r)
    {
        setState(GtProcessComponent::FAILED);
        gtError() << tr("%1: Failed to run task, invalid runnable!")
                         .arg(objectName());
        return;
    }

    setState(GtTask::RUNNING);
    pimpl->dataToMerge.clear();

    setRunnable(r);

    QThreadPool* tp = QThreadPool::globalInstance();

    if (!tp)
    {
        return;
    }

    runnable()->setAutoDelete(false);

    connect(runnable().data(), &GtAbstractRunnable::runnableFinished,
            this, &GtTask::handleRunnableFinished);

    tp->start(runnable());

    qDebug() << "#### exec event loop...";

    pimpl->eventLoop.exec();

    qDebug() << "#### exec finished";
}

```

src/core/process_management/gt_task.h
```diff
@@ -54,11 +54,6 @@ class GT_CORE_EXPORT GtTask : public GtProcessComponent
      */
     bool exec() override;
 
-    /**
-     * @brief run
-     */
-    void run(GtAbstractRunnable* runnable);
-
     /**
```
jensschmeink, line 57:
```suggestion
    [[deprecated("This function is not part of the supported process management and therefore not 
                          maintained and updated")]]
    /**
     * @brief run
     */
    void run(GtAbstractRunnable* runnable);

     /**
```

jensschmeink:
@rainman110 I really like your version dependent deprecated function 👍🏼

jensschmeink:
> * Deprecating it is a good idea
> 
>     * You could ask @OliverKunc , whether he is using this function. I known that BT are manually running calculators outside the workflow logic using the run function, hence we don't really know whether this function is also used.

I Talked to @OliverKunc: They do not use this function and are fine if it is deprecated

rainman110:
> @rainman110 I really like your version dependent deprecated function 👍🏼

Thank you. My intention was, that we can simple search for `deprecated_from(3, 0 ...` when we are switching. I actually wanted to make the compile fail, when we have set the gtlab version to 3.0, but i did not succeed :( .

rainman110 (approved):

PR #1292, opened 2024-09-17
Add module interface documentation

NitenIchiry:
Add module interface documentation and add link to documentation in GtModuleInterface.

## Checklist:


rainman110:
@NitenIchiry Great! Thank you for you work. Could you please upload the missing image? I think you just forgot to add  it to git.

I will keep the issue open such that we can continue working on the documentation in this branch.

NitenIchiry:
Yes, that would be ideal. DO you suggestions for updating the .gitignore ? The path docs/modules was ignored because of the field /modules in the ignore file.

rainman110:
> Yes, that would be ideal. DO you suggestions for updating the .gitignore ? The path docs/modules was ignored because of the field /modules in the ignore file.

Just add `!docs/modules/`

NitenIchiry:
Done. I guess we can merge this now.

NitenIchiry:
> As is have written above, let's keep the MR open to add more documentation before merging 😉 
> 
> Thanks for your work !

Yeah good point, keeping the issue open doesn't make a lot of sense 🙂.

NitenIchiry:
@rainman110 I added documentation for exporter and importer interfaces. Maybe you can look over these and correct any mistakes I made. I also made afirst version for the mdi documentation. But I was not really sure about all of the virtual function. So you can improve on my first outline. I did not get to write about all of the interfaces. Mostly because there also is not any documentation yet in the wiki.

rainman110:
@jensschmeink I think we should keep tasks and calculators on one page.

docs/basics/data_modelling.rst
```diff
@@ -1,15 +1,111 @@
...
+To perform a recursive query, including sub-objects, replace ``findDirectChild`` with ``findChild`` etc.
+
+Query child objects  by UUID or path
+""""""""""""""""""""""""""""""""""""
+
+To retrieve objects using unique identifiers (UUIDs) or hierarchical paths using both direct and recursive searches,
+the functions :cpp:func:`GtObject::getObjectByUuid()` and :cpp:func:`GtObject::getObjectByPath()` are used.
+
+
+Adding child objects
```
jensschmeink, line 59: @rainman110 We have to keep in mind (and maybe should mention), that there are cases in which the gtDatamodel has to be used to add a child correctly.

rainman110: Good point. I have no clue, when it should be used though and when not. Maybe @real-ct-ac can clarify?

jensschmeink:
> @jensschmeink I think we should keep tasks and calculators on one page.

done

docs/basics/data_modelling.rst
jensschmeink: It is always needed if the add is called in the GUI  (as in UI elements) and not via calculator, importer

rainman110: What is the difference? I am sure I use appendChild also from GUI commands (e.g. when importing something) and it also seems to work.

jensschmeink: I think the main difference is the update of the model in the GUI.

rainman110: Is it safe to just always use the datamodel->appendChild() function? If so, we should only document this.

jensschmeink: No. In calculators it should not be used.

PR #1296, opened 2024-09-18
Deprecate unique object name

jensschmeink:
Deprecation of the unique object name function in the core datamodel to reduce access to the datamodel instance if it is not needed

## Checklist:


rainman110 (approved):
Nice!

PR #1299, opened 2024-09-25
Refactored the toolbar to be more modular

rainman110:
## Description
This PR refactors the toolbar to be modular.  This support the following features:
- Dynamically add action groups
- Dynamically change visibility of groups and actions
- Dynamically remove actions from groups
- Add actions as status icons (TODO: no interface for modules yet defined)
- Added the `GtMdiItem::addToolbarAction`  to populate the editor context of the toolbar
- Pimpled the `GtMdiItem` as much as possible (there is the friend MdiLauncher) to make the class more flexible to change
- Extended the MdiExtInterface example to demonstrate the features

__Please note:__
 - the ABI of  `GtMdiItem` has changed, hence I updated the interface version accordingly.
 - the toolbar animations are currently removed. I am no qml expert. Please feel free to add these again.
 - I have created a new `QmlAction` class. Should we though use `GtObjectUIAction` instead and adapt it (we need a icon url)?

## How Has This Been Tested?

In long sessions at night 😏 .  Please note, the previous toolbar is completely removed! Also, have a look at the mdi_interface_ext example module.

Screenshot, with custom toolbar actions through the ext viewer:
image

When switching to the normal viewer, the toolbar items vanish:
image

## Checklist:


src/gui/gt_mdiitem.cpp
```diff
@@ -18,12 +18,48 @@
...
+        foreach (GtQueuedMdiEvent* e, m_eventQueue)
+        {
+            if (T* ce = qobject_cast<T*>(e))
+            {
+                m_eventQueue.removeOne(e);
+                return ce;
+            }
+        }
+
+        return NULL;
```
mariusalexander, line 55: I think this should work and avoids the copy of foreach. But I'm not 100% certain.
 ```suggestion
        auto iter = std::find_if(m_eventQueue.begin(), 
                                 m_eventQueue.end(), 
                                 [](GtQueuedMdiEvent* e){
            return qobject_cast<T*>(e);
        });
        if (iter != m_eventQueue.end()) return nullptr;
        
        T* ce = static_cast<T*>(*iter);
        m_eventQueue.erase(iter);
        return ce;
```

File: src/gui/gt_mdiitem.cpp [hunk 2]
```diff
@@ -101,6 +135,12 @@ GtMdiItem::isPrintable() const
     return false;
 }
 
+std::vector<GtQmlAction *>
+GtMdiItem::toolbarActions() const
+{
+    return pimpl->m_toolbarActions;
+}
```
mariusalexander, line 142: Since the function is const, shouldn't this return a vector with const elements? Also, I am unsure whether we should use `std::vector` or `QVector` here and whether we should return it as a copy or reference (A `QVector` should be cheaper to shallow copy?). 

For such use cases it would be neat to return a more abstract range-like object, like a span just that it does not need to care about the underlying container. That does not heap allocate on copy and is also ABI safe-ish. But I might very well be overthinking this 😂 This function might very well be not called that often anyways. 
```suggestion
std::vector<const GtQmlAction*>
GtMdiItem::toolbarActions() const
{
    return gt::const_container_cast(pimpl->m_toolbarActions);
}
```

src/gui/gt_mdilauncher.cpp
```diff
@@ -411,6 +411,22 @@ GtMdiLauncher::isPrintable(QWidget* subWindow) const
     return false;
 }
 
+std::vector<GtQmlAction*>
```
mariusalexander, line 414: same here regarding constness
```suggestion
std::vector<const GtQmlAction*>
```

src/gui/gt_qmlaction.h
mariusalexander, line 1: A few more comments would be nice, even though all functions and members are self-explanatory

src/gui/gt_qmlaction.h
```diff
@@ -23,35 +25,44 @@ class GT_GUI_EXPORT GtQmlAction : public QObject
...
     QString text() const;
     void setText(const QString& text);
 
     QUrl iconUrl() const;
-    void setIconUrl(const QUrl&url);
+    void setIconUrl(const QUrl& url);
+
+
+    QString tooltip() const;
+    void setTooltip(QString t);
```
mariusalexander, line 47: Just as a suggestion (as I have noticed this recently): Qt uses `toolTip` throughout its API. I propose we should stick to the same naming convetion (even though I would have named the function all lower case ass well intuitively). I'll mark only these functions here, of couse this applies to all symbols named `tooltip`.
```suggestion
    QString toolTip() const;
    void setToolTip(QString t);
```

src/gui/gt_mdiitem.cpp
rainman110: This was moved code into the pimpl.

src/gui/gt_mdiitem.cpp
rainman110: The non-constness is by design. Since the toolbar potentially interacts in a non-const way with the actions. Slots are called and so on.

src/gui/gt_qmlaction.h
rainman110: Good suggestion. I will change it.

src/gui/tools/gt_qmltoolbar.h
```diff
@@ -0,0 +1,60 @@
...
+
+    Q_INVOKABLE QVariantListModel* toolbarGroups();
+
+    Q_INVOKABLE void addToolbarGroup(GtQmlToolbarGroup* group);
+    Q_INVOKABLE void removeToolbarGroup(GtQmlToolbarGroup * group);
+
+    Q_INVOKABLE void addStatusAction(GtQmlAction* action);
+
+    Q_INVOKABLE bool darkmode() const;
+    Q_INVOKABLE void setDarkmode(bool d);
```
mariusalexander, line 45: nitpick: I *think* we use `darkMode` throughout the application instead of `darkmode` (please verify) and decide whether to update the function name accordingly
```suggestion
    Q_INVOKABLE bool darkMode() const;
    Q_INVOKABLE void setDarkMode(bool d);
```

src/gui/tools/gt_qmltoolbar.h
mariusalexander, line 1: Please also add a few more comments for the API documentation

src/gui/tools/gt_qmltoolbar.cpp
```diff
@@ -0,0 +1,120 @@
...
+{
+    return &pimpl->m_toolbarGroupsList;
+}
+
+void
+GtQmlToolbar::addToolbarGroup(GtQmlToolbarGroup* group)
+{
+
+    if (!group)
+        return;
```
mariusalexander, line 61: I think this is our coding convection for such if-statements
```suggestion
    if (!group) return;
```

src/gui/tools/gt_qmltoolbar.cpp
```diff
@@ -0,0 +1,120 @@
...
+GtQmlToolbar::darkmode() const
+{
+    return pimpl->m_darkMode;
+}
+
+void
+GtQmlToolbar::setDarkmode(bool d)
+{
+    if (d == pimpl->m_darkMode)
+        return;
```
mariusalexander, line 104: See comment above (Btw here you have used `darkMode` ;) )
```suggestion
    if (d == pimpl->m_darkMode) return;
```

src/gui/tools/gt_qmltoolbar.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+#include <QLayout>
+#include <QQuickWidget>
+#include <QQmlContext>
+
+#include <iterator>
+
+struct GtQmlToolbar::Impl
+{
+    QVariantListModel m_toolbarGroupsList;
```
mariusalexander, line 20: Why did you chose to use a QVariant as value type? I could only see that you append  `GtQmlToolbarGroup*`

src/gui/gt_mdiitem.cpp
rainman110: In fact, we could simply return it as a const reference. The action vector is already part of the mdi item.

src/gui/tools/gt_qmltoolbargroup.h
```diff
@@ -0,0 +1,68 @@
...
+#ifndef GTQMLTOOLBARGROUP_H
+#define GTQMLTOOLBARGROUP_H
+
+#include <gt_qmlaction.h>
+
+#include <gt_listmodel.h>
+
+#include <gt_gui_exports.h>
+
+using ActionModel = GtListModel<GtQmlAction*>;
```
mariusalexander, line 18: Maybe we should use a more descriptive typename
```suggestion
using GtQmlActionModel = GtListModel<GtQmlAction*>;
```

src/gui/tools/gt_qmltoolbargroup.h
mariusalexander, line 1: Same here, a few more comments for the API documentation. Please also check the other header files for missing comments

src/gui/tools/gt_qmltoolbargroup.cpp
```diff
@@ -0,0 +1,94 @@
...
+    // if all items are visible, it is also not visible
+
+
+    bool oneIsVisible =
+        std::any_of(std::begin(m_data), std::end(m_data),
+                    [](GtQmlAction* action)
+    {
+        if (!action) return false;
+        return action->visible();
+    });
```
mariusalexander, line 59:
```suggestion
                    [](GtQmlAction* action) {
        return action && action->visible();
    });
```

src/gui/tools/gt_qmltoolbargroup.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+    return m_visible && oneIsVisible;
+}
+
+void
+GtQmlToolbarGroup::setVisible(bool visible)
+{
+    if (m_visible == visible)
+        return;
```
mariusalexander, line 68:
```suggestion
    if (m_visible == visible) return;
```

src/gui/tools/gt_qmltoolbargroup.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+    m_visible = visible;
+    emit visibleChanged();
+}
+
+bool
+GtQmlToolbarGroup::append(GtQmlAction* action)
+{
+    if (!action)
+        return false;
```
mariusalexander, line 78:
```suggestion
    if (!action) return false;
```

src/gui/tools/gt_qmltoolbargroup.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+    auto success = GtListModel::append(action);
```
mariusalexander, line 80:
```suggestion
    bool success = GtListModel::append(action);
```

src/resources/qml/ToolbarGroup.qml
```diff
@@ -0,0 +1,59 @@
...
+
+    visible: true
+
+    Rectangle {
+        id: rect
+        implicitWidth: row.width + 10
+        implicitHeight: 40
+
+        visible: toolbarGroup.visible && toolbarGroup.count > 0
+        color: dark_mode ? "#0a111f" : "#ffffff"
```
mariusalexander, line 25: Is it possible to use predefined colors from `gt::gui::color` or `QPalette`? That way we can more easily swap themes in the future.

src/app/tools/gt_maintoolbar.h
```diff
@@ -0,0 +1,52 @@
...
+    // System context
+    GtQmlAction *homeAction, *newProjectAction, *openProjectAction;
+
+    // project context
+    GtQmlAction *saveProjectAction, *undoAction, *redoAction, *infoAction;
+
+    // edit actions
+    GtQmlAction* printAction;
+
+    GtQmlToolbarGroup *projectContext, *editorContext;
```
mariusalexander, line 41: Consider adding default initializations
```suggestion
    GtQmlAction *homeAction{}, *newProjectAction{}, *openProjectAction{};

    // project context
    GtQmlAction *saveProjectAction{}, *undoAction{}, *redoAction{}, *infoAction{};

    // edit actions
    GtQmlAction* printAction{};

    GtQmlToolbarGroup *projectContext{}, *editorContext{};
```

src/app/tools/gt_maintoolbar.cpp
```diff
@@ -0,0 +1,194 @@
...
+GtMainToolbar::GtMainToolbar(GtMainWin *parent)
+    : GtQmlToolbar(parent)
+{
+    auto tbSystemContext = new GtQmlToolbarGroup("System", this);
+
+    homeAction
+        = new GtQmlAction(tr("Home"), QUrl("/icons/home.svg"), this);
+
+    newProjectAction
+        = new GtQmlAction(tr("New Project"), QUrl("/icons/projectAdd.svg"),
```
mariusalexander, line 24: can you use `gt::gui::icon::url(...)` here?

mariusalexander (commented):
I have finished my review but have not checked out the code and tested it myself. Maybe I'll find time in the next days. But the code looks good to me apart from minor issue!

src/gui/gt_mdiitem.cpp
mariusalexander: I see, but maybe you could touch it up a little while you are at it? Or should we leave it as is?

src/gui/gt_mdiitem.cpp
mariusalexander: Ah I see

src/gui/tools/gt_qmltoolbar.cpp
rainman110: Because it makes the interaction with QtQuick very easy.

src/resources/qml/ToolbarGroup.qml
rainman110: Yes. I'd definetly like to do this. But I don't want to make the PR much bigger. Lets do this in a separate issue.

src/app/tools/gt_maintoolbar.cpp
rainman110: Good catch. The function was made later than writing these lines.

src/gui/gt_mdiitem.cpp
rainman110: I have thought about it. Your solution does not work and you need to cast twice, which makes the code more complicated as before. We can still avoid the copy by using a normal for loop instead of a foreach

rainman110:
@mariusalexander I integrated your comments... mostly.

I did not change:
 - constness of actions getters, as i think they will be sementically used in a non-const way
 - return actions by range (instead of a plain vector). This function will be only called, when switching between mdi items, so this would be somewhat over engineered ;)
 - Color theme integration. I'd like to do this in a separate PR.

mariusalexander (approved):
Looks good to me! Could you also open an issue (if there isn't one already) for the theming of the QML Toolbar?

rainman110:
> Looks good to me! Could you also open an issue (if there isn't one already) for the theming of the QML Toolbar?

Done, see #1309

PR #1306, opened 2024-10-10
Try to use QString instead of keysequence to avid Gui dependency

jensschmeink:
The core has a dependency to the QtGUI library only because of the usage of the QKeySequence to store ShortCut information.

This is changed in this PR. The GtShortCutSettingsData now stored the key information in a QString. This is an ABI change!
And as also the constructor changes this is also an API change!

First overview in our modules showed there would be the need in 2 modules to make simple adaptions but external module developers might be affected as well if they use short cuts with this old interface.

As the transfer from keySequence to string and vice versa could lead to wrong usage the interface was extended to avoid the usage of the  GtShortCutSettingsData for module developers but use interface functions still have the option to use keysequences to store the used keys.

## How Has This Been Tested?
First a GUI test failed as one feature had been missing but now all tests are successfull and test usage of few short cuts lead to the same behavior as before. 

## Checklist:

### DIFF
--- a/src/gui/gt_application.h
+++ b/src/gui/gt_application.h
@@ -317,7 +317,7 @@ public:
      */
     void extendShortCuts(const QString& id,
                          QString const& category,
-                         const QKeySequence k,
+                         const QKeySequence& k,
                          bool readOnly = false);
 
     /**
--- a/src/gui/gt_application.h
+++ b/src/gui/gt_application.h
@@ -306,6 +306,9 @@ public:
      * @brief extendShortCuts
      * @param shortcut to add
      */
+    [[deprecated("Use extendShortCuts(const QString& id, "
+                 "QString const& category, const QKeySequence k, "
+                 "bool readOnly = false) instead")]]
     void extendShortCuts(const GtShortCutSettingsData &shortcut);
 
     /**
--- a/src/gui/gt_application.h
+++ b/src/gui/gt_application.h
@@ -309,6 +309,15 @@ public:
     void extendShortCuts(const GtShortCutSettingsData &shortcut);
 
     /**
+     * @brief extendShortCuts
+     * @param shortcut to add
+     */
+    void extendShortCuts(const QString& id,
+                         QString const& category,
+                         const QKeySequence k,
+                         bool readOnly = false);
+
+    /**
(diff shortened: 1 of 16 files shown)


jensschmeink:
This PR leads to an ABI/API change and leads to the need of small adaptions in few modules concerning the registration of short cuts

src/gui/gt_application.h
```diff
@@ -306,8 +306,20 @@ class GT_GUI_EXPORT GtApplication : public GtCoreApplication
...
+    [[deprecated("Use extendShortCuts(const QString& id, "
+                 "QString const& category, const QKeySequence k, "
+                 "bool readOnly = false) instead")]]
     void extendShortCuts(const GtShortCutSettingsData &shortcut);
 
+    /**
+     * @brief extendShortCuts
+     * @param shortcut to add
+     */
+    void extendShortCuts(const QString& id,
```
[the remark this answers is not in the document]
jensschmeink: Done

rainman110:
@jensschmeink Please fill in the description page above. (What has changed, why, ABI changes etc!)

src/gui/gt_dockwidget.h
```diff
@@ -77,12 +77,16 @@ class GT_GUI_EXPORT GtDockWidget : public QDockWidget
      * @param data - short cut data to register
      * @return the key sequence used for this id
      */
+    [[deprecated("Use registerShortCut(QString const& id,"
```
rainman110, line 80: Can you please explain, why you have deprecated these functions? I see no need to do so, did I oversee anything?

rainman110 (commented):
Look generally good to me. I just don't understand, why you  have deprecated so many functions, e.g. those in MDIitem or  in DockWidget.

src/gui/gt_dockwidget.h
jensschmeink: The idea is to avoid the usage of the GtShortCutSettingsData class outside the core as the the interface with the QString for the key sequence might be used wrong. In my opinion in GTlab version > 2 the GtShortCutSettingsData should be kept internal in GTlab and not be usable anymore and only offer good interface functions.

rainman110 (approved):

PR #1307, opened 2024-10-10
Gtobjectui regexp validation using objects

jensschmeink:
Currently the validation for renaming of ObjectUI elements have no access to the underlying object. This is added here with new interface functions.

A new kind of regular expression is added to be not so strict with intermediate states of the renaming. It is allowed to type everything and only in the end this is checked.
As this allows more freedom it is also hared to detect problems while renaming.
Therefore it is up the the programmer to use this new system or a version which is closer to the old behavior by a flag.

And additionally the objectUI can also define a hint how the regular expression limits the renaming as regular expressions are not readable for users that good.

## How Has This Been Tested?
It has been tested by a implementation in the performance module to use this new features and in some basic UI elements to check if the behavior keeps the same for the other classes.

## Hint for reviwers how to test usage:
The test modules of GTlab should be build and a project using the testing modules ("TestDatamodelInterface") should be opened.
- Select an object of the class "TestDmiExternalObject"
- Open the context menu there and select "Rename"
- The renaming is restricted to the reg exp for letters, numbers few brackets and spaces. So if you use a sign like a "." the text should become orange and should not be able to be used for the renaming. 
- For existing implementations nothing should be changed while renaming but the hint in the logging widget with the hint which restrictions are applied if an invalid sign is used.

## Checklist:


rainman110:
@jensschmeink Are you finished editing? If yes, please provide some context and description for the MR. Cppcheck is also still failing.

jensschmeink:
This changes needs discussion: For a more user friendly usage of the renaming I also added the registration of hints for the renaming as the regular expressions are hard to understand for normal users. Additional there is the flag to switch between the old behavior not to allow invalid intermediate strings or the new one which is only checked in the end.

But this results in new interface functions in the GtObjectUI which might not be desired.

rainman110:
@jensschmeink Could please fill out the description page etc of this PR.

jensschmeink:
Use new functions as registered functions and not as new virtual functions. Best case would be to add them to a pimpl to avoid further changes later.
  
Additionally add a usage in one of the testing modules

jensschmeink:
Add a color change while renaming if the string is not valid

src/gui/gt_textfilterdelegate.cpp
```diff
@@ -33,20 +35,19 @@ GtTextFilterDelegate::createEditor(QWidget* parent,
...
+    bool checkWhileEditing = true;
+    QString hint = gt::re::onlyLettersAndNumbersHint();
+
+    if (GtProject* proj = gtApp->currentProject())
     {
-        GtObject* obj = proj->getObjectByUuid(uuid);
+        QString uuid = index.data(GtCoreDatamodel::UuidRole).toString();
 
-        if (obj)
+        if (GtObject* obj = proj->getObjectByUuid(uuid))
```
rainman110, line 50: Suggestion: The if is pretty much nested. Please extract the whole if into a separate function or private method, returning the regexp from the object.

rainman110 (changes requested):
I have some small suggestions, see comments.

I don't know though how I can test this using the mdi ext module. Please provide some testing steps.

rainman110:
@jensschmeink I think this could also be added to the changelog

rainman110 (approved):

PR #1317, opened 2024-10-25
Replace QValidator by RegExpPattern-String

jensschmeink:
Usage of QValidator is the last reason for the core libraries (DataProcessor and Core) to have a dependency to QtGUI.

## Description
The QValidator is replaced by a string to hold the validator pattern.
As it is part of the GtStringProperty the needed changes leads to ABI and API breaks.

## How Has This Been Tested?
Manual in the GTlab GUI

## Checklist:

### DIFF
--- a/src/dataprocessor/property/gt_stringproperty.h
+++ b/src/dataprocessor/property/gt_stringproperty.h
@@ -15,6 +15,7 @@
 #include "gt_datamodel_exports.h"
 
 #include "gt_property.h"
+#include "gt_regexp.h"
 
 class GT_DATAMODEL_EXPORT GtStringProperty : public GtProperty<QString>
 {
@@ -39,13 +40,18 @@ public:
      * @param name
      * @param brief
      * @param value
+     * @param validationPattern - this pattern can be set for validation of
+     * vale manipulation. The default value is the GTRegExp "forExpressions"
+     * which means allowed are letter, numbers and a list of special
+     * characters which are related to usage
+     * in mathematical descriptions: _ - + ^ ° / * . , ( ) [ ]
      */
     GtStringProperty(const QString& ident,
                      const QString& name,
                      const QString& brief,
                      const QString& value = QString(),
                      const QString& validationPattern =
-            "[A-Za-z0-9\\_\\-\\+\\^\\°\\/\\*\\.\\,\\(\\)\\[\\]]*");
+                        gt::re::forExpressions().pattern());
 
     // operator overloads
     void operator+=(const QString& b);
--- a/src/dataprocessor/property/gt_stringproperty.h
+++ b/src/dataprocessor/property/gt_stringproperty.h
@@ -16,6 +16,7 @@
 
 #include "gt_property.h"
 #include "gt_regexp.h"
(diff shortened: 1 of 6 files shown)


jensschmeink:
@rainman110 This is a first base for a discussion. This was the simplest approach to remove the GUI dependency.
What do you think about this?

src/gui/dock_widgets/properties/items/gt_propertyitem.cpp
```diff
@@ -285,7 +285,8 @@ GtPropertyItem::editorWidget(QWidget* parent,
 
             if (s)
             {
-                validator = s->validator();
+                validator = new QRegExpValidator(QRegExp(s->validator()),
+                                                 lineEdit);
```
jensschmeink, line 289:
```suggestion
                if (!s->validator().isEmpty()) 
                    validator = new QRegExpValidator(QRegExp(s->validator()), lineEdit);
```

src/dataprocessor/property/gt_stringproperty.h
```diff
@@ -48,7 +44,8 @@ class GT_DATAMODEL_EXPORT GtStringProperty : public GtProperty<QString>
                      const QString& name,
                      const QString& brief,
                      const QString& value = QString(),
-                     QValidator* validator = nullptr);
+                     const QString& validationPattern =
+            "[A-Za-z0-9\\_\\-\\+\\^\\°\\/\\*\\.\\,\\(\\)\\[\\]]*");
```
rainman110, line 48: Minor: Please comment, what this regex is. Ideally, use a named regex from gt_regexp.h

Major: Do you know, how often this is used in the modules? We already have API changes due to the qkeysequence change, requiring source modifications of a module. This would be another breaking API change. I though need to think about, how we can maintain API compatibility, but still without requiring QValidator in the dataprocessor.

jensschmeink: In the master build I had to do 3 or 4 adaptions in the modules from TWK (Performance, IntelliGraph & Geometry).

I used now the regexp namespace to cleanup the usage

jensschmeink:
@rainman110 I applied your suggestion to use the QRegularExpression class for a clearer and modern interface. Please check again, if this is ok

src/dataprocessor/property/gt_stringproperty.h
```diff
@@ -43,12 +41,18 @@ class GT_DATAMODEL_EXPORT GtStringProperty : public GtProperty<QString>
      * @param name
      * @param brief
      * @param value
+     * @param validationPattern - this pattern can be set for validation of
+     * vale manipulation. The default value is the GTRegExp "forExpressions"
```
rainman110, line 45:
```suggestion
     * @param validationPattern - this regular expression can be set to validate the
     * manipulation of the value. The default value is the GTRegExp "forExpressions"
```

rainman110 (changes requested):
I just have a small suggestion for the docstring,

jensschmeink:
> I just have a small suggestion for the docstring,

done

rainman110 (approved):

PR #1321, opened 2024-11-13
Allow creation and renaming of custom task groups

marvinoe21:
## Checklist:

### DIFF
--- a/src/app/CMakeLists.txt
+++ b/src/app/CMakeLists.txt
@@ -97,6 +97,7 @@ set(headers
     dock_widgets/process/gt_processview.h
     dock_widgets/process/gt_processconnectioneditor.h
     dock_widgets/process/gt_processcomponentsettingsbutton.h
+    dock_widgets/process/gt_editablecombobox.h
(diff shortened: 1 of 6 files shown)


rainman110:
@marvinoe21 this pr seems to have been slipped through. Should this be integrated into 2.11?  Is this pr ready?

marvinoe21:
@rainman110 I fixed the issue where the user task group was not created when opening a new project (see [Line 265 of GtProcessData](https://github.com/dlr-gtlab/gtlab-core/blob/422f5293b90ab29eb6edb5ff4357fb2016aca13a/src/core/process_management/gt_processdata.cpp#L265)). It occurred because when switching to a new project the current project of `gtApp` is invalid. This makes it impossible to append an object via `gtDataModel->appendChild()`. Now I check if the current project is valid. If it is, the task group is appended via the data model to support redo/undo. Otherwise it is appended directly to the container.

rainman110 (approved):
Looks good to me. I have tested it and it seems to work fine.

PR #1326, opened 2024-11-27
Offer basic regular expression to restrict siblings names usage

jensschmeink:
## Description
A function is added to add siblings names to a regular expression.
A function like this is needed for different use cases in GTlab when it should be prohibited to name objects like its siblings.
As I implemented this now in two modules I think it should be offered in core to help developers.

## How Has This Been Tested?
The identical code has been used in the intelligraph module and leads to good results there

## Checklist:

### DIFF
--- a/src/dataprocessor/gt_regexp.h
+++ b/src/dataprocessor/gt_regexp.h
@@ -15,6 +15,8 @@
 
 #include "gt_datamodel_exports.h"
 
+#include "gt_object.h"
+
 /**
  * namespace for RegualrExpressions, used in GTlab
  */
@@ -119,6 +121,41 @@ QRegExp GT_DATAMODEL_EXPORT forSemVers();
  * "Textfiles (*.txt)"
  */
 GT_DATAMODEL_EXPORT const QRegExp& forFileDialogFilters();
+
+template <typename T>
+inline void restrictRegExpWithSiblingsNames(GtObject& obj,
+                                            QRegExp& defaultRegExp)
+{
+    GtObject* parent = obj.parentObject();
+
+    if (!parent) return;
+
+    QList<T*> siblings = parent->findDirectChildren<T*>();
+
+    if (siblings.isEmpty()) return;
+
+    QStringList names;
+
+    for (auto* s : qAsConst(siblings))
+    {
+        names.append(s->objectName());
+    }
+
+    names.removeAll(obj.objectName());
+
+    QString allNames = names.join("|");
+
+    QString pattern = "^";
+
+    for (QString name : names)
+    {
+        pattern += "(?!" + name + "$)";
+    }
+
+    pattern += defaultRegExp.pattern();
+
+    defaultRegExp = QRegExp(pattern);
+}
 
 } // namespace re
 


jensschmeink:
@rainman110 As I implemented such a function now multiple times I want to offer it in the core as others might need it, too. Please review these few lines, but no hurry as it is intended for 2-1

rainman110:
@jensschmeink Sorry for ignoring this PR so long. I just rediscovered it. Please provide some (multiple) unit tests for the function.

It could also make sense to split the function in two parts, with one core function signature 

    void restrictRegExpWithObjectSiblingsNames(QString childNames, QRegExp& defaultRegExp)

In this way, you could test it much easier.

rainman110:
@jensschmeink Please rebase on current master

jensschmeink:
@rainman110 please have another look. I added tests and defined a default type for the template argument (is that a good idea?).

tests/unittests/datamodel/test_gtregexp.cpp
```diff
@@ -30,3 +30,59 @@ TEST(RegExp, fileFilter)
...
+
+    QStringList list{"Hello", "World"};
+
+    gt::re::restrictRegExpWithObjectNames(list, re);
+
+    auto match = [](const auto& re, const auto& string) {
+        return re.indexIn(string) >= 0;
+    };
+
+    EXPECT_TRUE(match(re, "Python"));
```
rainman110, line 45: Can you please also add non-letter strings to the test, to check, if the normal regex part is really evaluated?

src/dataprocessor/gt_regexp.h
```diff
@@ -126,6 +128,49 @@ GT_DATAMODEL_EXPORT const QRegExp& forFileDialogFilters();
...
+ * @brief Modifies a given QRegExp to restrict usage of sibling objects of the given GtObject obj
+ * The template type T has to be defined as the class of the sibling object to restrict.
+ * Note: The template type must be const
+ * e.g. gt::re::restrictRegExpWithObjectSiblingsNames<const GtObject>(obj, defaultRegExp);
+ * @param obj - The object to find the siblings and in most use cases the object which should be renamed
+ * @param defaultRegExp - the basic regular expression to be extended with the siblings names to restrict.
+ * If the object obj has no parent or siblings the RegExp is not modified
+ */
+template <typename T = const GtObject>
+inline void restrictRegExpWithObjectSiblingsNames(GtObject const& obj,
```
rainman110, line 151: Maybe we should omit the default type. This should make the user think more about, what types really to restrict and not simply ignoring it.

rainman110: Also, in most cases (findGtChild...), we need to provide the type  including the Pointer *. Here you don't. 

This breaks the usage compared to the other methods. So I would suggest, to change

    QList<T*> siblings = parent->findDirectChildren<T*>();
    
into
    QList<T> siblings = parent->findDirectChildren<T>();
    
such  that you are forced to use T=GtObject* or similar.

jensschmeink:
@rainman110 I applied your suggestions and fixed the tests.

rainman110 (approved):
Approved: Please stash the commits before merging

PR #1330, opened 2024-12-06
Add color property

jensschmeink:
## Description
The color property from geometry module were moved to core and extended with a dialog to select color instead of hexacode.
The property was not added as the others in core in the data model library as it is important not to use QColor there for the QtGui dependency.
Instead is is registered in the GtApplication 

## How Has This Been Tested?
I replaced the geometry version in the shape color node and it worked fine.

## Checklist:


jensschmeink:
The implementation is just an idea. Fell free to modify

rainman110:
@jensschmeink I fear, that providing a core property only for gui cases is the wrong approach. This leads to the same sitation that we are currently facing, namely that we cannot use QColor, because its defined in the Qt Gui Library. What happens, if some workflow that contains a color property is used in batch mode? 

Instead, we need to provide the color prorperty in out core library. To do so, we need to get rid of QColor, which is not a problem though. We could provide getters and setters for the RGB values such that there is no direct need for QColor. In the GUI library, we can provide a converter from ColorProperty to QColor.

jensschmeink:
> @jensschmeink I fear, that providing a core property only for gui cases is the wrong approach. This leads to the same sitation that we are currently facing, namely that we cannot use QColor, because its defined in the Qt Gui Library. What happens, if some workflow that contains a color property is used in batch mode?
> 
> Instead, we need to provide the color property in our core library. To do so, we need to get rid of QColor, which is not a problem though. We could provide getters and setters for the RGB values such that there is no direct need for QColor. In the GUI library, we can provide a converter from ColorProperty to QColor.

@rainman110 my current implementation does not affect out problem with QtGUI as it is not implemented in the core but in the GUI-library of GTlab. Of course I think it woule be possible to store the property in core using the RGB code. I see the advantage to register all core properties at the same place but with this implementation the property implementation is cleaner using QColor directly.

jensschmeink:
> @jensschmeink I fear, that providing a core property only for gui cases is the wrong approach. This leads to the same sitation that we are currently facing, namely that we cannot use QColor, because its defined in the Qt Gui Library. What happens, if some workflow that contains a color property is used in batch mode?
> 
> Instead, we need to provide the color property in our core library. To do so, we need to get rid of QColor, which is not a problem though. We could provide getters and setters for the RGB values such that there is no direct need for QColor. In the GUI library, we can provide a converter from ColorProperty to QColor.

@rainman110 I understand your point. I'll implement a new solution

jensschmeink:
@rainman110 There is already a conversion option of string and colors. Therefore I now implemented a solution based on the hex representation string of the colors. It worked as well and now is independent of GUI and now part of the core. We could add some converters maybe in the gt_color class.

jensschmeink:
@rainman110 I think now I have a version for a first review. Please have a look and give hints, advice and critics as needed.

jensschmeink:
New feature request: Usage of alpha value for transperancy

src/gui/dock_widgets/properties/items/gt_colorpropertyitem.h
```diff
@@ -0,0 +1,83 @@
...
+#define GTPROPERTYTURBOMAPITEM_H
+
+#include "gt_propertyitem.h"
+
+class GtColorProperty;
+
+/**
+ * @brief The GtColorPropertyItem class
+ */
+class GtColorPropertyItem : public GtPropertyItem
```
rainman110, line 21: I'd like to test the color property in GTlab, could you please add the property to the test module?

src/dataprocessor/property/gt_colorproperty.cpp
```diff
@@ -0,0 +1,212 @@
...
+{
+}
+
+gt::rgb::rgb(const QString& hexString)
+{
+    fromString(hexString);
+}
+
+bool
+gt::rgb::valid() const
```
rainman110, line 117: you don't need the function, if you switch to uint_8

src/dataprocessor/property/gt_colorproperty.cpp
```diff
@@ -0,0 +1,212 @@
...
+}
+
+QString
+gt::rgb::toHexString() const
+{
+    if (!valid()) return {};
+
+    std::stringstream ss;
+    ss << "#"
+       << std::setw(2) << std::setfill('0') << std::hex << m_alpha
```
rainman110, line 204: Please add alpha to the end, and only, if alpha != 255.

src/gui/gt_application.cpp
```diff
@@ -78,6 +81,11 @@ GtApplication::GtApplication(QCoreApplication* parent,
 
     connect(sc, SIGNAL(changed()),
             this, SIGNAL(shortCutsChanged()));
+
+    /// register gui related properties
```
rainman110, line 85: This looks wrong to me. The color prop is not gui only! Where are the other properties registered?

tests/unittests/datamodel/test_gt_colorproperty.cpp
```diff
@@ -0,0 +1,93 @@
...
+    virtual void TearDown()
+    {
+        delete m_prop;
+    }
+
+    GtColorProperty* m_prop;
+
+};
+
+TEST_F(TestGtColorProperty, initialization)
```
rainman110, line 36: Please replace all ASSERT_* by EXPECT_* (use assert only, if you cannot continue the test, if the statement fails)

rainman110 (changes requested):
Thanks for the PR.

In general, the code looks good. The most critical change is to correct the string conversion of colors with alpha. The alpha value needs to be at the end of the string, not at the beginning (this is how is it used by all standards).

I'd like to test the usability from GTlab. Therefore, please also add a color property to the test module.

src/dataprocessor/property/gt_colorproperty.h
```diff
@@ -0,0 +1,148 @@
...
+     * @param r - red rgb value (between 0 and 255 is valid)
+     * @param g - green rgb value (between 0 and 255 is valid)
+     * @param b - blue rgb value (between 0 and 255 is valid)
+     * @param alpha - transparency value (between 0 and 255) where
+     * 0 is fully transparent while 255 represents a fully opaque color
+     */
+    GtColorProperty(QString const& ident,
+                    QString const& name,
+                    QString const& brief,
+                    int r, int g, int b, int alpha = 255);
```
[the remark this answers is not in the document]
jensschmeink: I used the value -1 to make the rgb object invalid. What would be a nice version to do this with the uint8_t arguments?

jensschmeink:
> Thanks for the PR.
> 
> In general, the code looks good. The most critical change is to correct the string conversion of colors with alpha. The alpha value needs to be at the end of the string, not at the beginning (this is how is it used by all standards).
> 
> I'd like to test the usability from GTlab. Therefore, please also add a color property to the test module.

For the hex code of the color with alpha I used the def

> Thanks for the PR.
> 
> In general, the code looks good. The most critical change is to correct the string conversion of colors with alpha. The alpha value needs to be at the end of the string, not at the beginning (this is how is it used by all standards).
> 
> I'd like to test the usability from GTlab. Therefore, please also add a color property to the test module.

The hex code with alpha in the beginning is based on the version as it is used in Qt. Shall I introduce a kind of wrapper for the QColor <-> QString conversion?

src/gui/gt_application.cpp
jensschmeink: This is a relic of my old implementation. I'll fix this

jensschmeink:
@rainman110 I used your suggestions from the review and also made the line edit for the color definition not "read only" anymore. I added an interface to use the color-namespace for initialization to the interface. 
I hope now the property is ready for another review.

jensschmeink:
bad luck. I had no look at the unittests and need some minutes to fix them

src/dataprocessor/property/gt_colorproperty.cpp
```diff
@@ -0,0 +1,330 @@
...
+        gtError() << QObject::tr("Invalid hex code in "
+                                 "property (%1)").arg(hexCodeInput);
+        return false;
+    }
+
+    int shift = 0;
+
+    std::stringstream ss;
+
+    int newIntA = 255;
```
jensschmeink, line 173: I found some strange behavior if the hex code is written directly to the uint8_t variables. The solution here is to use integer to fix the conversion as the std::hex works fine to write the int values and from there to the uint8 variables.

src/dataprocessor/property/gt_colorproperty.cpp
```diff
@@ -0,0 +1,212 @@
...
+
+    std::stringstream ss;
+
+    // Check size of string
+    if (hexCode.length() == 8)
+    {
+        ss << std::hex << hexCode.substr(0, 2); // alpha
+        ss >> m_alpha;
+        ss.clear();
+        shift = 2;
```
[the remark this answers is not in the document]
jensschmeink: As we discussed the standard to write alpha in the end is more html and browser related style while alpha in the beginning is e.g. also used in QColor.

jensschmeink:
@rainman110 Please review this PR so it might be merged soon.

PR #1337, opened 2025-02-05
Implemented map-style properties containers

rainman110:
## Description
The only difference between map container and the normal  (sequential) is, that map-style container don't use uuid as keys but user-defined keys.

These user-defined keys are then displayed in the UI, see:

image

## Why is this important? ##

These could be used, to e.g. define input / output variables for python nodes. No need to use name / value entries. Instead,  name = key. It is much nicer to display and use.

## Checklist:


jensschmeink (approved):
Only two small suggestions for code cleanup. Overall I really enjoy the functionality

jensschmeink:
@rainman110 I added my review few months ago. Maybe have a look and merge

jensschmeink:
@rainman110 some ci jobs are failing

PR #1354, opened 2025-03-21
Allow for module dependency regexes

rainman110:
## Description

If a regex should be used, the identifier must start with "regex:".

I also added a check for module dependency cycles.

## How Has This Been Tested?

The `mdi interface ext` module now uses a regex to define its dependency to the "data model interface" module.

## Checklist:


jensschmeink (commented):
LGTM (only few code style suggesions).
I used it in my master build and all modules had been loaded

rainman110:
@jensschmeink Please re-approve

PR #1359, opened 2025-04-03
Remove property factory from code

jensschmeink:
The GtPropertyFactory was only used in one unittest and not exported. To reduce the number of factories in GTlab it seems to be a good idea to remove this one.

## Description
Factory is removed. There was one closly related test which was removed as well.

## Checklist:


jensschmeink:
@rainman110 please check this. The factory was only used in a unittest and at least in no code of my master build (including acdc and turbine tools)

jensschmeink:
> As written above (add changelog)

Added a changelog entry

PR #1382, opened 2025-07-28
Add new structure to handle property boundaries and add convenience functions for double property construction for container property usage

jensschmeink:
## Description
The code changes relate to two enhancements.

For use in container properties, new convenience functions have been added for double properties to make handling easier.

In addition, this led to the need to handle the numerical limits for properties in the API of the property constructors more cleanly.

To this end, a new class called “Boundaries” was implemented. This provides a clearer interface and is currently used for double and int properties. 

The upper and lower limits are used as optional values.

Translated with DeepL.com (free version)

## How Has This Been Tested?
Unit tests were added, and manual testing of usage was evaluated using an example.

## Checklist:

### DIFF
--- a/src/dataprocessor/property/gt_doubleproperty.h
+++ b/src/dataprocessor/property/gt_doubleproperty.h
@@ -244,7 +244,7 @@ GT_DATAMODEL_EXPORT
  * @param name - name in the GUI
  * @param brief - description for tool tip in GUI
  * @param unitCategory - category for the GTlab unit system
- * @param boundaries - baoundary description
+ * @param boundaries - The numeric boundaries allowed for `:value`
  * @param value Default - value
  * @return function factory for usage in container definitions
  */


jensschmeink:
@rainman110 There was a problem with the CI as the images for windows-2019 were deprecated and not working anymore. I updated the ci file to windows-2022. 
The main topic of the PR is to add few convenience functions for easier usage of double properties in containers as I had to add several generation functions which could be offered in a more general way as it was already done for other properties.

rainman110 (commented):
I like having have these new functions.

I though like to change the interface with a string Bound type. Currently, with all these many doubles, it is easy to mix up lower, upper bound and the value.

Having passing a typed Bound object is imho the way to go. The Bound objects can also have an enum, which bound is active (upper, lower, both)

rainman110:
No, please let not slip in an unfinished API. As said, please provide a typed Bound object. I don't want to have a situation, where there are many doubles, which you will definitely mix up (what is value, what is lower and upper bound).

jensschmeink:
> Here are my two cents on the current approach:
> 
> * Please have a boundaries typed object with lower and upper bound, instead of passing one boundary for lower and one for upper (then you can still mix lower and upper).
> * From the users's point of view, the following API could be good
> 
> ```c++
> auto bounds = Boundaries<int>(10, 20); // set an lower and upper bound
> 
> auto boundsUpper = Boundaries<int>(20, OnlyUpper); // to set only an upper bound
> auto boundsLower = Boundaries<int>(10, OnlyLower); // to set only a lower bound
> 
> Boundaries b2;  // no bounds set yet
> b2.setLower(10); // explicitily set lower bound
> 
> bool hasLower = b2.lowerActive();
> int lowerBound = b2.lowerBound();
> ```
> 
> What do you think?

I think this looks nice. 
How and where would you handle it if the programmer would write 

```c++
auto bounds = Boundaries<int>(20, 10); // in wrong order
```

jensschmeink:
@rainman110 Now there is a solution with a new implementation of gt_optional.h, gt_boundaries.h using this and the initial targeted convenience functions.

src/core/process_management/gt_parameterloop.cpp
```diff
@@ -15,7 +15,7 @@
 GtParameterLoop::GtParameterLoop() :
     m_startVal("start", "start"),
     m_endVal("end", "end"),
-    m_steps("steps", "steps"),
+    m_steps("steps", "steps", "steps", gt::Boundaries<int>::makeLower(1), 1),
```
jensschmeink: This was used as a basic test and there had been no bound before. In the later code there is a division with m_steps so it is important to introduce a bound.

src/dataprocessor/property/gt_doubleproperty.h
```diff
@@ -205,6 +224,39 @@ namespace gt
 GT_DATAMODEL_EXPORT
 gt::PropertyFactoryFunction makeDoubleProperty(double value);
 
+/**
+ * @brief Creates a property factory for doubles with a default value and a unit
+ * @param name - name in the GUI
+ * @param brief - description for tool tip in GUI
+ * @param unitCategory - category for the GTlab unit system
+ * @param lowSideBoundary - Lower side boundary
+ * @param highSideBoundary - High side boundary
```
rainman110:
```suggestion
```

Remove these, the parameters don't exist anymore

rainman110 (changes requested):
I just have some minor change requests to the code.

But please also change the PR title and add a sensible PR description.

rainman110:
> What do you think? Currently even a unit test of int property checks if the bounds are invalid in case of wrong usage.

I think, we should provide a warning message, that the bounds will be reverted. From a UI perspective, can a user also set bounds? Can he also decide to use no bounds? with the old way, the user only would have to invert the bounds (resulting in no bounds). This would not be possible anymore, if i understand it correctly.

rainman110 (approved):
Please squash the commits

PR #1389, opened 2025-09-09
Resolve "Preference Page does not update correctly when changing themes"

### LINKED ISSUE #1388: 🐛 [BUG] - Preference Page does not update correctly when changing themes
### Summary

The Preference Dialog does not update its style/theme correctly when changing themes.

It seems to only affect this dialog

### Expected Behaviour

When changing the theme in the Preference Dialog, it should update the style/theme of the dialog too.

### Reproduction steps

```bash
1. Open Preference Dialog
2. Change Theme
3. Before closing, the dialogs looks quite messy as the styles are only updated partially
```

### Screenshots

Image

### Logs

```bash

```

### GTlab environment

```bash
GTlab Version: 2.1.0-dev
```

### OS

Windows

mariusalexander:
## Description

Closes #1388 - fixed preference dialog not updating correctly when applying theme

## How Has This Been Tested?

Manually

## Checklist:

### DIFF
--- a/src/app/gt_mainwin.cpp
+++ b/src/app/gt_mainwin.cpp
@@ -1260,4 +1260,10 @@ GtMainWin::setTheme(bool /*dark*/)
     gt::gui::applyThemeToApplication();
     gt::gui::applyThemeToWidget(this);
     gt::gui::applyThemeToWidget(ui->mdiArea);
+
+    // update all standalone widgets (i.e. windows)
+    QWidgetList const& windows = QApplication::topLevelWidgets();
+    std::for_each(windows.begin(), windows.end(), [](QWidget* w){
+        gt::gui::applyThemeToWidget(w);
+    });
 }


src/app/gt_mainwin.cpp
```diff
@@ -1258,6 +1258,11 @@ void
 GtMainWin::setTheme(bool /*dark*/)
 {
     gt::gui::applyThemeToApplication();
-    gt::gui::applyThemeToWidget(this);
     gt::gui::applyThemeToWidget(ui->mdiArea);
+
+    // update all standalone widgets (i.e. windows)
+    QWidgetList const& windows = QApplication::topLevelWidgets();
```
rainman110, line 1264: For discussion: Instead of manually looping over all top level widgets, should't we change the function `applyThemeToWidget` such that it also goes trough its child widgets?

mariusalexander: But wouldn't looping over all child elements of a widget be redundant? Firstly, I think `setPalette`, `setStyleSheet` and `setStyle` will already apply to all children. Secondly, I think we skip over the top level widgets (which have no parent, if I am not mistaken)

mariusalexander:
Also I cant really change the target branch in this PR. GitHub thinks it should apply all changes we have done to 2.1 to 2.0...

PR #1407, opened 2025-11-07
Add support for multiple module directories

### LINKED ISSUE #1406: 💡 [REQUEST] - Support for multiple module directories
### Summary

It would be beneficial, to support multiple module directories.

In a typical installation, the module dir is not writeable, so the user should be able to setup own directories.

### Improvements

See above

### Drawbacks

Dlls required be the modules need to be located inside the directory of the module. GTlab will then though try also to load normal DLLs, resulting in an error message.

### Current workarounds

-

rainman110:
## Description

This adds the support for multiple module directories.

 - Module directories are stored in the settings
 - There is a default user and app directory, in addition to multiple user locations
 - The module location now is also displayed in the module about dialog
 - The module location is used to determine module meta data files (Changelog, Readme ...)
 - The environment variable GTLAB_MODULE_DIRS can be used to configure module directories as well.

## How Has This Been Tested?

grafik

grafik

## Checklist:

### DIFF
--- a/src/core/settings/gt_settings.cpp
+++ b/src/core/settings/gt_settings.cpp
@@ -18,8 +18,6 @@
 #include <QKeySequence>
 #include <QString>
 
-#include <gt_logging.h>
-
 struct GtSettings::Impl
 {
 
@@ -575,7 +573,7 @@ GtSettings::userModuleDirs() const
 }
 
 void
-GtSettings::setUserModuleDirs(const QStringList &dirs)
+GtSettings::setUserModuleDirs(const QStringList& dirs)
 {
     QString dirStr = dirs.join(";");
     pimpl->userModuleDirs->setValue(dirStr);
--- a/src/core/settings/gt_settings.cpp
+++ b/src/core/settings/gt_settings.cpp
@@ -18,6 +18,8 @@
 #include <QKeySequence>
 #include <QString>
 
+#include <gt_logging.h>
+
 struct GtSettings::Impl
 {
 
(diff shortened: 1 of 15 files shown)


rainman110:
@real-ct-ac I'd love if you could try it out. What do you think from the user point of view? Should we also incorporate bash environment variables, or only gtlab settings?

src/core/settings/gt_settings.cpp
```diff
@@ -18,6 +18,8 @@
 #include <QKeySequence>
 #include <QString>
 
+#include <gt_logging.h>
```
jensschmeink: This include seems not to be used?!

src/core/settings/gt_settings.cpp
```diff
@@ -557,3 +566,17 @@ GtSettings::setAutostartProcessRunner(bool value)
...
+
+QStringList
+GtSettings::userModuleDirs() const
+{
+    auto dirString = pimpl->userModuleDirs->getValue().toString();
+    return dirString.split(";", Qt::SkipEmptyParts);
+}
+
+void
+GtSettings::setUserModuleDirs(const QStringList &dirs)
```
jensschmeink:
```suggestion
GtSettings::setUserModuleDirs(const QStringList& dirs)
```

jensschmeink (changes requested):
@rainman110 I did the code review with only very small requests. Overall I really like the new ability to select the module folders. 
An application-oriented test should still be carried out (maybe in the context of the project it was needed for)

jensschmeink (approved):
Approved

real-ct-ac (approved):
nice functionality! thank you @rainman110

PR #1416, opened 2025-11-14
[3.0] Don't merge yet: Add const to version function in module interface

jensschmeink:
The function GtModuleInterface::version() is const now.

There has to be discussed how to add this change to a release as it indicates a need for a change in every module (as the function is overwritten)

## Checklist:


jensschmeink:
If this would be part of 2-1 there could be a macro like GT_2_0_ADD_CONST as it is part of gt_compat.h for compatibility from 1-7 to 2-0.

PR #1422, opened 2025-11-17
Xml helper functions to simplify module upgrades

jensschmeink:
For easier writing of module upgrader some basic functions (already used in some modules) were added in generic form to the core repo and some tests were added

## How Has This Been Tested?
Several unittests have beed added including a test for the exsisting use case for python process components.

## Checklist:

### DIFF
--- a/src/core/gt_moduleupgradeutilities.h
+++ b/src/core/gt_moduleupgradeutilities.h
@@ -39,7 +39,7 @@ namespace module_upgrade {
  * @return success indicator (may fail if given container element is empty
  * or not a property-container)
  */
-    GT_CORE_EXPORT bool convertPropertyContainerToMap(
+GT_CORE_EXPORT bool convertPropertyContainerToMap(
     QDomElement& container, const QString& formerNameKey,
     QMap<QString, QString>& replaceMap);
 


src/core/gt_moduleupgradeutilities.h
```diff
@@ -0,0 +1,210 @@
+/* GTlab - Gas Turbine laboratory
```
rainman110: As I said, you can move most of the functions into the gt_xmlutilities. They are very generic and not related to the module upgrade process alone.

src/core/gt_moduleupgradeutilities.h
```diff
@@ -0,0 +1,210 @@
...
+ * @param container
+ * @param formerNameKey
+ * @param replaceMap - a map to collect the mapping of the former uuid based
+ * identification and the new name (for later usage)
+ */
+GT_CORE_EXPORT void normalizePropertyContainerId(
+    QDomElement& container, const QString& formerNameKey,
+    QMap<QString, QString>& replaceMap);
+
+namespace properties {
```
rainman110: Remove `property` namespace. It's too much for my taste. Also you'd have something like

`...::property::setPropertyTypeAndValue`

where property is redundant

src/core/gt_moduleupgradeutilities.h
```diff
@@ -0,0 +1,210 @@
...
+
+    /**
+     * @brief appendProperty
+     * @param parent to append the property to
+     * @param propertyName - name of the new property
+     * @param propertyType - type of the new property
+     * @param value - value of the new property
+     */
+    GT_CORE_EXPORT void appendProperty(
+        QDomElement& parent, const QString& propertyName,
```
rainman110: Rename parameter `parent` to `objectElement`. Or is it not limited to objects? Does it also for for subprops?

src/core/gt_moduleupgradeutilities.h
```diff
@@ -0,0 +1,210 @@
...
+     * @brief appendDoubleProperty
+     * @param parent to append the property to
+     * @param propertyName - name of the new property
+     * @param val
+     */
+    GT_CORE_EXPORT void appendDoubleProperty(
+        QDomElement& parent, const QString& propertyName, double val);
+
+    /**
+     * @brief removeProperty
```
rainman110: @brief !!!

src/core/gt_moduleupgradeutilities.cpp
```diff
@@ -0,0 +1,325 @@
...
+#include <gt_logging.h>
+
+#include "gt_moduleupgradeutilities.h"
+
+namespace {
+    constexpr auto CLASS_STRING = "class";
+    constexpr auto PROP_STRING = "property";
+    constexpr auto OBJ_STRING = "object";
+    constexpr auto NAME_STRING = "name";
+    constexpr auto TYPE_STRING = "type";
```
rainman110: Please use the strings defined in gt_xmlexpr.h, don't duplicate code:

https://github.com/dlr-gtlab/gtlab-core/blob/master/src/dataprocessor/gt_xmlexpr.h

src/core/gt_moduleupgradeutilities.h
```diff
@@ -0,0 +1,210 @@
...
+     * @param parent of the old property element to remove
+     * @param propertyName - name of the property to remove
+     */
+    GT_CORE_EXPORT void removeProperty(QDomElement& parent,
```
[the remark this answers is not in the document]
rainman110: Also, how do you indicate an error?

src/core/gt_moduleupgradeutilities.cpp
```diff
@@ -0,0 +1,325 @@
...
+    parent.appendChild(newObject);
+
+    return newObject;
+}
+
+QDomElement
+gt::module_upgrade::addObjectList(QDomElement& parent)
+{
+    QDomDocument dom = parent.ownerDocument();
+    QDomElement childrenListElement = dom.createElement(QStringLiteral("objectlist"));
```
rainman110: Same here

src/core/gt_moduleupgradeutilities.cpp
```diff
@@ -0,0 +1,325 @@
...
+}
+
+void
+gt::module_upgrade::normalizePropertyContainerId(
+    QDomElement& container, const QString& formerNameKey,
+    QMap<QString, QString>& replaceMap)
+{
+    QDomNode child = container.firstChild();
+
+    while (!child.isNull())
```
rainman110: Use a for loop here, that would allow you to early continue, i,e,

```cpp
for (auto child : container.firstChild(); !child.isNull(); child = child.nextSibling())
```

src/core/gt_moduleupgradeutilities.cpp
```diff
@@ -0,0 +1,325 @@
...
+
+            if (prop.tagName() == PROP_STRING)
+            {
+                QString oldUUID = prop.attribute(NAME_STRING);
+
+                // Search for sub element <property name="$$formerNameKey$$">NewName</property>
+                QDomElement nameElem;
+                QDomNode sub = prop.firstChild();
+
+                while (!sub.isNull())
```
rainman110: Replace with for loop over sub again

tests/unittests/core/test_gt_moduleupgradeutilities.cpp
```diff
@@ -0,0 +1,310 @@
...
+    bool comparisonCheck = domElementsEqual(rootOld,
+                                            doc2.documentElement());
+
+    if (!comparisonCheck)
+    {
+        qWarning() << "old: " << rootOld.ownerDocument().toString(4);
+        qWarning() << "new: " << doc2.toString(4);
+    }
+
+    EXPECT_TRUE(comparisonCheck);
```
rainman110: Please replace, such that you can write

    EXPECT_EQ(rootOld, doc2.documentElement())

I.e. make google test aware of QDomElement, how to compare them and print them

src/core/gt_moduleupgradeutilities.cpp
```diff
@@ -0,0 +1,325 @@
...
+                    results.append(elem);
+
+                    if (searchMode == SearchMode::DisallowNested)
+                    {
+                        continue;  // IMPORTANT: skip children
+                    }
+                }
+            }
+
+            recurseRef(child, recurseRef);
```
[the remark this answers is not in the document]
jensschmeink: The intension here was given by the original usecase: A python calculator can not have a python calculator as child. And so the search shuld stop then.
But it needs to be communicated better, I agree.

src/core/gt_moduleupgradeutilities.h
jensschmeink: It is not limited to parent elements which are objects

src/core/gt_moduleupgradeutilities.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include "gt_xmlutilities.h"
+#include "gt_xmlexpr.h"
+
+#include "gt_moduleupgradeutilities.h"
+
+void
+gt::module_upgrade::convertPropertyContainerToMap(
+    QDomElement& container, const QString& formerNameKey,
+    QMap<QString, QString>& replaceMap)
+{
```
rainman110: Please check, whether `container` is really a property-container by checking the tag name.

Return false, if this is not the case.

src/dataprocessor/gt_xmlutilities.h
```diff
@@ -97,28 +139,58 @@ QDomElement GT_DATAMODEL_EXPORT createPropertyElement(QDomDocument& doc,
...
+}
+
+/**
+ * @brief addNewDoublePropertyElement
+ * Add a new generated property element of type double to a given parent.
+ * @param parent to append the property to
+ * @param propertyName - name of the new property
+ * @param val
+ */
+inline QDomElement addNewDoublePropertyElement(QDomElement& parent,
```
rainman110: Think about removing "New", since there are existing function like "addObjectElement", that don't use "new"

jensschmeink: I removed the "New"

src/dataprocessor/gt_xmlutilities.h
```diff
@@ -97,28 +139,58 @@ QDomElement GT_DATAMODEL_EXPORT createPropertyElement(QDomDocument& doc,
...
+/**
+ * @brief addNewPropertyElement
+ * Add a new generated property element to a given parent.
+ * @param parent to append the property to
+ * @param propertyName - name of the new property
+ * @param propertyType - type of the new property
+ * @param value - value of the new property
+ * @return the created element
+ */
+QDomElement GT_DATAMODEL_EXPORT addNewPropertyElement(QDomElement& parent,
```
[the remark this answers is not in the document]
jensschmeink: I removed the "New"

PR #1424, opened 2025-11-20
Fix bug for boundary application for double properties

jensschmeink:
There was a copy paste bug. For double properties the case of definition wih a low and a high bound was not working correctly.

## Description
This now has been fixed. There was a simple mistake in the code.

## How Has This Been Tested?
Yes. Unittests are added as part of this PR.
The tests cover the existing error as they fail with the old code.

## Checklist:

### DIFF
--- a/tests/unittests/datamodel/test_gt_doubleproperty.cpp
+++ b/tests/unittests/datamodel/test_gt_doubleproperty.cpp
@@ -430,20 +430,20 @@ TEST_F(TestGtDoubleProperty, boundaries)
 
     double hundred = 100.1;
     double fourH = 400.2;
-    bool setError = false;
+    bool success = false;
 
     EXPECT_TRUE(l.lowSideBoundaryActive());
     EXPECT_FALSE(l.highSideBoundaryActive());
     ASSERT_DOUBLE_EQ(l.getVal(), 340.2);
 
     // set valid value
-    l.setVal(fourH, &setError);
-    EXPECT_TRUE(setError);
+    l.setVal(fourH, &success);
+    EXPECT_TRUE(success);
     EXPECT_DOUBLE_EQ(l.getVal(), fourH);
 
     // set invalid value
-    l.setVal(hundred, &setError);
-    EXPECT_FALSE(setError);
+    l.setVal(hundred, &success);
+    EXPECT_FALSE(success);
     EXPECT_FALSE(l.getVal() == hundred);
     EXPECT_DOUBLE_EQ(l.getVal(), fourH);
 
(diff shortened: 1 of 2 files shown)


tests/unittests/datamodel/test_gt_doubleproperty.cpp
```diff
@@ -418,3 +418,69 @@ TEST_F(TestGtDoubleProperty, extendenOperators)
...
+{
+    // low bound check
+    GtDoubleProperty l("propBoundsLow", "test double",
+                       "test brief",
+                       GtUnit::Category::Area,
+                       gt::Boundaries<double>::makeLower(200.43), 340.2);
+
+    double hundred = 100.1;
+    double fourH = 400.2;
+
```
rainman110: Please add an assert, that checks the current value, before changing it.  Please do it after each property creation.

src/dataprocessor/property/gt_doubleproperty.cpp
```diff
@@ -138,7 +138,7 @@ GtDoubleProperty::GtDoubleProperty(const QString& ident, const QString& name,
 
     if (bounds.low())
     {
-        m_boundLo = bounds.high().value();
+        m_boundLo = bounds.low().value();
```
rainman110: Good find!

tests/unittests/datamodel/test_gt_doubleproperty.cpp
```diff
@@ -418,3 +418,69 @@ TEST_F(TestGtDoubleProperty, extendenOperators)
...
+    EXPECT_FALSE(l.highSideBoundaryActive());
+
+    // set valid value
+    l.setVal(fourH);
+    EXPECT_DOUBLE_EQ(l.getVal(), fourH);
+
+    // set invalid value
+    l.setVal(hundred);
+    EXPECT_FALSE(l.getVal() == hundred);
+    EXPECT_DOUBLE_EQ(l.getVal(), fourH);
```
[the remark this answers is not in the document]
jensschmeink: There is the option to give a bool pointer as second argument to check success. I added this to the test.

tests/unittests/datamodel/test_gt_doubleproperty.cpp
```diff
@@ -418,3 +419,80 @@ TEST_F(TestGtDoubleProperty, extendenOperators)
...
+{
+    // low bound check
+    GtDoubleProperty l("propBoundsLow", "test double",
+                       "test brief",
+                       GtUnit::Category::Area,
+                       gt::Boundaries<double>::makeLower(200.43), 340.2);
+
+    double hundred = 100.1;
+    double fourH = 400.2;
+    bool setError = false;
```
rainman110: Please rename to `okay` or `success`. Error is very misleading since setError=true means okay and not error.

PR #1434, opened 2026-01-14
Fixed incorrect packages after backup restore

### LINKED ISSUE #1211: 🐛 [BUG] - Packages/modules are not considered in project backups
### Summary

Activation (and probably also deactivation) of "Project Modules" are ignored by the backup functionality. Also, changes to un-backuped modules (aka packages) are ignored when restoring backups.

### Expected Behaviour

The backup functionality should store and restore the entire project state including the state of activeness of modules and their contents.

### Reproduction steps

```bash
1. Backup a project
2. Activate a module
3. Make some changes to that module, save the project
4. Restore the backup

Now the module, which was activated post-backup, is not only still active, but it even contains the post-backup changes.
```

### Screenshots

DESCRIPTION

### Logs

_No response_

### GTlab environment

```bash
GTlab Version: 2.0.6
```

### OS

Windows

rainman110:
## Description

In fact, this fixed two issues:
 1. The list of packages is incorrect after restore. A package that is in the project but not in the backup remained incorrectly in the project.
 2. If a project backup does not contain a package but the project did, and when reenabling the package after restore, package data were incorrectly added to the project.

Fixes #1211 

## How Has This Been Tested?

Manually with 3 test cases. **Reviewer: Please verify these cases!!!**
 - [x] Case 1: Restore backup, when Project has no package XXX but Backup has. Project should then have package XXX
 - [x] Case 2: Restore backup, when Project has package XXX but Backup has not. Project should not have package XXX anymore.
 - [x] Case 3: Project has a Package XXX with data YYY. Backup does not have Package XXX. Restoring the Backup should remove package XXX. When manually adding the Package XXX, no YYY data should be appended to package XXX.

## Checklist:

### DIFF
--- a/src/core/gt_project.cpp
+++ b/src/core/gt_project.cpp
@@ -338,24 +338,64 @@ GtProject::restoreBackupFiles(const QString& timeStamp)
         return S::ErrorNoBackupFound;
     }
 
-    static QRegularExpression projectAndModuleRex(R"(^[\w\s\-]+.(gtlab|gtmod)$)");
-    auto projectFiles = gt::filesystem::directoryEntries(projectDir,
-                                                         false,
-                                                         projectAndModuleRex);
+    static QRegularExpression projectModuleTasksRex(
+        R"(^[\w\s\-{}]+\.(gtlab|gtmod)$|^tasks\/[\w\s\-{}\/]+\.gttask$)");
+
+    const auto projectFiles = gt::filesystem::directoryEntries(projectDir,
+                                                         true,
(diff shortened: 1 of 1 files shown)


rainman110:
@marvinoe21 I changed the code.

my changes
 - Revert the refactoring, so that adding / removing packages works again
 - Included tasks into the temporary backup (stash)
 - Fixed the dot "." matching in the regex
 - Address style issue

marvinoe21 (approved):
The changes look good! I tested the described test cases again and everything works as expected.

PR #1455, opened 2026-03-23
Improve object selection dialog filtering performance on large projects

### LINKED ISSUE #1454: 🐛 [BUG] - GtObjectSelectionDialog becomes extremely slow on large projects when type-filtered
### Summary

  GtObjectSelectionDialog becomes extremely slow on large projects when used from GtPropertyObjectLinkEditor. In practice, typing a single character into the search field can take several seconds, and in some
  cases around 10 seconds.

  The dialog uses GtObjectModel together with GtStyledModel and GtObjectFilterModel. The same underlying object model performs well in the explorer dock, even for large projects.

  The likely root cause is the type filter in GtObjectFilterModel: it performs recursive subtree checks for every row on every filter update. In the dialog this is always active because setFilterData() is called
  with the allowed object classes before the dialog is shown. This combines badly with expandAll() on each filter update.

  Relevant code:

  - src/gui/tools/gt_objectselectiondialog.cpp
  - src/core/gt_objectfiltermodel.cpp
  - src/core/gt_treefiltermodel.h
  - src/core/gt_explorermodel.cpp

### Potential fix

  The current implementation appears to do too much recursive work during filtering:

  - GtObjectFilterModel::filterAcceptsRow() always uses the recursive tree-filter overload with the predicate allowed class AND text matches.
  - For most nodes in a large tree, the class predicate is false, so the filter keeps descending into children to decide visibility.
  - This likely causes repeated subtree traversals on every keystroke.
  - GtObjectSelectionDialog also calls expandAll() after every filter update, which amplifies the cost.

  A targeted fix would be:

  1. Precompute the set of visible nodes when setFilterData() changes the allowed classes.

  - include all objects whose class is allowed
  - include all ancestors of those objects
  - store them in a set for O(1) membership checks

  2. During text filtering, only evaluate rows inside that precomputed visible subset.

  - avoid recursive type-filter scans on every keystroke

  3. As secondary cleanup:

  - remove one of the duplicate search signal connections (textEdited / textChanged)
  - avoid unconditional expandAll() on every text change
  - enable setUniformRowHeights(true) for the dialog tree view

  The main expected win is from removing repeated recursive type filtering during live search.

### Expected Behaviour

Filtering in GtObjectSelectionDialog should remain responsive even for large projects.

  Typing into the search field should feel comparable to filtering in the explorer dock, even when only specific object classes are allowed.

### Reproduction steps

```bash
1. Open a large project.
2. Open a property editor that uses GtPropertyObjectLinkEditor.
3. Open the object selection dialog.
4. Use a property that restricts selectable objects to one or a few classes.
5. Type into the dialog search field.
6. Observe that each keystroke can take several seconds.
```

### Screenshots

-

### Logs

```bash
No dedicated runtime error log is required to reproduce this issue.
 The problem is UI performance during filtering.
```

### GTlab environment

```bash
GTlab Version: all
```

### OS

Windows

rainman110:
## Description

  This PR improves filtering performance in GtObjectSelectionDialog for large projects.

  The main speedup comes from changing the filter/update order in the dialog:

  - collapse the tree before applying a new search filter.
  - then expand only the visible branches afterwards
  - small result sets stay automatically expanded (i.e. if less than 11 results are found, the view keeps expanded)

  This avoids refiltering an already expanded large tree, which was the main reason incremental search became very slow after the first typed character.

  Additional small improvements:

  - use a hashed lookup for allowed object types in GtObjectFilterModel
  - remove duplicate search-trigger connections
  - enable uniform row heights for the dialog tree view

  A small unit test was also added for GtObjectFilterModel to cover the new core-side behavior.

 Fixes #1454.

## How Has This Been Tested?

  - manually tested on large projects with runtime instrumentation
  - verified that the dominant cost dropped from refiltering an expanded tree to acceptable collapse/expand work
  - verified that matches stay visible in the hierarchy while filtering
  - added and ran TestGtObjectFilterModel.*

## Checklist:

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -13,6 +13,8 @@ ColumnLimit: 80
 PointerAlignment: Left
 MaxEmptyLinesToKeep: 2
 
+AllowShortIfStatementsOnASingleLine: WithoutElse
+
 
 # classes
 BreakConstructorInitializers: AfterColon 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  - Module updater get process model files to have the opportunity to modify process elements for new versions - #1414
 
 ### Fixed
+ - Improved performance of the object selection dialog filtering on large projects, especially during incremental search with broad type filters - #1454
  - Fixed alphabetically sorting of Shortcuts in Preference View #482
 
 ## [2.0.11] - 2025-04-03
--- a/src/core/gt_objectfiltermodel.cpp
+++ b/src/core/gt_objectfiltermodel.cpp
@@ -23,6 +23,7 @@ void
 GtObjectFilterModel::setFilterData(const QStringList& val)
 {
     m_filterData = val;
+    m_filterLookup = QSet<QString>(m_filterData.begin(), m_filterData.end());
     invalidate();
 }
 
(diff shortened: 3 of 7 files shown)


jensschmeink (approved):
I really enjoy the speed up!

PR #1459, opened 2026-04-15
Use object name in "open with" context menu for MDI items

### LINKED ISSUE #1124: Use the Object Name or Display Name for MDI Items in the Object UI Context Menu
This issue was migrated from our internal issue tracker (issue no. 88)

Currently the class name is displayed in the "open with" context menu. Instead the object name of the MDI Item or a display name should be used.

Currently, we do not have a dedicated interface/API for providing a dedicated display name, thus requiring an ABI/API change. Creating an instance of the object could be expensive or lead to other problems and thus sup-optimal as well. However, when the module is registered/loaded, we may already create an instance of the MDI object, which we could use to retrieve the object name. But I/we have to check that.

To avoid an ABI change a solution could be to "engrave" the display name in the return string of the `openWith` function in the ObjectUI: Instead of returning `CLASS_NAME` one could return `CLASS_NAME (DISPLAY_NAME)`

We have a helper macro to access the classname (`GT_CLASSNAME`), we could introduce another macro like `GT_OPENWITH(CLASS_TYPE, DISPLAY_NAME)` to simplify this.

```cpp
QStringList
MyObjectUI::openWith(GtObject *obj)
{
    return QStringList{
        MyMDI::staticMetaObject.className() + // <- class name
        QStringLiteral("(My Editor)")         // <- display name
    };
    // OR
    return QStringList{
        GT_CLASSNAME(MyMDI) +          // <- class name
        QStringLiteral("(My Editor)")  // <- display name
    };
    // OR
    return QStringList{
        GT_OPENWITH(MyMDI, "My Editor") 
    };
}
```

rainman110:
## Description

Implemented customizable "Open with" strings

It now uses the object name of the MDI item (if available at construction time).

During module initialization, we create all objects - also MDI items.  I exploited this to get the object name from the freshly created object and register the name to a map, now included into the mdi launcher.

I needed to extend the mdi launcher. I therefore used the timing, to pimplify the mdi laucher.

**ABI CHANGE**

**NOTE**: This does not require to change current modules. It could potentially be back-ported to 2.0.X **without an ABI change**.

Closes #1124

## How Has This Been Tested?

With the geometry module

Also, with the test modules:
grafik

## Checklist:

### DIFF
--- a/src/gui/gt_mdilauncher.cpp
+++ b/src/gui/gt_mdilauncher.cpp
@@ -32,6 +32,16 @@ static const QString S_WINDOW_PREFIX = QStringLiteral("GTlab - ");
 namespace
 {
 
+QString hardCodedOpenWithName(const QString& className)
+{
+    if (className == "GtdPreDesignPlot") return "Pre Design Plot";
+    else if (className == "GtdPreDesignPlot3D") return "3D Pre Design Plot";
+    else if (className == "GtdProfilePlot")  return "Profile Plot";
+    else if (className == "GtdComponentEditor") return "Component Editor";
+
+    return "";
+}
+
 /// helper function to make tab buttons
 QWidget* makeTabButtons(QPointer<GtTabWidget> tabWidget,
                         QIcon const& tabIcon,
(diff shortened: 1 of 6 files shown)


rainman110:
@jensschmeink We need to change the GUI tests here. But if we change it on master, all other branches will fail as the action names won't match.

Do you have any idea, how to solve it?

rainman110:
> @jensschmeink We need to change the GUI tests here. But if we change it on master, all other branches will fail as the action names won't match.
> 
> Do you have any idea, how to solve it?

Fixed

src/gui/gt_mdilauncher.cpp
```diff
@@ -32,6 +32,16 @@ static const QString S_WINDOW_PREFIX = QStringLiteral("GTlab - ");
 namespace
 {
 
+QString hardCodedOpenWithName(const QString& className)
```
jensschmeink: I think the content of this function is not needed. The first three editors already have the object names in their constructors and the 4th is unknown to me (is it part of geo copa?)

jensschmeink (commented):
Overall a nice feature! I just have one annotation about the hard coded list

jensschmeink:
And one question: why is it not an ABI change if there are removed several member elements of a class which is exported?

src/gui/gt_mdilauncher.cpp
rainman110: I have put them for backwards compatibility. If we really don't need it - the better!

rainman110: I don't know either, what the fourth module is coming from. It seems to be in there for quite a long time.

rainman110:
> And one question: why is it not an ABI change if there are removed several member elements of a class which is exported?

It is an ABI change. But we could port it back without ABI change to 2.0.X by introducing a dummy member enforcing the same object size. not sure though, if we should do that. I am in favour of only having this in 2.1.

jensschmeink:
> > And one question: why is it not an ABI change if there are removed several member elements of a class which is exported?
> 
> It is an ABI change. But we could port it back without ABI change to 2.0.X by introducing a dummy member enforcing the same object size. not sure though, if we should do that. I am in favour of only having this in 2.1.

Ok, I just missunderstood your comment in the description. I'm fine if it is a 2-1 solution aswell.

src/gui/gt_mdilauncher.cpp
rainman110: I removed it, since it is not required anymore

jensschmeink (approved):

PR #1460, opened 2026-04-16
Deprecate 'Custom' unit category

### LINKED ISSUE #1450: [CHANGE] - Custom unit is missleading
### Summary

The unit category Custom is missleading and should be removed in the next major release. 
Especially the unit conversion related to that type is very strange and physically not consistent

### Impact on the codebase

There are a few usages of the unit type. These should be analyzed to find out which definitions should be used instead and might have to be added.

### Current workarounds

The unit type is used rarely e.g. for
- stage loading
- sealing coefficient (mass / area)
- heat conductivity
- condenser kA
- heat transfer coefficient

jensschmeink:
The unit type custom is a mess.
There are a lot of unit conversion registered which have nothing in common. 
Overall the unit-type should be avoided.

Therefore a first step is to mark the enum entry as deprecated.
I spoke with users of "GtUnit::Custom" and for their needs to replace it with a more meaningfull alternative.

This alternatives are meant to be added by this PR aswell

## Description
- Mark GtUnit::Custom as deprecated to avoid further usage
- Add more unit types to handle heat transfer and related topics

## Checklist:

Note:
Overall I think it would be beneficial to start a unit-interface. For more usecases of GTlab the number of needed units will be increasing. Currently unit definitions can only be added to the core

### DIFF
--- a/src/dataprocessor/property/gt_unit.h
+++ b/src/dataprocessor/property/gt_unit.h
@@ -52,6 +52,9 @@ public:
                    Current,
                    Resistance,
                    Impedance,
+                   HeatTransferCapability,
+                   HeatTransferCoefficient,
+                   ThermalConductivity,
                    /// non physical units
                    NonDimensionalPercentage = 64,
                    NonDimensional,
--- a/src/dataprocessor/property/gt_unit.h
+++ b/src/dataprocessor/property/gt_unit.h
@@ -43,7 +43,7 @@ public:
                    EnergyDensity,
                    PowerDensity,
                    PowerTempArea,
-                   Custom,
+                   Custom [[deprecated("Do not use Custom. It is not a valid unit category.")]],
                    MassMomentInertia,
                    VGVGradients,
                    VGVGradientsPow2,
--- a/src/dataprocessor/property/gt_unit.h
+++ b/src/dataprocessor/property/gt_unit.h
@@ -10,6 +10,7 @@
 #include "gt_datamodel_exports.h"
 
 #include <QString>
+#include "gt_version.h"
 
 class GT_DATAMODEL_EXPORT GtUnit
 {
@@ -43,7 +44,7 @@ public:
                    EnergyDensity,
                    PowerDensity,
                    PowerTempArea,
-                   Custom [[deprecated("Do not use Custom. It is not a valid unit category.")]],
+                   Custom GT_DEPRECATED_ATTR(2, 2, "Do not use Custom. It is not a valid unit category."),
                    MassMomentInertia,
                    VGVGradients,
                    VGVGradientsPow2,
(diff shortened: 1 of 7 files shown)


rainman110:
@jensschmeink still draft?

jensschmeink:
> @jensschmeink still draft?

Sorry, no. Only a relict in the name

src/dataprocessor/property/gt_unit.h
```diff
@@ -43,7 +43,7 @@ class GT_DATAMODEL_EXPORT GtUnit
                    EnergyDensity,
                    PowerDensity,
                    PowerTempArea,
-                   Custom,
+                   Custom [[deprecated("Do not use Custom. It is not a valid unit category.")]],
```
rainman110: Ideally, please use the macro `GT_DEPRECATED_REMOVED_IN` to mark and enforce the version of removal.

When would it be a good idea?

jensschmeink: Cool. I was not aware of that. For all modules which are part of the master build will replace the usage as soon as the new units of this PR are available. As it is really bad code I would remove it soon, maybe 2.2?!

rainman110: 2.2 could be working. with 2.1, these deprecations will be seen first to developers. so the devs have one minor version time to fix it.

rainman110:
Please add UNIT tests to fix the coverage gaps as well. You added those units to the test modules, which though don't count for coverage.

jensschmeink:
> Please add UNIT tests to fix the coverage gaps as well. You added those units to the test modules, which though don't count for coverage.

GT_DEPRECATED_REMOVED_IN is not working at that position in the code, but GT_DEPRECATED_ATTR does. Is this ok aswell? I think the static assert of the first one does not work inside a enum.

rainman110:
Please then add GT_REMOVAL_GUARD(2,2, "msg") in front of the line with the deprecation. I makes the code fail, when we arrive at gtlab 2.2. GT_DEPRECATED_ATTR  does only print a deprecation message, but does not enforce.

jensschmeink:
> Please then add GT_REMOVAL_GUARD(2,2, "msg") in front of the line with the deprecation. I makes the code fail, when we arrive at gtlab 2.2. GT_DEPRECATED_ATTR does only print a deprecation message, but does not enforce.

Even in the line above it does not work.  :/

rainman110:
> > Please then add GT_REMOVAL_GUARD(2,2, "msg") in front of the line with the deprecation. I makes the code fail, when we arrive at gtlab 2.2. GT_DEPRECATED_ATTR does only print a deprecation message, but does not enforce.
> 
> Even in the line above it does not work. :/

Then put is just somewhere, e.g. above the enum

rainman110:
@jensschmeink I removed some warnings due to the deprecation and added that extra compile guard. please check

jensschmeink:
@rainman110 I checked your changes and they are fine. I added few more lines of testing code to fullfill coverage. Please also have a look

rainman110 (approved):

PR #1466, opened 2026-05-22
Improve connection editor to check compatibility of object link and double properties

### LINKED ISSUE #1147: Connection Editor does not respect allowed classes
Calculator A has object link property with GtdRotorBladeRow as only allowed class. Calculator B has object link property with GtdStatorBladeRow as only allowed class. The Connection Editor allows these two properties to be connected.
It would be nice if the Connection Editor would not allow this. Else, there is additional workload for the process developer in order to handle errors.

Originally posted by O.K. in bug tracking group

jensschmeink:
The connection editor allows connections between properties of the same type. Sometimes it might be helpfull to make more restirctions. 
Therefore in the PR the connection of object links checks if object links are at least in one class compatible.
Additionally double properties checks for the same unit. This is ignored if at least one of the units is None or Nonedimensional to avoid problems with values from python-process elements

## Description
- checks are extended for the mentioned configurations
- the process connection item added a public getter to the related abstract property

## How Has This Been Tested?
Manual GUI tests in larger process setups

## Checklist:

## Summary by Sourcery

Improve process connection validation by enforcing compatibility checks for object link and double properties when connecting ports.

New Features:
- Restrict connections between object link properties to those with at least one compatible target class.
- Restrict connections between double properties to those sharing the same unit, while allowing unitless or nondimensional values for interoperability with Python-defined elements.

Enhancements:
- Expose the associated abstract property from process connection items via a new public getter.
- Simplify connection and valid-port checks with early returns in the process connection scene and port entity logic.

### DIFF
--- a/src/app/mdi_items/process_env/entities/gt_processpropertyportentity.cpp
+++ b/src/app/mdi_items/process_env/entities/gt_processpropertyportentity.cpp
@@ -20,6 +20,8 @@
 #include "gt_processconnectionitem.h"
 #include "gt_processpropertyportentity.h"
 #include "gt_stringmonitoringproperty.h"
+#include "gt_objectlinkproperty.h"
+#include "gt_doubleproperty.h"
 
 #include "gt_colors.h"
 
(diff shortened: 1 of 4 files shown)


jensschmeink:
@rainman110 Please check for a review. In a later PR we can follow the idea of a general registration how properties are defined to be compatible.

jensschmeink:
@rainman110 Can you also have a look here? This PR is a bit older but would also be benefitial I think

mariusalexander (approved):
LGTM

PR #1468, opened 2026-05-26
Adding a search functionality to the memento viewer

### LINKED ISSUE #380: Search function in memento viewer
In GitLab by @jensschmeink on Jan 5, 2023, 07:48

There is a feature request from Oliver in the GTlab BugReport-project: [#31](https://gitlab.dlr.de/gtlab-bugtracking/gtlab-issuetracking/-/issues/31).

jensschmeink:
Thememento viewer needs a kind of search function which is implemented in this PR.

## Description
- the search widget is added to the memento viewer
- The search widget got buttons to jump to next or prevoius found element
- This needs to be discussed: most search functions in GTlab currently work as a finlter and not a jump-to-search. So the buttons are not needed in general. Should there be a flag in the existing search widget to enable these buttons or should it be a separate implementation? 

## Checklist:

## Summary by Sourcery

Add in-editor search with match navigation to the memento viewer and generalize the shared search widget/editor infrastructure to support this behavior.

New Features:
- Add an in-place search bar to the memento viewer with buttons and shortcuts to jump between search matches.
- Introduce F3 and Shift+F3 core shortcuts for navigating to the next and previous search result across the application.

Enhancements:
- Extend the shared code editor to highlight all occurrences of a search term while preserving existing line highlighting.
- Augment the generic search widget to optionally show next/previous navigation buttons and encapsulate its internals via a PIMPL implementation.
- Add a dedicated search highlight color to the code editor color palette for consistent theming of search results.

### DIFF
--- a/src/app/mdi_items/memento_viewer/gt_mementoviewer.cpp
+++ b/src/app/mdi_items/memento_viewer/gt_mementoviewer.cpp
@@ -126,11 +126,13 @@ GtMementoViewer::onSearchTextChanged(const QString& text)
         startPos = cursor.selectionEnd();
     }
 
-    if (text.isEmpty())
+    if (text.isEmpty() || m_matches.isEmpty())
     {
-        m_editor->setTextCursor(QTextCursor());
+        QTextCursor cur = m_editor->textCursor();
+        cur.clearSelection();
+        m_editor->setTextCursor(cur);
     }
-    else if (!m_matches.isEmpty())
+    else
     {
         m_currentMatch = 0;
         m_editor->setTextCursor(m_matches.first());
--- a/src/app/mdi_items/memento_viewer/gt_mementoviewer.cpp
+++ b/src/app/mdi_items/memento_viewer/gt_mementoviewer.cpp
@@ -48,6 +48,7 @@ GtMementoViewer::GtMementoViewer()
     m_searchWidget = new GtSearchWidget(widget());
     searchLay->addWidget(m_searchWidget);
     searchLay->addStretch(1);
+    m_searchWidget->enableFindNextButtons();
     lay->addLayout(searchLay);
 
     // Shortcut (Ctrl+F) triggers search widget
--- a/src/app/mdi_items/memento_viewer/gt_mementoviewer.cpp
+++ b/src/app/mdi_items/memento_viewer/gt_mementoviewer.cpp
@@ -152,12 +152,6 @@ GtMementoViewer::goToNextMatch()
     QTextCursor cur = m_matches.at(m_currentMatch);
     m_editor->setTextCursor(cur);
     m_editor->ensureCursorVisible();
-
-    // Reapply search highlights after cursor movement (preserves line highlight)
-    if (!m_searchText.isEmpty())
-    {
-        m_editor->highlightOccurrences(m_searchText);
-    }
 }
 
 void
@@ -170,11 +164,4 @@ GtMementoViewer::goToPrevMatch()
     QTextCursor cur = m_matches.at(m_currentMatch);
     m_editor->setTextCursor(cur);
     m_editor->ensureCursorVisible();
-
-    // Reapply search highlights after cursor movement (preserves line highlight)
-    if (!m_searchText.isEmpty())
-    {
-        m_editor->highlightOccurrences(m_searchText);
-    }
 }
-
--- a/src/app/mdi_items/memento_viewer/gt_mementoviewer.cpp
+++ b/src/app/mdi_items/memento_viewer/gt_mementoviewer.cpp
@@ -10,27 +10,62 @@
 
 #include <QIcon>
 #include <QVBoxLayout>
+#include <QHBoxLayout>
 
 #include "gt_mementoviewer.h"
 #include "gt_icons.h"
 #include "gt_mementoeditor.h"
 #include "gt_xmlhighlighter.h"
+#include "gt_searchwidget.h"
+#include <QShortcut>
+#include <QKeySequence>
+#include <QTextDocument>
+#include <QTextCursor>
+#include "gt_application.h"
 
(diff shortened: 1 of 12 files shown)


jensschmeink:
As I added some features to the exported search widget so there will be an ABI change

jensschmeink:
@rainman110 I moved the buttons to the right of the search widget and use the small icons as in the python editor now. Please review

src/app/mdi_items/memento_viewer/gt_mementoviewer.h
```diff
@@ -46,9 +52,28 @@ class GtMementoViewer : public GtMdiItem
...
+    GtSearchWidget* m_searchWidget;
+    /// Cached match positions
+    QList<QTextCursor> m_matches;
+    int m_currentMatch = -1;
+    QString m_searchText;
+
+    // Navigation shortcuts
+    QShortcut* m_nextShortcut = nullptr;
+    QShortcut* m_prevShortcut = nullptr;
+
```
rainman110, line 73: Initialize all pointers via the {nullptr} syntax, not via "=". This equality assign in header is not portable.

src/app/mdi_items/memento_viewer/gt_mementoviewer.cpp
```diff
@@ -10,27 +10,72 @@
...
+    searchLay->addStretch(1);
+    m_searchWidget->enableFindNextButtons();
+    lay->addLayout(searchLay);
+
+    // Shortcut (Ctrl+F) triggers search widget
+    auto* shortcut = new QShortcut(gtApp->getShortCutSequence("search"),
+                                   m_editor);
+    shortcut->setContext(Qt::WidgetWithChildrenShortcut);
+    connect(shortcut, &QShortcut::activated, m_searchWidget,
+            &GtSearchWidget::enableSearch);
```
rainman110, line 59: Bind search navigation shortcuts to the viewer .  Because `enableSearch()` moves focus into the search line, these QShortcuts attached to `m_editor` with `WidgetWithChildrenShortcut` never fire while the user is actually searching. In practice F3/Shift+F3 only work after focus is moved back into the XML editor, so the new keyboard   navigation is unreachable from the search field.

src/app/mdi_items/memento_viewer/gt_mementoviewer.cpp
```diff
@@ -52,3 +97,67 @@ GtMementoViewer::onThemeChanged()
...
+    }
+}
+
+void
+GtMementoViewer::goToPrevMatch()
+{
+    if (m_matches.isEmpty() || !m_editor) return;
+
+    // Move to previous match
+    m_currentMatch = (m_currentMatch - 1 + m_matches.size()) % m_matches.size();
```
rainman110, line 164: Fix the initial previous-match index.  After every query change `m_currentMatch` is reset to `-1`, so the first Previous/Shift+F3 action computes `(m_currentMatch - 1 + size) % size` and lands on `size-2` instead of the last hit whenever there is more than one match. Users who start with the previous shortcut will jump to   the wrong occurrence.

src/gui/tools/gt_codeeditor.cpp
```diff
@@ -207,3 +207,56 @@ GtCodeEditor::lineNumberAreaPaintEvent(QPaintEvent* event)
...
+        QTextEdit::ExtraSelection sel;
+        sel.cursor = cursor;
+        sel.format = fmt;
+        selections.append(sel);
+        // Move start position past this match to avoid infinite
+        // loop on zero‑length search text
+        startPos = cursor.selectionEnd();
+    }
+
+    setExtraSelections(selections);
```
rainman110, line 272: Keep search highlights when the caret moves! These matches are written into `extraSelections()`, but `GtCodeEditor::highlightCurrentLine()` is still connected to `cursorPositionChanged` and clears that list for read-only editors. In the memento viewer that means simply clicking elsewhere in the document immediately removes the   highlighted hits until the search text changes again.

rainman110 (changes requested):
The new search UI has a broken initial Previous navigation path, its keyboard shortcuts are effectively unusable while the search field has focus, and search highlights disappear on normal caret movement in the read-only viewer. These are functional regressions in the newly added feature.

jensschmeink:
@rainman110 thank you for the good feedback. I'm not completly sure if I understand the whole problem but I think I solved it now. Maybe we can talk tomorrow to make sure that I addressed your review correctly

rainman110 (commented):
I still found two issues in your implementation after testing.

### 1. Wrong previous found item after initial search:

*Steps to reproduce* :

  1. Open a memento in the viewer.
  2. Press Ctrl+F and search for a term that appears multiple times.
  3. Do not press Next first.
  4. Press Shift+F3 or click Previous.

  *Expected*:

  - It should jump to the last match in the document.

  *Actual*:

  - It jumps to the second-to-last match instead.

  Why this happens:

  - `onSearchTextChanged()` resets `m_currentMatch to -1`.
  - The first Previous computes (m_currentMatch - 1 + size) % size, which becomes size - 2 on the first press.

### 2. Search result is only highlighted AFTER mouse move

At least I experienced it on Linux.

*Steps to reproduce*:

  1. Open a memento in the viewer.
  2. Press Ctrl+F and search for a term that appears multiple times.
  3. Press next: Nothing gets highlighted! After mouse movement, the first match will be highlighted though.

*Expected* :
   - The highlight should immediately be done after pressing "next"

*Actual*:
   - The highlighting still needs a mouse move to trigger repaint.

jensschmeink:
@rainman110 I worked on the problems and now they should be fixed and windows and linux

jensschmeink:
@rainman110 I like your adaption to fix the problem for linux. I think we can merge this now. Please check for approval

rainman110 (approved):

PR #1476, opened 2026-06-10
Deprecate unused monitoring data classes to reduce complexity of task execution

### LINKED ISSUE #1475: [CHANGE] - Remove unused Monitoring data table
### Summary

There is a relict from early days of GTlab: The MonitoringDataTable. I think it is not used anymore by anybody but makes the process structure more complex. 

### Impact on the codebase

ABI/API changes in GTCore when the class is removed.
But some parts of the Task implementation might also be less complex.

### Current workarounds

Currently just unused

jensschmeink:
Currently there are some old classes in the code which are not used anymore.
This should be removed soon or at least be marked as deprecated

## Description
Monitoring data classes had been used for some (already removed) gui representation of process flows (which looked a bit like our modern intelligraphs).
The handling of these classes makes the task processing more complex without any benefit.

## How Has This Been Tested?
After removing the classes a task from the optimization module still interacted correctly via connection editor

## Checklist:

## Summary by Sourcery

Deprecate legacy task monitoring data structures and remove their runtime usage from task execution flow.

Enhancements:
- Stop collecting and propagating monitoring data in task and parameter loop execution to simplify process handling.
- Remove task runner signal wiring and debug logging related to monitoring data transfer.
- Mark GtMonitoringData, GtMonitoringDataSet, GtMonitoringDataTable, and related GtTask monitoring methods as deprecated for removal in version 2.2.


rainman110:
@jensschmeink I just quickly looked at the build output. It triggers many deprecation warnings, i.e. we are now internally using many of the functions, that are actually deprecated. So please fix this first.

jensschmeink:
> @jensschmeink I just quickly looked at the build output. It triggers many deprecation warnings, i.e. we are now internally using many of the functions, that are actually deprecated. So please fix this first.

I used the given macros to supress the deprecated warnings as recommended. Now please have a look

src/core/process_management/gt_task.cpp
```diff
@@ -414,26 +420,20 @@ GtTask::runChildElements()
...
-    GtMonitoringDataSet monData = collectMonitoringData();
-
-    // check whether monitoring data has entries
-    if (!monData.isEmpty())
-    {
-        // monitoring data available - emit signal
-        emit monitoringDataTransfer(m_currentIter, monData);
-    }
-
     return true;
```
rainman110, line 432: Please keep collecting and emitting the legacy monitoring data until its announced removal in 2.2. Removing this now means callers still compile against the deprecated API in 2.1, but normal task execution silently leaves the monitoring table empty.

src/core/process_management/gt_task.cpp
```diff
@@ -507,6 +510,8 @@ GtTask::collectMonitoringDataHelper(GtMonitoringDataSet& map,
         // collect data for each child recursively
         collectMonitoringDataHelper(map, child);
     }
+
+    GT_SUPPRESS_DEPRECATED_BEGIN
```
rainman110: This should be `GT_SUPPRESS_DEPRECATED_END`, not `BEGIN`. As written, the diagnostic state is pushed a second time and deprecated-declaration warnings remain suppressed for the rest of this translation unit.

src/core/process_management/gt_task.cpp
```diff
@@ -592,17 +597,19 @@ void
 GtTask::onMonitoringDataAvailable(int iteration, GtMonitoringDataSet const& set)
 {
     // append data set to data table and check success
+    GT_SUPPRESS_DEPRECATED_BEGIN
```
rainman110, line 609: Please preserve `emit monitoringDataAvailable()` after a successful append during the deprecation window. Direct users of this still-present public slot otherwise observe a silent behavioral change even outside the removed runner wiring.

src/core/process_management/gt_taskrunner.cpp
```diff
@@ -224,25 +223,6 @@ GtTaskRunner::setupElements(GtProcessComponent* orig,
     connect(cloned, &GtProcessComponent::transferMonitoringProperties,
```
rainman110, line 223: Please retain the task-specific monitoring signal/slot connections for 2.1 and deprecate the public surface consistently. Removing this wiring now makes the retained table accessors non-functional for normal runner execution.

rainman110 (changes requested):
The legacy monitoring-data-table path appears to be a good candidate for removal, but I do not think we should disable its runtime behavior in the same release in which we first deprecate it.

The current change creates a half-alive API: the types and most task methods remain callable and advertise removal in 2.2, while normal task execution no longer collects, forwards, clears, or notifies about the data. Existing downstream code will still compile on 2.1 but silently receive empty tables or miss notifications.

I suggest a two-step migration:

1. In 2.1, keep the complete behavior and signal/slot wiring intact, while deprecating the whole legacy surface consistently:
   - `GtMonitoringData`, `GtMonitoringDataSet`, and `GtMonitoringDataTable`;
   - the related `GtTask` methods and slots;
   - `monitoringDataTransfer`, `monitoringDataAvailable`, and `triggerClearMonitoringData`.
2. Document that the historical table API is obsolete and that live monitoring-property transfer remains supported.
3. Keep focused compatibility tests for collection, runner propagation, table clearing, and notification signals during the deprecation window.
4. In 2.2, remove the classes, task table state, meta-type registration, signals/slots, collection code, runner wiring, sources, tests, and documentation together.

Please therefore restore the removed runtime behavior in this PR and limit it to a consistent deprecation. The existing developer example in `docs/dev/basics/process_elements.rst` also still shows `collectMonitoringData()` and `monitoringDataTransfer`; it should either remain valid during 2.1 or be updated with a clear migration explanation.

I did not find an ABI layout or symbol removal in the current patch, but changing the behavior before 2.2 is an API compatibility regression. I also could not run the unit-test build locally because the available environment lacks `GTlabLoggingConfig.cmake`.

jensschmeink:
@rainman110 New status is (as discussed):
- calls of deprecated functions in deprecated functions of the same deprecation-level using warning suppression
- Behaviour of the functions is not changed
- functions which are not deprecated calling deprecated functions are not using the suppress macros to remind us to have a look for the next version.

Please give feedback

src/core/process_management/gt_parameterloop.cpp
```diff
@@ -128,15 +128,5 @@ GtParameterLoop::runChildElements()
...
-    GtMonitoringDataSet monData = collectMonitoringData();
-
-    // check whether monitoring data has entries
-    if (!monData.isEmpty())
-    {
-        // monitoring data available - emit signal
-        emit monitoringDataTransfer(m_currentIter, monData);
-    }
-
     return true;
```
rainman110, line 141: The legacy collection/emission block is still removed here. To keep behavior unchanged throughout the 2.1 deprecation window, please restore the same `collectMonitoringData()` / `monitoringDataTransfer` logic for `GtParameterLoop` as well.

src/core/process_management/gt_task.cpp
```diff
@@ -63,7 +64,6 @@ GtTask::GtTask() :
 
     m_currentIter.setVal(0);
 
-    qRegisterMetaType<GtMonitoringDataSet>("GtMonitoringDataSet");
 
     registerProperty(pimpl->processRunner, tr("Execution"));
```
rainman110, line 69: Please restore `qRegisterMetaType<GtMonitoringDataSet>("GtMonitoringDataSet")`. The restored clone-to-original `monitoringDataTransfer` connection can cross from worker execution to the original task; without registering this custom by-value argument, Qt may reject queued delivery and the monitoring table will still remain empty.

rainman110 (changes requested):
Thanks, the latest commits address most of the previous review:

- the regular `GtTask::runChildElements()` collection/emission path is restored;
- task-runner signal/slot wiring is restored;
- `monitoringDataAvailable()` is emitted again;
- the mismatched suppression macro is fixed;
- suppression is now generally limited to deprecated implementations and explicit legacy tests.

There are still two behavior regressions that need to be restored before this is a deprecation-only change:

1. `GtParameterLoop::runChildElements()` still no longer collects/emits its per-iteration monitoring dataset.
2. `qRegisterMetaType<GtMonitoringDataSet>()` is still removed. The cloned task emits `monitoringDataTransfer(int, GtMonitoringDataSet)` from the runnable's worker execution toward the original task. That queued delivery requires the custom value type to remain registered; otherwise Qt can reject the queued argument and the restored runner wiring will not actually transfer the table.

The public legacy signals `monitoringDataTransfer`, `monitoringDataAvailable`, and `triggerClearMonitoringData` also remain undeprecated. Since the agreed plan is to deprecate the complete table surface in 2.1 and remove it in 2.2, these should carry the same deprecation/removal annotation if moc/compiler support permits it (otherwise the exception should be documented).

Please also add a focused runner/queued-signal compatibility test. The existing tests cover direct collection and manual table mutation, but would not detect the missing meta-type registration or the ParameterLoop regression.

rainman110:
@jensschmeink Thanks, your suppression approach now matches what we discussed: deprecated implementations may suppress calls to APIs at the same deprecation level, while non-deprecated callers should keep producing warnings so that the remaining migration work stays visible.

One additional consistency point: I agree that the three public `GtTask` signals belonging to the legacy table path should also be deprecated for removal in 2.2:

- `monitoringDataTransfer(int, GtMonitoringDataSet)`
- `monitoringDataAvailable()`
- `triggerClearMonitoringData()`

External modules can connect to these signals and they are part of the Qt meta-object API, so leaving them unannotated would make the deprecation incomplete. This does **not** apply to the supported live-property path via `GtProcessComponent::transferMonitoringProperties()`, which should remain unchanged.

If the removal macro does not work cleanly in a `signals:` section with moc, using the attribute-only form or documenting that technical exception would be reasonable. The preferred result is still a compiler-visible deprecation where possible.

The two remaining behavioral points from my follow-up review are restoring the `GtParameterLoop` collection/emission block and restoring `qRegisterMetaType<GtMonitoringDataSet>()` for queued signal delivery.

jensschmeink:
@rainman110 I restored the removed lines for gt_task and gt_parameterloop.

rainman110:
@jensschmeink Thanks, the behavior is restored now. Please also deprecate the three legacy signals `monitoringDataTransfer`, `monitoringDataAvailable`, and `triggerClearMonitoringData` for removal in 2.2. After that, we can approve.

jensschmeink:
@marvinoe21 Please check the last change. @rainman110 is ok with most of it and requested only three additional deprecation hints which I added recently.

rainman110:
@jensschmeink Tanks for deprecating the three methods as asked. Now though, these deprecations cause new warnings in the build, which we need to fix before merge.

rainman110 (changes requested):
See, comment above. The new deprecation warnings need to be fixed before merge

jensschmeink:
@rainman110 In an earlier conversation here we decided to keep deprecation warnings if they help to remember to remove relevant code later which cannot be restuctured now with a later version. This is here the case as the marked code should be removed in the next version. The whole structure of monitoring table is not used anymore and the corresponding GUI already had been removed several version ago as it had not worked for years.

PR #1479, opened 2026-06-12
Checks for gtApp validity before dark mode is checked

### LINKED ISSUE #1477: 🐛 [BUG] - Crash getting theme-dependent colors when no gtApp is running
### Summary

Many colors in `gt_colors.cpp` use `gtApp->inDarkMode()` without checking, whether gtApp is actually valid.

This makes integration tests hard, since they require a running GtApplication.

Proposal: just check gtApp first. If not existent, just return the bright mode color.

### Expected Behaviour

No crash, when no gtApp is available.

### GTlab environment

```bash
GTlab Version: master
```

### OS

All

jensschmeink:
As mentioned in a bug report for gtApp-nullptr there might be crashes in several parts of the code.
This should be reduced by additional checks in various GUI classes

## Description
Before the dark mode is requested the check for the validity of the gtApp pointer is checked. If it is not valid the bright mode is selected

## How Has This Been Tested?
Manual 

## Checklist:

## Summary by Sourcery

Guard dark mode dependent GUI logic against null gtApp to avoid crashes and default to light mode when the application instance is unavailable.

Bug Fixes:
- Prevent potential null-pointer dereferences when querying dark mode state across color, stylesheet, icon, dialog, and tool components by checking gtApp validity first.

Enhancements:
- Default various GUI themes, palettes, and syntax highlighting styles to light mode when gtApp is not available while preserving dark mode behavior when it is.

### DIFF
--- a/src/app/dialogs/gt_checkforupdatesdialog.cpp
+++ b/src/app/dialogs/gt_checkforupdatesdialog.cpp
@@ -20,7 +20,7 @@
 #include <QHeaderView>
 #include <QFont>
 
-#include "gt_application.h"
+#include "gt_guiutilities.h"
 #include "gt_updatechecker.h"
 #include "gt_logging.h"
 #include "gt_icons.h"
@@ -161,7 +161,7 @@ GtCheckForUpdatesDialog::updateAvailable()
     m_checkButton->setEnabled(true);
 
     QString fontString = "darkgreen";
-    if (gtApp && gtApp->inDarkMode())
+    if (gt::gui::isApplicationDarkTheme())
     {
         fontString = "green";
     }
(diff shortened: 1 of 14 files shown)


rainman110:
I agree with the comments from the review bot. Please extract the check into a free function to reduce code duplication.

Another idea is to provide a template function with two arguments, one with the return value for bright mode, one for the dark mode and use this .

jensschmeink:
> I agree with the comments from the review bot. Please extract the check into a free function to reduce code duplication.
> 
> Another idea is to provide a template function with two arguments, one with the return value for bright mode, one for the dark mode and use this .

Extracting the function makes sense. But I'm not sure about a template function: If there would be more themes as now it would have to be restructured.
I would prefer a free function returning an enum (currently with the options bright and dark). I think this could be extended later more easily. Or do you think, if there would be more modes we would need a greater refactoring at all and should stay with a simple solution now?

jensschmeink:
@rainman110 as a small test I implemented a encapulated function in the gui_utilities using a enum for the theme. But espacially the last part would be easy to remove before merging

rainman110 (approved):
LGTM

rainman110:
> Extracting the function makes sense. But I'm not sure about a template function: If there would be more themes as now it would have to be restructured. I would prefer a free function returning an enum (currently with the options bright and dark). I think this could be extended later more easily. Or do you think, if there would be more modes we would need a greater refactoring at all and should stay with a simple solution now?

I agree. It's fine like this

PR #1483, opened 2026-06-22
Reduce code doubling in GtProcessComponentSettingsButton

jensschmeink:
Reduce some code doublings in the code.
As the TaskData and CalculatorData do not share much functions so some if structures are still needed. 

## How Has This Been Tested?
Manual checks of the GUI reaction

## Checklist:

## Summary by Sourcery

Refactor process component settings handling to reduce duplicated logic for calculators and tasks by centralizing wizard lookup and memento application.

Enhancements:
- Introduce shared helper functions to retrieve custom wizards for calculators and tasks and to apply provider data back to process components.
- Simplify null and readiness checks in process component setup and wizard opening to make the control flow more concise.

### DIFF
--- a/src/app/dock_widgets/process/gt_processcomponentsettingsbutton.cpp
+++ b/src/app/dock_widgets/process/gt_processcomponentsettingsbutton.cpp
@@ -20,12 +20,44 @@
 #include "gt_project.h"
 #include "gt_task.h"
 #include "gt_calculator.h"
-#include "gt_command.h"
 #include "gt_extendedtaskdata.h"
 #include "gt_stylesheets.h"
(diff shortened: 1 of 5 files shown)


jensschmeink:
@rainman110 As I mentioned there is the problem of different GUI related classes and also other problems for better abstractions as different factories, ...

Please have a look to check if this code change is still valueble and should be merged

jensschmeink:
This change does not completly fullfil the target to simplify the settings button as I think a connected factory of process elements should be used for that. But at least it is a start to reduce redundant definitions

mariusalexander (approved):
LGTM

PR #1491, opened 2026-07-08
Refactoring of the Gt-Graphics-API

### LINKED ISSUE #575: GtGrid should not be depend on GtGraphicsView but on QGraphicsView
In GitLab by @mariusalexander on Jul 17, 2023, 12:31

The `GtGrid` class requires an instance of `GtGraphicsView` as a parameter, however none of its features are used (*as far as I can see*). Therefore I would propose to remove the dependency, so that the class could be used with a `QGraphicsView` instance instead too.

### LINKED ISSUE #576: GtGraphicsView should use QGraphicsScene instead
In GitLab by @mariusalexander on Jul 17, 2023, 12:34

Similarly to #575 the `GtGrahpicsView` requieres a `GtGraphicsScene` as parameter however none if its features are used.

### LINKED ISSUE #577: [GtGraphicsScene] Extract helper method for accessing items of a certain type and consider removing the class
In GitLab by @mariusalexander on Jul 17, 2023, 12:37

Currently the GtGraphicsScene contains only a single helper method, that could be refactored out so that the class may be removed all together.

### LINKED ISSUE #1224: [CHANGE] - Refactor API of GtGrid class to expose members and allow customization of minor/major gird lines settings
### Summary

IMO `GtGrid` has multiple issues that needs addressing to improve its usability:
- For starters, grid height and width are poorly named (IMHO) since these properties refer to the spacing between horizontal and vertical lines. This should be renamed to horizontal/vertical-advance for example.
- The `GtGrid` class currently also provides not getters for the "height" and "width"
- The minor gird lines cannot be customized whatsoever

### Impact on the codebase

API/ABI change

### Current workarounds

One has to remember the grid height/width manually in code. This is a no-go

mariusalexander:
## Goals:

- Remove `GtGraphicsScene` as it does not add any functionality except for a template function
- Make `GtGraphicsView` use `QGraphicsScene`s and allow null-scenes
- Make `GtGrid` not depend on `GtGraphicsView`
- Make `GtGrid` not responsible to render `GtRuler`
- Make Ownership-Model more clear (Grid is owned by view privately (not using parent child), rulers are not owned by the view)
- Cleanup API of `GtGrid`, add getters and options to tweak a grid to ones liking

## Description

**GtGraphicsScene:** 
- deprecated the class as it does not add any functionality except for a template function.
- Moved template function to `gt_guiutilities.h`
- Closes #577

**GtGraphicsView:** 
- Now uses `QGraphicsScene` - Closes #576
- Pimpled the class
- Added option to control whether the scene is deleted if the view is deleted, as this behavior deviates from `QGraphicsView` and is not entirely clear either.
- `GtGrid` is no longer deleted by `GtGraphicsView` explicitly, but using parent child relations
- Separation of concerns:
  - Previously one hat to register the rulers on the grid AND the view, which is very unintuitive. Now the view triggers the drawing of the grid and updates rulers, serving as a convenience class and a "bridge" between rulers and grids. 
- Smooth zooming is more stable and zooms more quickly. Moving the mouse while the zoom animation is active no longer causes the view to jump unexpectedly.
- One can now disable smooth zooming feature using a flag.

**GtGrid:** 
- Separation of concerns:
  - No dependencies to any `QGraphicsView`, entirely decoupled using dedicated `paint` method - Closes #575
  - No longer responsible for rendering `GtRuler` (in fact, the grid does no longer know about rulers)
  - Only tiny struct defining the actual grid spacing (`GtGridSpacing`) is shared
- Pimpled the class
- `GtGrid` got a dedicated `QObject* parent = nullptr` constructor, making ownership more clear
- Standardizes function naming: 
  - Renamed grid's width and -height to HSpacing and VSpacing respectively (It was ambiguous if actual size of grid was changed or just the spacing of the grid lines!)
  - Use `updated` signal instead of `update` signal (which sounded more like it can be used to update the grid instead)
- Added more setters and getters for all properties
  - Allows to set custom pens for major, minor and axis lines (e.g. thickness of lines, dashed or solid)
  - Allows to specify subdivisons of minor grid
  - Defining various scaling strategies: Besides default Base2, added Base10 and 1-2-5 Scheme
- Deprecated old setters and redundant methods
  - Many methods did nothing!
- Closes #1224

**GtRuler:**
- Pimpled the class
- Separation of concerns:
  - Now renders itself
  - Not dependent on `GtGrid`, operates on the actual grid spacing (`GtGridSpacing`)
  - Not dependent on `QGraphicsView`, entirely standalone

## Minor API changes

`GtRuler::buffer()` now returns `QPixmap` instead of `QImage` (more optimized for `QWidget` rendering)

## Showcase

New API allows to set number of minor grid lines, allows grid lines and axis to be rendered with dotted lines (for example), different thicknesses are supported. Font of rulers can also be changed:

grafik

## Documentation

Detailed all changes and a migration plan in the dev documentation

## How Has This Been Tested?

Manually, made sure old uses of graphics view still look and feel the same.
Updated the MDI-Test-Module to make use of the new Grid-API.

## Checklist:

## Summary by Sourcery

Refactor the graphics API around standard Qt scenes and independently configurable grid and ruler components while preserving migration paths for legacy consumers.

New Features:
- Expose a configurable grid API with independent spacing, subdivisions, visibility, pens, axes, and zoom scaling strategies.
- Allow graphics views to work with nullable QGraphicsScene instances and configurable scene ownership, smooth zooming, and grid snapping.
- Make rulers independently configurable and render them from shared grid spacing data.

Bug Fixes:
- Stabilize smooth zooming so cursor movement during zoom no longer causes unexpected view jumps.

Enhancements:
- Decouple grids and rulers from graphics views and separate their rendering responsibilities.
- Replace the custom graphics scene abstraction with standard QGraphicsScene support and a shared graphics-item lookup utility.
- Clarify ownership and modernize the graphics view, grid, and ruler implementations using private implementation details and deprecated compatibility APIs.

Documentation:
- Add a migration guide covering the graphics API changes and deprecations.

Tests:
- Add comprehensive unit coverage for grid spacing strategies, snapping calculations, configuration updates, and painter-state preservation.

Chores:
- Update the MDI viewer showcase to use the new graphics view, grid, and ruler APIs.


mariusalexander:
@sourcery-ai review

mariusalexander:
@sourcery-ai guide

mariusalexander:
@sourcery-ai review

mariusalexander:
@sourcery-ai review

mariusalexander:
@sourcery-ai review

mariusalexander:
@rainman110 May you have a look at the changes? It ballooned into quite a big PR and I am sorry for that, its difficult to review now. The `GtGraphicsView`, `GtGrid` and `GtRuler` were quite dependent on each other. It was easiest to just tackle all classes in one PR, for testing and development.

You can have a look at the description, which gives a detailed overview of what has changed. TLDR, each class is much more self contained, the API has been refined and a lot more customization options are now possible.

Best you look at the `TestMdiViewer` and play with the view yourself, change some settings etc.

tests/modules/mdi_interface/mdi/test_mdi_viewer.cpp
```diff
@@ -13,26 +13,67 @@
...
+    hruler->setFont(font);
+
+    vruler->flipAxis(true);
+
+    auto* lay = new QGridLayout();
+    lay->setContentsMargins(0, 0, 0, 0);
+    lay->setSpacing(0);
+    lay->addWidget(hruler, 0, 1);
+    lay->addWidget(vruler, 1, 0);
+    lay->addWidget(m_view, 1, 1);
```
mariusalexander, line 76: @rainman110 Have a look at this code, to see how the API operates now.

As a side note, previously the API for adding rulers looked like this:
```cpp
    view->setGrid(new GtGrid(*view));
    view->grid()->setShowAxis(true);
    auto* hRuler = new GtRuler(Qt::Horizontal);
    auto* vRuler = new GtRuler(Qt::Vertical);
    view->setHorizontalRuler(hRuler);
    view->setVerticalRuler(vRuler);
    view->grid()->setHorizontalRuler(hRuler);
    view->grid()->setVerticalRuler(vRuler);

    auto* lay = new QGridLayout();
    lay->setContentsMargins(0, 0, 0, 0);
    lay->setSpacing(0);
    lay->addWidget(hRuler , 0, 1);
    lay->addWidget(vRuler , 1, 0);
    lay->addWidget(view, 1, 1);
```

src/gui/tools/gt_grid.cpp
```diff
@@ -2,483 +2,619 @@
...
+
+GtGrid::~GtGrid() = default;
+
 void
-GtGrid::setHorizontalRuler(GtRuler* ruler)
+GtGrid::setHSpacing(unsigned value)
 {
-    m_hRuler = ruler;
+    pimpl->hSpacing = value;
+    emit updated();
```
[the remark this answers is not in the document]
mariusalexander: Done

src/gui/tools/gt_grid.cpp
```diff
@@ -2,483 +2,619 @@
...
+            return baseSpacing;
+        }
     }
-}
 
-void
-GtGrid::paintRuler(GtRuler* ruler)
-{
-    if (!ruler)
+    GtGridSpacing scaledGridSpacing(double zoom) const
```
rainman110, line 164: `scaledGridSpacing()` is declared `const`, but it assigns to `cachedSpacing` and `cachedZoom` below. Neither member is `mutable`, so the current head should fail to compile. Please either make caching an explicitly mutable implementation detail or make this method non-const (and keep the public const contract in mind). A focused compile/unit test around `scaledGridSpacing()` would catch this.

src/gui/tools/gt_grid.cpp
```diff
@@ -2,483 +2,619 @@
...
+//    const double decadeExp = std::floor(std::log10(ideal));
+    const double fraction   = normalised / std::pow(10.0, decadeExp); // in [1, 10)
+    assert(fraction >= 1 && fraction <  10);
+
+    const double multiplier =
+            (fraction < 1.5) ? 1.0 :
+            (fraction < 3.5) ? 2.0 :
+            (fraction < 7.5) ? 5.0 : 10.0;
+
+    return multiplier * std::pow(10, decadeExp);
```
rainman110: The 1-2-5 result loses `baseSpacing`. For example, with `baseSpacing == 100` and `ideal == 100`, `normalised == 1` and this returns `1`, whereas the Base2/Base10 strategies return `100`. This should presumably be `baseSpacing * multiplier * pow(10, decadeExp)` (or the normalization should be changed consistently). Please add table-driven tests with a non-unit base spacing and zooms below/above 1.

src/gui/tools/gt_graphicsview.cpp
```diff
@@ -2,346 +2,381 @@
...
-GtGraphicsView::setHorizontalRuler(GtRuler *ruler)
+GtGraphicsView::setGrid(GtGrid* grid)
 {
-    if (!m_grid)
+    if (pimpl->grid && pimpl->grid->parent() == this)
     {
-        gtWarning().verbose().nospace()
-                << __FUNCTION__ << ": " << tr("Could not set Ruler. Set grid first!");
-        return;
+        delete pimpl->grid;
```
rainman110: The same-pointer case is still unsafe: `view.setGrid(view.grid())` enters this branch, deletes the grid, and then assigns/parents the dangling `grid` argument. Please return early when `grid == pimpl->grid` before deleting anything. A `QPointer<GtGrid>` test that sets the same instance twice is a cheap regression test.

src/gui/tools/gt_grid.cpp
```diff
@@ -2,483 +2,619 @@
...
+{
+    GT_REMOVAL_GUARD(2, 2, "remove this connection");
+    connect(this, &GtGrid::updated, this, &GtGrid::update);
+}
+
+GtGrid::~GtGrid() = default;
+
 void
-GtGrid::setHorizontalRuler(GtRuler* ruler)
+GtGrid::setHSpacing(unsigned spacing)
```
rainman110, line 313: Changing the base spacing does not invalidate `cachedZoom`. If spacing was already computed for the current zoom, the next `scaledGridSpacing(zoom)` hits the `qFuzzyCompare` fast path and returns the old cached spacing. Please invalidate/recompute the cache from every setter that affects spacing (H/V spacing, subdivisions where relevant, strategy), and test changing spacing without changing zoom.

src/gui/tools/gt_grid.h
```diff
@@ -2,197 +2,505 @@
...
+    void setAxisColor(const QColor& color);
+
+    /**
+     * @brief Returns the axis pen color.
+     * @return Axis color
+     */
+    QColor axisColor() const;
+
+    GT_DEPRECATED_REMOVED_IN(2, 2, "Use `GtGraphicsView::connectHorizontalRuler` instead.")
+    void setHorizontalRuler(GtRuler* ruler) {}
```
rainman110: This preserves compilation but silently drops the old behavior. Existing clients can keep calling the deprecated API and receive no indication that ruler integration stopped working. Please retain it as a functional adapter to the new view/ruler path if possible; otherwise remove/document it as an intentional source/API break rather than using a silent no-op. The same applies to `setVerticalRuler()` and `paintRuler()`.

src/gui/tools/gt_ruler.cpp
```diff
@@ -2,114 +2,290 @@
...
+    invalidate();
 }
 
-QImage&
-GtRuler::buffer()
+void
+GtRuler::flipAxis(bool flipped)
 {
-    return m_buffer;
+    pimpl->axisFlipped = flipped;
```
rainman110, line 100: Changing the flip setting affects tick labels, but the cached ruler is not invalidated or updated here. If the rect/transform stay unchanged, the next `paint()` returns early and continues displaying the old labels. Please call `invalidate()` (and request an update as appropriate). The same cache-invalidation principle should be tested for orientation and font changes.

src/gui/tools/gt_ruler.h
```diff
@@ -2,84 +2,156 @@
...
+    ~GtRuler() override;
 
-    /** Returns ruler orientation.
-        @return Ruler orientation */
-    Qt::Orientation orientation();
+    /**
+     * @brief Returns the buffer used for caching the rendered ruler.
+     * @return Buffer
+     */
+    QPixmap& buffer();
```
rainman110: This changes the public source-level contract from `QImage&` to `QPixmap&`; existing code binding the result to `QImage&` no longer compiles. ABI breaks are allowed for the new major, but the project’s stated goal is to avoid API/source breaks. Could the raw buffer accessor remain compatible (or be deprecated with a separate new pixmap accessor)? If the type change is unavoidable, it should be called out explicitly as an intentional API break with affected callers migrated.

rainman110 (changes requested):
Thanks for the substantial cleanup and for addressing the earlier Sourcery feedback. I reviewed all 15 changed files at PR head `a853e61d`.

There are a few issues I consider blocking:

- `Impl::scaledGridSpacing(double) const` writes non-mutable cache members and should not compile as written.
- The `OneTwoFive` strategy drops `baseSpacing`, producing values in raw 1/2/5 units instead of multiples of the configured spacing.
- `setGrid(view.grid())` still deletes the object and then reuses the dangling pointer.
- Several deprecated ruler methods silently do nothing. Keeping a function only to compile, while discarding its observable behavior, is a behavioral API break that is harder for clients to diagnose than a compile failure.
- Grid spacing caches are not invalidated when base spacing changes, so snapping/rulers can observe stale values.
- Ruler configuration such as `flipAxis()` does not invalidate the cached rendering.

The change from the previous `qPow(2, factor)` behavior should also be locked down explicitly in tests. The current implementation has evolved into scaling strategies, which is good, but the intended values—especially for zooms below 1—should be documented by tests rather than manual inspection.

A small, high-value automated test set would be enough:

1. Table-driven `scaledGridSpacing()` tests for Base2, Base10, OneTwoFive, and Fixed at zooms `0.25, 0.5, 1, 2, 4`, using a non-unit base spacing such as 100.
2. Verify changing H/V spacing at unchanged zoom immediately changes `currentGridSpacing()`/snap results.
3. Verify spacing and subdivision setters clamp zero to one.
4. Verify snapping without a grid and mouse movement with a null scene are safe.
5. Verify `setGrid()` with the same pointer twice, replacement, `nullptr`, external deletion, and view deletion using `QPointer`.
6. Verify owned vs. unowned scene destruction, including replacing the active scene.
7. Render to a small `QImage` and assert major/minor/axis options, axis orientation, visibility, and preservation of painter state.
8. Use `QSignalSpy` to verify visual setters emit `updated`.
9. Verify deprecated adapters still have observable behavior, particularly ruler connection/painting.
10. Verify changing ruler orientation, axis flipping, font, and size invalidates/rebuilds its cache.

These can be ordinary GTest/Qt tests using `QPointer`, `QSignalSpy`, and an off-screen `QImage`; no fragile screenshot/golden test infrastructure is necessary.

src/gui/tools/gt_grid.cpp
mariusalexander: @rainman110, please have a look again. `Impl::scaledGridSpacing` uses the `cachedSpacing` and `cachedZoom` but does not set it.

src/gui/tools/gt_graphicsview.cpp
mariusalexander: Good catch

src/gui/tools/gt_grid.cpp
mariusalexander: Only `setHSpacing` and `setVSpacing` were affected. Subdivisions are recalculated every draw event.

src/gui/tools/gt_ruler.cpp
mariusalexander: Good catch

mariusalexander:
@rainman110 Where do I note explicit API changes? In the Changelog? Or just by "breaking" the API instead of allowing silent or implicit conversions in the code?

src/gui/tools/gt_ruler.h
mariusalexander: Seems unavoidable to me. I've renamed the function and made it protected (no need for it to be public IMO)

src/gui/tools/gt_grid.h
mariusalexander: Got it

src/gui/tools/gt_grid.cpp
mariusalexander: good catch!

mariusalexander:
@rainman110, I have added tests and revised the code where necessary. I replied to each bullet point of yours.

One glaring issue: How should I test the GUI stuff in the unittests? I cannot create widgets without a `QApplication` and creating widgets in a console application seems wrong. We need a different testing environment for this. The GUI Tests are hard to maintain, time intensive to extend and one cannot easily test things like setting the same grid twice without adding a `QAction` in some bespoke test module. Ideally we need some kind of GUI-Test-Application that operate like unittests and tests itself. I think this would need to be addressed in a different PR.

> Impl::scaledGridSpacing(double) const writes non-mutable cache members and should not compile as written.

I could not verify this observation. Please check again

> The OneTwoFive strategy drops baseSpacing, producing values in raw 1/2/5 units instead of multiples of the configured spacing.

Updated the code to incorporate base scaling. The unittests show that the scaled grid spacings follow the base scaling. 

> setGrid(view.grid()) still deletes the object and then reuses the dangling pointer.

Updated the code.

> Several deprecated ruler methods silently do nothing. Keeping a function only to compile, while discarding its observable behavior, is a behavioral API break that is harder for clients to diagnose than a compile failure.

Functions that were there only for the sake of backwards compatibility and that did not perform any operation are now removed.

> Grid spacing caches are not invalidated when base spacing changes, so snapping/rulers can observe stale values.

Only `setHSpacing` and `setVSpacing` were affected and updated accordingly in the latest commits. `setSubdivisons` is not affected. Unittests verify this behavior.

> Ruler configuration such as flipAxis() does not invalidate the cached rendering.

Updated the code.

> The change from the previous qPow(2, factor) behavior should also be locked down explicitly in tests. 

Testing if qPow is equal to std::exp2 feels unnecessary.
Testing if the previous way of calculating the scaled grid still yields the same results is hard to test:
There have been many issues in the grid implementation, the `factor` was previously set externally.
If one really wants to make sure the grid does not change in its appearance, visual regressions tests are necessary.
Further, even if they differ, the grid should now behave/render more correctly.

> The current implementation has evolved into scaling strategies, which is good, but the intended values—especially for zooms below 1—should be documented by tests rather than manual inspection.

Unittests were added

> A small, high-value automated test set would be enough:
> Table-driven scaledGridSpacing() tests for Base2, Base10, OneTwoFive, and Fixed at zooms 0.25, 0.5, 1, 2, 4, using a non-unit base spacing such as 100.

done
    
> Verify changing H/V spacing at unchanged zoom immediately changes currentGridSpacing()/snap results.

done

> Verify spacing and subdivision setters clamp zero to one.

done

> Verify snapping without a grid and mouse movement with a null scene are safe.

Would need to create QWidgets -> different setup is necessary

> Verify setGrid() with the same pointer twice, replacement, nullptr, external deletion, and view deletion using QPointer.
  
Would need to create QWidgets -> different setup is necessary

> Verify owned vs. unowned scene destruction, including replacing the active scene.
  
Would need to create QWidgets -> different setup is necessary

> Render to a small QImage and assert major/minor/axis options, axis orientation, visibility, and preservation of painter state.

This seems overkill (and honestly quite a hassle to do). Tell me if you disagree.
Further: Would need to create QWidgets -> different setup is necessary

> Use QSignalSpy to verify visual setters emit updated.

done for grid only using counters and lambdas, as `QSingalSpy` is Qt6 feature.

> Verify deprecated adapters still have observable behavior, particularly ruler connection/painting.
   
I think looking at the what functions the deprecated adapters call is good enough. Tell me if you disagree.
   
> Verify changing ruler orientation, axis flipping, font, and size invalidates/rebuilds its cache.

Would need to create QWidgets -> different setup is necessary

src/gui/tools/gt_grid.cpp
rainman110: You are right — I misread the assignments in the surrounding code. `Impl::scaledGridSpacing()` only reads `cachedSpacing` and `cachedZoom`; the cache is updated by the non-const callers. My compilation concern was incorrect, and the green CI confirms that. Sorry for the noise; this thread can be resolved.

src/gui/tools/gt_grid.cpp
```diff
@@ -2,483 +2,638 @@
...
+{
+    pimpl->hSpacing = std::max(spacing, 1u);
+    pimpl->invalidateChachedSpacing();
```
rainman110, line 316: Invalidating `cachedZoom` fixes subsequent `scaledGridSpacing(zoom)` calls, but `currentGridSpacing()` still returns the old `cachedSpacing` until the next paint or explicit `setCurrentZoom()`. Consequently, direct snapping immediately after changing H/V spacing—and snapping before the first paint—can use stale/default `{1,1}` spacing. Please make the current spacing coherent immediately (e.g. recompute using the last known zoom or lazily recompute from an invalid-cache state). This is cheap to cover in the existing pure Grid tests: set zoom, change spacing without painting, then assert `currentGridSpacing()` and `computeNearestGridPoint()` use the new value.

src/gui/tools/gt_grid.cpp
```diff
@@ -2,483 +2,638 @@
...
-    if (m_scaleGrid)
+    if (pimpl->showAxis && options.testFlag(PaintAxis))
     {
-        if (m_gridFactor >= 0)
-        {
-            return (length * qPow(2, m_gridFactor + 1));
-        }
+        pimpl->paintAxis(painter, rect);
+    }
+    painter.setRenderHints(oldRenderHints);
```
rainman110, line 606: Only render hints are restored, while `paintGridLines()`/`paintAxis()` also replace the painter pen. Since `GtGrid::paint()` is public, it should leave the complete painter state unchanged. Please use `painter.save()`/`restore()` (preferably RAII-safe) or explicitly restore every changed property. This needs no GUI harness: paint into a small `QImage`, set a distinctive pen beforehand, call `grid.paint(...)`, and assert the pen and transform are unchanged afterwards.

rainman110 (changes requested):
Thanks for addressing the first review so thoroughly. I checked the current head `0a6d0a31` and the new grid test suite. The same-pointer ownership bug, OneTwoFive scaling, spacing-cache invalidation for scaled queries, ruler flip invalidation, zero handling, and update-signal coverage are all substantially improved.

I am retracting my previous const-compilation comment separately: you were correct!

Regarding the two source-API breaks. I accept them for the 2.1 development:
- removal of the legacy grid/ruler methods instead of retaining silent no-ops;
- removal of the public `GtRuler::buffer()` API and encapsulation of the pixmap cache.

Please document these explicitly in the migration/deprecation notes of the developer documentation, including a short old/new ruler setup example:
- `GtGraphicsScene` -> `QGraphicsScene`;
- `findItems<T>()` -> `gt::gui::findGraphicItems<T>()`;
- scene ownership options;
- removed `GtGrid::setHorizontalRuler`, `setVerticalRuler`, and `paintRuler`;
- rulers connected through `GtGraphicsView::connectHorizontalRuler` / `connectVerticalRuler`;
- removed public `GtRuler::buffer()` access.

Regarding tests: the existing unit-test executable uses `QCoreApplication`, so I agree that widget-level tests for `GtGraphicsView` and `GtRuler` would require a separate `QApplication`/offscreen test setup and are disproportionate for this PR. I am not requesting those as merge blockers. The existing `TestMdiViewer` can remain the manual integration check, and a dedicated headless GUI-test target can be follow-up work.

The two remaining inline points are both testable cheaply with the current infrastructure: one pure `GtGrid` cache/snap test and one `QImage`/`QPainter` state test.

rainman110:
> @rainman110 Where do I note explicit API changes? In the Changelog? Or just by "breaking" the API instead of allowing silent or implicit conversions in the code?

Please add it to the developer documentation. There is a migration to 2.1 section.

src/gui/tools/gt_grid.cpp
mariusalexander: It seems rendering to a pixmap also requires a GUI Application (and I cant create a GUI Application because a Core Application already exists).

mariusalexander: A QImage works!

src/gui/tools/gt_grid.cpp
mariusalexander: done

mariusalexander:
> > @rainman110 Where do I note explicit API changes? In the Changelog? Or just by "breaking" the API instead of allowing silent or implicit conversions in the code?
> 
> Please add it to the developer documentation. There is a migration to 2.1 section.

Done. @rainman110 ready for the next review :)

src/gui/tools/gt_grid.cpp
```diff
@@ -2,483 +2,648 @@
...
+        assert(cachedSpacing.hSpacing > 0);
+        assert(cachedSpacing.vSpacing > 0);
 
-        int h = ruler->buffer().height();
+        BufferedLineRender<1000> buffer{painter};
 
-        for (qreal x = left; x < m_rect.right(); x += tmpWidth)
+        const double majorLineDistance = cachedSpacing.hSpacing * pixelsPerSceneUnit;
+        const double cutoffDistance    = minorGridTooDenseThreshold * hSubdivisions;
+        if (showMinorGrid && majorLineDistance >= cutoffDistance)
```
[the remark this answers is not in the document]
mariusalexander: done

src/gui/tools/gt_grid.cpp
```diff
@@ -2,483 +2,670 @@
...
+    if (!rect.isValid()) return;
+
+    const double zoom = painter.worldTransform().m11();
 
-    if (val == Qt::Horizontal)
+    if (!pimpl->isVisible || !pimpl->showGrid)
     {
-        length = m_width;
+        // update spacing
+        pimpl->cachedSpacing = scaledGridSpacing(zoom);
```
rainman110: @mariusalexander, this updates `cachedSpacing` without updating `cachedZoom`, so the two cache members can describe different zoom levels. For example, after caching zoom 1, painting a hidden grid at zoom 2 updates only the spacing; a subsequent `setSpacing()` then calls `recomputeChachedSpacing()` with the stale zoom 1. A normal GUI repaint will often repair the state quickly, which makes this difficult to reproduce manually, but direct API use or event ordering before the repaint can expose incorrect snapping. Could we update both cache values together here (for example through the same helper used by `setCurrentZoom()`) and cover the invariant with a focused test such as:

```cpp
TEST(Grid, hidden_paint_keeps_cached_zoom_and_spacing_consistent)
{
    GtGrid grid;
    grid.setScalingStrategy(GtGrid::ScalingStrategy::Base2);
    grid.setSpacing(10);
    grid.setCurrentZoom(1.0);
    grid.setShowGrid(false);

    QImage image{100, 100, QImage::Format_ARGB32};
    QPainter painter{&image};
    painter.scale(2.0, 2.0);
    grid.paint(painter, QRectF{0, 0, 50, 50});

    grid.setSpacing(20);

    EXPECT_EQ(grid.scaledGridSpacing(),
              grid.scaledGridSpacing(2.0));
}
```

The exact expected values are less important than asserting that the no-argument cached result still represents the last painted zoom after a spacing/strategy change.

src/gui/tools/gt_grid.cpp
```diff
@@ -2,483 +2,670 @@
...
+            const double cutoffHDistance    = minorGridTooDenseThreshold * hSubdivisions;
+            const double cutoffVDistance    = minorGridTooDenseThreshold * vSubdivisions;
+
+            painter.setPen(minorPen);
+            if (majorLineDistance >= cutoffHDistance)
+            {
+                const double tmpHMinorSpacing = cachedSpacing.hSpacing / static_cast<double>(hSubdivisions);
+                paintHGridLinesImpl(rect, tmpHMinorSpacing, buffer);
+            }
+            if (majorLineDistance >= cutoffVDistance)
```
rainman110: @mariusalexander, the directions are split now, but this vertical decision still uses `majorLineDistance`, which is calculated from `hSpacing`. With asymmetric spacing, the vertical minor grid can therefore still be enabled or suppressed based on the horizontal density. Could you calculate separate horizontal and vertical major distances and use the vertical one here? A focused rendering test with strongly different `hSpacing`/`vSpacing` and subdivision counts would protect this path.

docs/dev/howtos/migration_2-1.rst
```diff
@@ -0,0 +1,398 @@
...
+     - Use ``scaledGridSpacing().hSpacing`` / ``scaledGridSpacing().vSpacing``.
+   * - ``GtGrid::update`` signal
+     - 2.1.0
+     - Use ``GtGrid::updated`` signal.
+   * - ``GtRuler()`` constructor
+     - 2.1.0
+     - Use ``GtRuler(Qt::Orientation, QWidget* parent)``.
+   * - ``GtRuler::buffer()`` (public, returned ``QImage``)
+     - 2.1.0
+     - Use ``GtRuler::cache()`` (now protected, returns ``QPixmap``).
```
rainman110: @mariusalexander, `buffer()` was removed rather than deprecated, and `cache()` is protected, so existing external callers cannot use the documented replacement. Could you move this entry to the breaking-changes table and state that there is no public replacement because the cache is now an internal implementation detail? Pointing users to a protected method would otherwise leave them with migration code that cannot compile.

docs/dev/howtos/migration_2-1_graphics_api.rst
```diff
@@ -0,0 +1,133 @@
...
+4. Update grid scaling API: ``setScaleGrid(bool)`` -> ``setScalingStrategy``.
+5. Update ruler constructor: ``GtRuler()`` -> ``GtRuler(Qt::Orientation)``.
+
+.. note::
+    Use preprocessor-macros for version dependent code:
+
+    .. code-block:: cpp
+
+        #if GT_VERSION < GT_VERSION_CHECK(2, 1, 0)
+            grid->setVerticalRuler(hruler);
```
rainman110: @mariusalexander, the vertical ruler call receives `hruler` here; this should be `vruler`. The same copy-and-paste error is present in the corresponding version-dependent example in `migration_2-1.rst`.

docs/dev/howtos/migration_2-1.rst
```diff
@@ -0,0 +1,398 @@
...
+     - **Replacement / Notes**
+   * - ``GtGraphicsScene``
+     - 2.1.0
+     - Use ``QGraphicsScene`` directly.
+   * - ``GtGraphicsScene::findItems<T>()``
+     - 2.1.0
+     - Use ``gt::gui::findGraphicsItems<T>()``.
+   * - ``GtGraphicsView::setHorizontalRuler`` / ``setVerticalRuler(GtRuler*)``
+     - 2.1.0
+     - Use ``GtGraphicsView::connectHorizontalRuler`` / ``connectHorizontalRuler(ruler)``. Do not take ownership.
```
rainman110: @mariusalexander, the second replacement should be `GtGraphicsView::connectVerticalRuler(ruler)`. As written, the migration table directs both old setter calls to the horizontal connection.

docs/dev/howtos/migration_2-1.rst
```diff
@@ -0,0 +1,398 @@
...
+.. rst-class:: compact-table
+
+.. list-table::
+   :header-rows: 1
+   :widths: 40 10 50
+
+   * - **Removed API**
+     - **Since**
+     - **Replacement / Notes**
+   * - Removed ``GtGraphicsView::repaintRuler(ruler)``
```
rainman110: @mariusalexander, the removed method had the signature `GtGraphicsView::repaintRuler()` without a ruler argument. Could you document the exact old signature so users can reliably find and migrate their existing calls?

docs/dev/howtos/migration_2-1_graphics_api.rst
```diff
@@ -0,0 +1,133 @@
...
+
+- ``ScalingStrategy::Fixed`` - No automatic scaling (like ``setScaleGrid(false)``).
+- ``ScalingStrategy::Base2`` - Scales in step sizes of 2 (like old ``setScaleGrid(true)`` default).
+- ``ScalingStrategy::Base10`` - Scales in step sizes of 10.
+- ``ScalingStrategy::OneTwoFive`` - Scales in step sizes of 1, 2, 5, 10, 20, 50, etc. New default.
+
+Grid spacing methods:
+
+- ``GtGrid::setGridWidth`` / ``setGridHeight`` -> ``GtGrid::setHSpacing`` / ``setVSpacing`` (or ``GtGrid::setSpacing`` for both).
+- ``GtGrid::scaledGridSpacing`` return a ``GtGridSpacing`` struct with ``hSpacing`` and ``vSpacing`` members denoting the spacing in pixels for the current zoom level.
```
rainman110: @mariusalexander, these members are scene-coordinate spacings for the given zoom, not pixel values. For example, a spacing of 5 scene units at zoom 2 corresponds to 10 device pixels. Could you describe them as scene-unit/grid-coordinate spacings so callers do not use them incorrectly for snapping or painting?

docs/dev/howtos/migration_2-1_graphics_api.rst
```diff
@@ -0,0 +1,133 @@
...
+    auto* vruler = new GtRuler(Qt::Vertical);
+    view->connectHorizontalRuler(hruler);    // does not take ownership
+    view->connectVerticalRuler(vruler);      // does not take ownership
+
+    auto* lay = new QGridLayout();
+    lay->setContentsMargins(0, 0, 0, 0);
+    lay->setSpacing(0);
+    lay->addWidget(hruler, 0, 1);
+    lay->addWidget(vruler, 1, 0);
+    lay->addWidget(view, 1, 1);              // layout owns rulers and view
```
rainman110: @mariusalexander, a layout does not own its widgets; after the layout is installed, the containing widget owns/reparents them. In this snippet the parentless layout is never installed at all, so the example as shown leaks the layout and leaves the ownership claim incorrect. Could you add a container widget and `container->setLayout(lay)` (or show the surrounding existing container) and adjust the comment accordingly?

tests/unittests/gui/test_gt_grid.cpp
```diff
@@ -0,0 +1,486 @@
...
+// test suit for different configurations of grids
+INSTANTIATE_TEST_SUITE_P(
+    Grid,
+    GridScaledSpacingTest,
+    testing::Values(
+        [](GtGrid& grid){ }, // default initialized grid
+        [](GtGrid& grid){ grid.setHSpacing(50.0); }, // custom hspacing
+        [](GtGrid& grid){ grid.setVSpacing(25.0); }, // custom vspacing
+        [](GtGrid& grid){ grid.setHSubdivisions(5); }, // custom hsubdivs
+        [](GtGrid& grid){ grid.setHSubdivisions(3); }, // custom vsubdivs
```
rainman110: @mariusalexander, this case is labelled as the custom vertical-subdivision configuration but calls `setHSubdivisions()` again. Please use `setVSubdivisions(3)` so the new parameterized suite actually covers asymmetric vertical subdivisions.

docs/dev/howtos/migration_2-1_graphics_api.rst
```diff
@@ -0,0 +1,133 @@
...
+- **Old**: ``GtGrid`` was constructed with a ``QGraphicsView&`` reference and
+  the grid was managed internally with unclear ownership semantics.
+- **New**: ``GtGrid`` uses standard Qt parent-child ownership.
+  ``GtGraphicsView::setGrid(GtGrid*)`` takes ownership. The grid is no longer fixed to ``GtGraphicsView``.
+
+Example (old)::
+
+    auto* scene = new QGraphicsScene();
+    auto* view = new GtGraphicsView(scene);  // view destroys its scene, could not work without a scene
+
```
rainman110: @mariusalexander, `git diff --check` currently reports trailing whitespace here and on lines 54, 60, 82, 86, and 88, plus line 282 of `migration_2-1.rst`. Could you clean these up before merge?

docs/dev/howtos/migration_2-1.rst
```diff
@@ -0,0 +1,398 @@
...
+- Replace all deprecated types with modern counterparts.
+
+Related Pages
+^^^^^^^^^^^^^
+
+
+.. toctree::
+   :maxdepth: 1
+
+   migration_2-1_graphics_api
```
rainman110: @mariusalexander, I find the documentation hierarchy asymmetric here: the 2.1 migration is otherwise organized as one page covering Core, GUI, Qt, and CMake, while only the Graphics API gets a separate child page. This also splits and partly duplicates the graphics migration information between two places. Unless there is a broader plan to give comparable migration topics their own consistently structured pages, could you integrate the graphics details into `migration_2-1.rst` and remove this one-off subpage? Alternatively, the whole migration guide would need a deliberate topic-based subpage structure rather than a special case for Graphics.

docs/dev/howtos/migration_2-1_graphics_api.rst
```diff
@@ -0,0 +1,133 @@
...
+    view->setGrid(grid);                     // transfers ownership to view
+
+    grid->setSpacing(50);                    // sets both horizontal and vertical spacing
+    grid->setSubdivisions(5);                // configure number of minor grid lines
+    
+    QPen minorPen = grid->minorPen();        // allows control over major, minor, and axis pen
+    minorPen.setStyle(Qt::DotLine);
+    grid->setMinorPen(minorPen);
+    
+    grid->setLineColor(Qt::grey);            // same as QPen::setColor on the major pen
```
rainman110:
```suggestion
    grid->setLineColor(Qt::gray);            // same as QPen::setColor on the major pen
```

This is a typo and need to be corrected

rainman110 (changes requested):

src/gui/tools/gt_grid.cpp
mariusalexander: Wow yeah good catch... Getting right to it

src/gui/tools/gt_grid.cpp
mariusalexander: Good catch too!

src/gui/tools/gt_grid.cpp
mariusalexander: Please verify

src/gui/tools/gt_grid.cpp
mariusalexander: Please verify

mariusalexander:
@rainman110 I added a few more buttons to the viewer to test different configurations and I found some issues with the ruler that are fixed now. Also, I refactored how the visibility of the grid is handled. 

I had multiple methods to change whether :
- ...the object itself is visible, includes grid and axis (`setVisible(bool)` or `hide`, `show` and `isVisible()`)
- the grid only is visible (`setGridVisible(bool)` and `isGridVisible()`)
- the minor grid only is visible (`setMinorGridVisible(bool)`  and `isMinorGridVisible()`)
- the axis only is visible (`setAxisVisible(bool)` and `isAxisVisible()`)
- and which axis to display (`setVisibleAxis(Axis)` and `visibleAxis()`)

This was quite messy. Further, `isGridVisible` returned also true if `setGridVisible(true)` and `setVisible(false)`, which was unintuitive.

I now use "enable" and "isEnabled" methods. Also, I left the "isVisible" methods, but made sure that they only return true if the elements are actually visible:

- visibility of the object itself (`setVisible(bool)` or `hide`, `show` and `isVisible()`), by default true.
- control grid (`enableGrid(bool)` and `isGridEnabled()`), default is true. `isGridVisible()` returns true only if the object is visible and the grid is enabled.
- control minor grid only (`enableMinorGrid(bool)` and `isMinorGridEnabled()` ), default is true. `isMinorGridVisible()` returns true only if the (major) grid is visible and the minor grid is enabled and the object itself is visible too.
- control the axis only (`setActiveAxis(Axis)` and `activeAxis()` to set whether horizontal and/or vertical are visible). If set to "0" no axis is enabled (the default). `isAxisVisible()` returns true only if any axis is enabled and the object itself is visible. 

Sorry for making a last minute refactoring but I think you'll agree that it is better now this way.

src/gui/tools/gt_gridspacing.h
```diff
@@ -0,0 +1,20 @@
...
+ *  Author: Marius Bröcker (AT-TWK)
+ */
+
+#ifndef GT_GRIDSPACING_H
+#define GT_GRIDSPACING_H
+
+/// Defines the actual spacing of a grid
+struct GtGridSpacing
+{
+    double hSpacing = 0.0, vSpacing = 0.0;
```
rainman110: Please define, what hSpacing and vSpacing means. (Document it)
I suppose
 - hSpacing is the y-Distance between two horizontal lines
 - vSpacing is the x-Distance between two vertical lines

I have the feeling, that it is used **inconsistently**:

 1. In gt_grid.h, it implies, that hspacing is the distance between two vertical lines and vice versa though!

https://github.com/dlr-gtlab/gtlab-core/blob/426c109f8be243559bdb778d6ae00e7cc0062efe/src/gui/tools/gt_grid.h#L72-L82

 2. on the other hand, the documentation of `setGridWidth` and `setGridHeight` say something different.
 
`computeNearestGridPoint()` uses `hSpacing` for X distances und `vSpacing` for Y distances, like discribed by 1.

https://github.com/dlr-gtlab/gtlab-core/blob/426c109f8be243559bdb778d6ae00e7cc0062efe/src/gui/tools/gt_grid.cpp#L308-L323

IMHO,
 - `vSpacing` should always be associated with *width* 
 - `hSpacing` with *height*.

**TLDR**: the interpretation of vSpacing and hSpacing is not consistent in the code!

src/gui/tools/gt_grid.h
```diff
@@ -2,197 +2,502 @@
...
+     */
+    bool isMinorGridEnabled() const;
 
-    /// Hide grid
-    bool m_hideGrid;
+    GT_DEPRECATED_REMOVED_IN(2, 2, "use `setActiveAxis` instead.")
+    Q_INVOKABLE void setShowAxis(bool show)
+    {
+        // not using setAxisVisible for backwards compatibility
+        setActiveAxis(show ? ActiveAxis{} : Qt::Horizontal);
```
rainman110: @mariusalexander, this deprecated compatibility adapter currently inverts the old behavior: `setShowAxis(true)` selects an empty `ActiveAxis`, while `setShowAxis(false)` enables the horizontal axis. The old implementation showed the horizontal axis for `true`. Could you reverse the ternary to `show ? Qt::Horizontal : ActiveAxis{}` ?

docs/dev/howtos/migration_2-1.rst
```diff
@@ -0,0 +1,512 @@
...
+.. rst-class:: compact-table
+
+.. list-table::
+   :header-rows: 1
+   :widths: 40 10 50
+
+   * - **Removed API**
+     - **Since**
+     - **Replacement / Notes**
+   * - Removed ``GtGraphicsView::repaintRuler(ruler)``
```
rainman110: @mariusalexander, the two removed ruler-related signatures are still swapped in the migration table. On `master`, the view method is `GtGraphicsView::repaintRuler()` without an argument, while the grid method below is `GtGrid::paintRuler(GtRuler*)`. Could you document both old signatures exactly so users can find and migrate their existing calls reliably?

docs/dev/howtos/migration_2-1.rst
```diff
@@ -0,0 +1,512 @@
...
+    auto* vruler = new GtRuler(Qt::Vertical);
+    view->connectHorizontalRuler(hruler);    // does not take ownership
+    view->connectVerticalRuler(vruler);      // does not take ownership
+
+    auto* lay = new QGridLayout(this);
+    lay->setContentsMargins(0, 0, 0, 0);
+    lay->setSpacing(0);
+    lay->addWidget(hruler, 0, 1);
+    lay->addWidget(vruler, 1, 0);
+    lay->addWidget(view, 1, 1);              // layout owns rulers and view
```
rainman110: @mariusalexander, parenting the layout with `new QGridLayout(this)` fixes the lifetime issue, but the comment is still inaccurate: the layout does not own the ruler and view widgets. Once they are added to the installed layout, its parent widget (`this`) owns/reparents them. Could you adjust the comment accordingly?

rainman110 (changes requested):
@mariusalexander Thanks for your update. Also, thanks for addressing the previous findings.  

I agree with the refactoring! The resulting API is clearer and more consistent.
The cache handling, directional minor-grid cutoff, migration structure, and related tests are substantially improved.

I found a few remaining issues concerning the deprecated axis adapter (i.e. the interpretation of vSpacing and hSpacing), migration signatures, and documented ownership. I also added a separate comment about keeping painting, rulers, and snapping consistent for asymmetric grid spacing.

docs/dev/howtos/migration_2-1.rst
mariusalexander: fixed

docs/dev/howtos/migration_2-1.rst
mariusalexander: now fixed

src/gui/tools/gt_gridspacing.h
mariusalexander: only the comment was inconsistent. Good catch, it is now fixed.

src/gui/tools/gt_grid.h
mariusalexander: good catch, fixed

src/gui/tools/gt_gridspacing.h
rainman110: But please write in the header file of GtGridSpacing, how it should be interpreted.So what is hspace and vspace?

src/gui/tools/gt_grid.cpp
```diff
@@ -2,483 +2,676 @@
...
+            buffer.flush();
+        }
 
-            /* horizontal ticks */
-            QByteArray tick;
-            tick.setNum(x);
-            QString tstr(tick);
+        painter.setPen(majorPen);
+        paintHGridLinesImpl(rect, cachedSpacing.hSpacing, buffer);
+        paintVGridLinesImpl(rect, cachedSpacing.vSpacing, buffer);
```
rainman110: I understood, that hSpacing is the horizintal space between two vertical lines. If this is the case, then the spacing args in these lines need to be exchanged, right?

src/gui/tools/gt_ruler.cpp
```diff
@@ -2,114 +2,294 @@
...
+    int height{};
+
+    double mult = isAxisFlipped() ? -1 : 1;
+
+    QTransform adjustedTransform;
+
+    switch (orientation())
+    {
+    case Qt::Horizontal:
+        lineDistance = spacing.vSpacing;
```
rainman110: if i understand it right, this draws a horinzontal ruler. then you also need the hSpacing here (i.e. the distance between two vlines)

src/gui/tools/gt_ruler.cpp
```diff
@@ -2,114 +2,294 @@
...
+        height = cache().height();
+        adjustedTransform = viewportTransform;
+        break;
+
+    case Qt::Vertical:
+        painter.rotate(90); // rotating to operate on pixmap as if it is horizontal
+        painter.translate(cache().height(), 0); // move to end of pixmap
+        painter.scale(-1, -1); // flip 180 degrees (so that the text is rotated correctly)
+
+        lineDistance = spacing.hSpacing;
```
rainman110: Same as above.

src/gui/tools/gt_grid.cpp
```diff
@@ -2,483 +2,676 @@
...
-        int h = ruler->buffer().height();
+        BufferedLineRender<1000> buffer{painter};
 
-        for (qreal x = left; x < m_rect.right(); x += tmpWidth)
+        // draw also minor grid lines
+        if (showMinorGrid)
         {
-            QPoint tmp = m_view.mapFromScene(QPointF(x, 0));
-            painter.drawLine(tmp.x(), h - 5, tmp.x(), h);
+            const double majorHLineDistance = cachedSpacing.hSpacing * pixelsPerSceneUnit;
```
rainman110, line 250: Also here, you need to exchange h and v. The horizontal sub-lines need to be painted with the vSubdivisions and vSpacing args.

src/gui/tools/gt_grid.cpp
```diff
@@ -2,483 +2,676 @@
...
+            const double cutoffHDistance    = minorGridTooDenseThreshold * hSubdivisions;
+
+            painter.setPen(minorPen);
+            if (majorHLineDistance >= cutoffHDistance)
+            {
+                const double tmpHMinorSpacing = cachedSpacing.hSpacing / static_cast<double>(hSubdivisions);
+                paintHGridLinesImpl(rect, tmpHMinorSpacing, buffer);
+            }
+
+            const double majorVLineDistance = cachedSpacing.vSpacing * pixelsPerSceneUnit;
```
rainman110, line 260: You need to exchange h and v. The vertical sub-lines need to be painted with the hSubdivisions and hSpacing args.

rainman110 (changes requested):
**Major blocker:**
 You still have mixed up (h/v) spacing and subdivisions in ruler and grid painting.

Please fix this before this branch can be merged.

src/gui/tools/gt_grid.cpp
mariusalexander: correct

mariusalexander:
@rainman110 thanks, now I should got it... I feel a bit dumb now haha but labeling the internal methods for painting the grid left-to-right and top-to-bottom made it clear to me

rainman110 (approved):
LGTM

PR #1493, opened 2026-07-13
Outputdock search filter implementation

### LINKED ISSUE #1121: Add search option for outputdock
Currently there is only a filter search. But if you have to find the messages before and after this message they are hard to find.

### LINKED ISSUE #1123: Add filter to output dock to filter for ids
Maybe it might be helpful to activate or deactivate ids in logging output to see only messages of selected modules (maybe via a drop down menu)

jensschmeink:
The current line based search filter is not a solution if the messages above or under the searched line are relevant, too.

## Description
The existing text filter search for the messages is extended in the output dock widget.
The new version looks like this:
image

The filter icons in some of the columns show that there are optional filters.
The one for the level works like the level filter button on the buttom.

The one for the ID is shown here:
image

Based on the existing IDs here a selection can be done which category should be shown or hidden.

Additionaly the existing filter based search for the messages is now part of the filter popup of the message column:
image

The "old" search widget on the buttom left is now a search function without filter but woth the option to jump to the found results:

image

All lines with matches are highlighted.  The small arrow-buttons or short cuts can be used to jump to the next match.

Overall this helps to search relevant log messages in their overall context.

## How Has This Been Tested?
Manual tests.

## Checklist:

## Summary by Sourcery

Introduce a new log filtering and search concept in the output dock using a dedicated proxy model and header UI.

New Features:
- Add a LogFilterProxyModel to support combined text, level, and category filtering of output log messages.
- Add interactive header filter controls and popup widgets for multi-select level and category filters in the log table.
- Add in-table search highlighting and keyboard/button navigation between message matches in the output dock.

Enhancements:
- Refactor GtFilteredLogModel to delegate text and category filtering to LogFilterProxyModel and act as a bridge for level toggles.
- Sync filter button state with active log levels and categories, preserving deactivated categories across model resets.
- Extend styled log model and icons to expose level data and filter icons needed by the new filtering UI.

Build:
- Register new filter-related source and header files in the app CMake configuration.


jensschmeink:
image
Here is a sample picture: The level column has no filter active, the ID column has "Core" as active selection.

rainman110:
@jensschmeink  Can you please add some screenshots to the PR description and extend the text.

Please describe the different new filters and the search functionality.

jensschmeink:
> @jensschmeink Can you please add some screenshots to the PR description and extend the text.
> 
> Please describe the different new filters and the search functionality.

I added some more description and pictures

src/app/dock_widgets/output/gt_outputdock.cpp
```diff
@@ -239,9 +243,12 @@ GtOutputDock::GtOutputDock()
...
     styleModel->setSourceModel(gtLogModel);
-    m_model = new GtFilteredLogModel(styleModel);
-    m_model->setSourceModel(styleModel);
-    m_model->setFilterCaseSensitivity(Qt::CaseInsensitive);
+
+    auto* filterModel = new gt::LogFilterProxyModel(this);
+    filterModel->setSourceModel(styleModel);
+
+    m_model = new GtFilteredLogModel(filterModel, this);
+    m_model->setSourceModel(filterModel);
```
rainman110: **[P1] Map selected indexes through every proxy layer before Copy/Delete.**

Adding this proxy layer means `m_model->mapToSource(index)` now returns an index owned by `LogFilterProxyModel`, but `openContextMenu()`, `onCopyRequest()`, and `onDeleteRequest()` pass that index directly to `GtLogModel`, which uses its row number to access or remove the raw entry. Once a filter hides earlier rows, that row number identifies a different log entry.

**How to reproduce:**
1. Generate several messages of different levels.
2. Enable a level filter so that earlier source rows are hidden.
3. Select a remaining visible message and use **Copy** or **Delete**.
4. Clear the filter. Copy returns content from a different raw row, or Delete removed a different message.

Please map through `LogFilterProxyModel` and `GtStyledLogModel` down to `GtLogModel` before performing these actions.

src/app/dock_widgets/output/gt_filteredlogmodel.cpp
```diff
@@ -9,112 +9,314 @@
...
+
+    QSet<QString> availableCategories;
+    const int rowCount = m_filterModel->sourceModel()->rowCount();
+
+    for (int row = 0; row < rowCount; ++row)
+    {
+        const QModelIndex index = m_filterModel->sourceModel()->index(row, 2);
+        const QString category = m_filterModel->sourceModel()->data(index).toString();
+
+        if (!category.isEmpty())
```
rainman110: **[P1] Keep log entries with an empty ID in the default category selection.**

This skips empty category values, but then installs all non-empty categories as an active filter. As soon as any named category exists, `matchesCategoryFilter()` rejects every message whose ID is empty even though the user did not disable it.

**How to reproduce:**
1. Clear the output.
2. Generate the standard log-level test messages (these have no custom ID) and verify they are visible.
3. Generate the custom-ID test messages.
4. The previous messages without an ID disappear automatically.

Please include the empty storage value in the unfiltered category state, or represent the initial state with no active category filter until the user changes the selection.

src/app/dock_widgets/output/gt_outputdock.cpp
```diff
@@ -698,6 +768,79 @@ GtOutputDock::onDeleteRequest()
...
+    int rowCount = m_model->rowCount();
+    
+    for (int row = 0; row < rowCount; ++row)
+    {
+        QModelIndex index = m_model->index(row, 3); // Messages column
+        QString itemText = m_model->data(index, Qt::DisplayRole).toString();
+        
+        if (!text.isEmpty() && itemText.contains(text, Qt::CaseInsensitive))
+        {
+            m_matches.append(index);
```
rainman110: **[P2] Rebuild search matches when the model changes.**

The match list stores ordinary `QModelIndex` objects and is rebuilt only when the search text changes. Insertions are therefore omitted, while removals, resets, and header-filter changes can leave indexes invalid or pointing at different rows.

**How to reproduce:**
1. Enter a jump-search term that matches an existing log message.
2. Emit another log message containing the same term.
3. The new message is not highlighted and F3 never navigates to it.
4. Alternatively, clear/delete/filter rows after searching and then press F3; navigation uses the stale list.

Please retain the search text and rebuild matches on relevant row/model/layout signals, or use persistent indexes where appropriate.

rainman110 (changes requested):
Thanks for the substantial improvement to output filtering and search. I found three functional regressions that should be addressed before merging: filtered Copy/Delete can target the wrong raw row, entries without an ID are hidden when named categories appear, and jump-search keeps stale/incomplete model indexes after the log model changes. Reproduction steps and suggested fix directions are included inline.

src/app/dock_widgets/output/gt_outputdock.cpp
jensschmeink: Adressed in [78471dc](https://github.com/dlr-gtlab/gtlab-core/pull/1493/commits/78471dc57d9cd5832af6d9af016efe3ccd2ac85f)

src/app/dock_widgets/output/gt_outputdock.cpp
jensschmeink: Adressed in [b26d520](https://github.com/dlr-gtlab/gtlab-core/pull/1493/commits/b26d520c23b386b4efdecc0d0978810615a36bf7)

src/app/dock_widgets/output/gt_filteredlogmodel.cpp
jensschmeink: Addressed in [b26d520](https://github.com/dlr-gtlab/gtlab-core/pull/1493/commits/b26d520c23b386b4efdecc0d0978810615a36bf7)

jensschmeink:
@rainman110 I addressed your points. I hope now everything works fine.

rainman110 (dismissed):
Hey @jensschmeink ,

The two previous bugs are fixed nicely: entries with empty categories remain visible, and Copy/Delete now work correctly through the proxy chain. Thanks!

One functional issue is still open:

**Search matches are not rebuilt when the level or message filter changes.**

Steps to reproduce:
1. Enter a search term in the Output Dock that matches messages from multiple log levels.
2. Apply a level or message filter in the header so that some of those matches become hidden.
3. Press F3 to navigate through the matches.

`m_matches` still contains indexes from the previous model state. As a result, matches may be skipped or the wrong rows may be selected. Please also connect `levelFilterChanged` and `filterTextChanged` to rebuilding the match list, or react to the appropriate structural changes of the outer model.

One additional point:

- There are no automated tests for the new filter logic. `LogFilterProxyModel` and `GtFilteredLogModel` can be tested without the GUI. Useful cases would include an empty category ID, combined level/category/text filters, and clearing individual filters.

jensschmeink:
@rainman110 I solved the problem of incorrect search matches after filter changes. There is one point to discuss for the unit tests: currently, the new code is part of the "app" and not one of the libraries. For effective testing, this would need to change. The classes could be moved, but it is also beneficial to have good encapsulation of code related only to the app. What is your opinion?

rainman110:
@jensschmeink Your argument regarding the tests makes sense. Since the filter model currently belongs exclusively to the application, I do not think we should move it into `GTlabGui` and export it solely to make it accessible to the unit-test target. That would turn an internal implementation detail into public API/ABI and create a long-term compatibility obligation that is disproportionate here. 

However, the latest fix introduces another issue in `GtOutputDock::onSearchTextChanged()`: `m_matches` is cleared before `QTimer::singleShot()` schedules the actual search. If the method is called several times before the event loop processes those callbacks, every callback appends its results to the same list without clearing it again. The callbacks also retain older search texts. This can mix or duplicate results from multiple search states.

Steps to reproduce:
1. Enter and remove several characters in the Output Dock search field quickly.
2. Before the UI settles, change the level or message filter.
3. Use F3/Shift+F3 to navigate through the matches.
4. Matches may appear more than once or may no longer correspond to the current search text and filter state.

Please coalesce pending updates into a single deferred refresh. The refresh callback should clear `m_matches` and read the current text from `m_searchWidget` when it actually runs. A reusable single-shot `QTimer` would be a straightforward solution.

For manual validation, please cover the individual level, ID, and message filters; combinations of them; clearing individual/all filters; empty IDs; changing filters during an active search; and rapid search-text changes followed by F3/Shift+F3 navigation.

rainman110:
@jensschmeink Implementation hint: keep one single-shot `QTimer` as a `GtOutputDock` member. Let `updateSearchResults()` only call `timer.start(0)`, which automatically coalesces repeated requests. In the timer callback, read `m_searchWidget->text()`, clear `m_matches`, and rebuild the list from the current model state.

jensschmeink:
@mariusalexander I used the hint by @rainman110 to implement the timer cleaner to collect multiple changes of an event loop. Please check the latest version.

src/app/dock_widgets/output/gt_styledlogmodel.cpp
```diff
@@ -65,6 +65,13 @@ GtStyledLogModel::data(const QModelIndex& index, int role) const
             return {};
         }
         break;
+    case Qt::UserRole:
+        // Pass through UserRole to get raw level data
+        if (col == GtLogModel::columnFromRole(GtLogModel::LevelRole))
+        {
+            return gtLogModel->data(index, GtLogModel::LevelRole);
+        }
+        break;
```
mariusalexander: IMO UserRole should not be used directly. It is opaque to the caller what data is accessed. Maybe you could forward the LogRoles like this?:

```suggestion
    case GtLogModel::LevelRole:
    case GtLogModel::TimeRole:
    case GtLogModel::IdRole:
    case GtLogModel::MessageRole:
        // Pass through log-roles to get raw data
        if (col == GtLogModel::columnFromRole(role))
        {
            return gtLogModel->data(index, role);
        }
        break;
```

src/app/dock_widgets/output/gt_logfilterproxymodel.cpp
```diff
@@ -0,0 +1,297 @@
...
+}
+
+bool
+gt::LogFilterProxyModel::matchesLevelFilter(int source_row,
+                                        const QModelIndex& source_parent) const
+{
+    if (m_filterState.levels.isEmpty()) return true;
+
+    const QModelIndex index = sourceModel()->index(source_row, 0, source_parent);
+    const int level = sourceModel()->data(index, Qt::UserRole).toInt();
```
mariusalexander: See my other comment, You have to use "LevelRole" here instead. Makes this part even easier to understand.
```suggestion
    const int level = sourceModel()->data(index, GtLogModel::LevelRole).toInt();
```

src/resources/icons/filterActive.svg
mariusalexander, line 1: I was a bit confused what this icon should indicate. Mybe consider partially filling the icon instead?

Like this

Image

In GTlab it looks like this then (added both icons to the icon viewer too)

Image

The xml-code:
```xml
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
	<path d="M 0.07617187,0.3046875 8.3300781,11.240234 V 23.574219 H 15.791016 V 11.240234 L 24.042969,0.3046875 Z m 17.16860313,1.4394531 3.909522,0 -6.802735,9.0136714 v 11.376954 h -3.293448 l -0.09431,-11.376954 z" />
</svg>
```

src/app/dock_widgets/output/gt_filteredlogmodel.h
```diff
@@ -12,61 +12,65 @@
...
 
-    enum FilterLevel
-    {
-        FilterTrace   = 1 << 0,
-        FilterDebug   = 1 << 1,
-        FilterInfo    = 1 << 2,
-        FilterWarning = 1 << 3,
-        FilterError   = 1 << 4,
-        FilterFatal   = 1 << 5,
-    };
```
mariusalexander, line 35: Why was this removed? You could think about moving the enum definitions to `GtlogModel` instead if you needed them you at multiple places.

src/app/dock_widgets/output/gt_outputdock.cpp
```diff
@@ -557,21 +623,50 @@ GtOutputDock::onRowsInserted(int start, int last)
...
 {
     updateFilterButtons();
+
+    updateSearchResults();
+}
+
+void
+GtOutputDock::updateSearchResults()
+{
+    m_timer.start(0);
```
mariusalexander: So this just means, trigger the function in the next EventLoop cycle and only once?
Can you give this timer member a more descriptive name? Something like "delayFilterTimer" or "lazyFilterTimer" come to mind.

src/app/dock_widgets/output/gt_outputdock.cpp
```diff
@@ -460,20 +514,32 @@ GtOutputDock::updateFilterButtons()
...
     auto& logger = gt::log::Logger::instance();
 
-    auto const hideLevel = [&logger](QPushButton& btn, gt::log::Level level){
+    // Get current filter state from LogFilterProxyModel
+    auto filterModel = static_cast<gt::LogFilterProxyModel*>(m_model->sourceModel());
+    auto activeLevels = filterModel ? filterModel->levelFilter() : QSet<int>();
+
+    auto const updateLevel = [this, &logger, &activeLevels](QPushButton* btn, gt::log::Level level){
+        if (!btn)
+            return;
```
mariusalexander: Put if in brackets or inline. Also "this" is not used.
```suggestion
    auto const updateLevel = [&logger, &activeLevels](QPushButton* btn, gt::log::Level level){
        if (!btn) return;
```

src/app/dock_widgets/output/gt_logfilterproxymodel.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+    return false;
+}
+
+bool
+gt::LogFilterProxyModel::hasActiveFiltersForColumn(int column) const
+{
+    switch (column)
+    {
+        case 0: // Level column
```
mariusalexander: You could also use this function, it should work in a switch-case statement. I'd resuse exisiting "magic numbers" as much as possible. Same applies to the other cases in this witch statement.
```suggestion
        case GtLogModel::columnFromRole(GtLogModel::LevelRole):
```

src/app/dock_widgets/output/gt_filteredlogmodel.h
```diff
@@ -12,61 +12,65 @@
...
  * @brief The GtFilteredLogModel class.
- * Filter proxy model for filtering out certain log levels
+ * 
+ * Simplified filter model that only handles level filters (trace, debug, info, etc.).
+ * Text search and category filters are now handled by LogFilterProxyModel.
+ * This class acts as a bridge between the UI toggle buttons and the filtering logic.
  */
 class GtFilteredLogModel : public QSortFilterProxyModel
 {
     friend class GtOutputDock;
```
mariusalexander: Is this friendship needed? Now GtFIlteredLogModel is only really "useful" for the output dock

src/app/dock_widgets/output/gt_filteredlogmodel.cpp
```diff
@@ -9,112 +9,305 @@
...
+    {
+        m_filterModel->setCategoryFilter(updatedCategories);
+    }
+}
+
+bool
+GtFilteredLogModel::filterAcceptsRow(int srcRow,
+                                     const QModelIndex& srcParent) const
+{
+    return true;
```
mariusalexander: Do I understand correctly, that this class does no longer filter anything on its own? Somehow the proxy filter model is the model that filters the messages, right? Can't we combine the two models? This seems quite confusing and redundant to me having this separation.

src/app/dock_widgets/output/gt_filteredlogmodel.cpp
```diff
@@ -9,112 +9,305 @@
...
-    invalidate();
+    if (m_filterModel)
+    {
+        m_filterModel->setFilterText(val);
+    }
 }
 
 void
-GtFilteredLogModel::setFilter(FilterLevel level, bool enabled)
+GtFilteredLogModel::setFilter(int levelBit, bool enabled)
```
mariusalexander: In my opinion this function and the bit-wise operations are a regression in code readability. Maybe consider using QSet<gt::log::Level> instead? This could work and simplify the logic substantially.

src/app/dock_widgets/output/gt_filteredlogmodel.h
```diff
@@ -77,28 +81,34 @@ public slots:
      */
     void filterData(const QString& val);
```
mariusalexander, line 116: Please add comments to the functions. Also I find the names quite confusing, maybe you can find more descriptive names?  This includes `saveAndPreserveDeactivatedCategories`

src/app/dock_widgets/output/gt_logfilterproxymodel.cpp
```diff
@@ -0,0 +1,297 @@
...
+                seenStorage.insert(storageValue);
+            }
+        }
+    }
+
+    return result;
+}
+
+QStringList
+gt::LogFilterProxyModel::availableLevels() const
```
mariusalexander: This function is not used as far as I can tell. If the names of the Logging Levels are needed at multiple locations (the output dock also uses these), maybe move their definition to the GtLogModel instead to allow reusing the names as much as possible.

src/app/dock_widgets/output/gt_outputdock.cpp
```diff
@@ -557,21 +623,50 @@ GtOutputDock::onRowsInserted(int start, int last)
...
     scrollToBottom();
+    m_model->updateCategoryFilter();
+
+    updateSearchResults();
+}
+
+void
+GtOutputDock::onModelAboutToBeReset()
+{
+    m_model->saveAndPreserveDeactivatedCategories(m_model->filterModel()->categoryFilter());
```
mariusalexander: Your moving QSets between the two models via the output dock, It seems to me that one cannot use the FilterModel without the ProxyFilterModel. If we could combine both filter models into one model a lot of complexity could be removed I think

src/app/dock_widgets/output/gt_logfilterproxymodel.cpp
```diff
@@ -0,0 +1,297 @@
...
+            return false;
+        case 3: // Message column
+            return !m_filterState.text.isEmpty();
+        default:
+            return false;
+    }
+}
+
+QList<QPair<QString, QString>>
+gt::LogFilterProxyModel::availableCategoriesWithStorage() const
```
mariusalexander: I dont like from a code readability point of view that this functions returns a quite complex object: a list of pairs of strings. I think this causes issues with the function name, which to me is quite unclear/confusing. This function is only needed by the `GtFilterHeaderView`. Maybe, instead of returning one complex object, that is tailored to the use case in the `GtFilterHeaderView` class, consider adding more smaller functions. Like a `allCategories` function and a `categroyToDisplayText` function and similar. While this causes you to have a slight more complex logic in `GtFilterHeaderView`, I think the code readability would improve and make the interface of this class more approachable.

src/app/dock_widgets/output/gt_logfilterproxymodel.h
```diff
@@ -0,0 +1,108 @@
...
+
+signals:
+    void levelFilterChanged(const QSet<int>& levels);
+    void categoryFilterChanged(const QSet<QString>& categories);
+    void filterTextChanged(const QString& text);
+
+public:
+    explicit LogFilterProxyModel(QObject* parent = nullptr);
+
+    /// Set text filter (applied only to message column)
```
mariusalexander: All these setters are there because of GtFilterLogModel if I am not mistaken. This makes the API of this class quite complex. I would expect a `filterLogLevel` function here or similar (which is part of `GtFilteredLogModel`) instead of having to set the filtered log levels using a QSet.

mariusalexander (changes requested):
@jensschmeink I find this PR quite complex. I see the main issue in the following areas:

- As I have noted in multiple comments, it is not clear to me why both filter models are necessary and how the work together. Maybe I am missing something, but I think the complexity of this PR and the filter-logic could be reduced significantly If you could merge the two models. 

- Also I find that there are quite a few "magic numbers", like 0 indicating the LogLevel, 1 the etc. Or 1 << 0 indicating TraceLevel etc.

- Similarly, a `QSet<int>` is much less clear to me than a `QSet<gt::log::Level>` would, or better still a `QFlags<LogLevelFilter>` object (where `LogLevelFilter` is basically the previous `FilterLevel` of the `GtFilteredLogModel` class).

mariusalexander:
@jensschmeink Note: I have pushed a small commit in which i have added icons to the icon browser and removed the inkscape data, that is not needed. The GtSvgIconEngine needs to parse the svg string so keeping it as simple as possible is best.

src/app/dock_widgets/output/gt_filteredlogmodel.h
mariusalexander: I failed to highlight the lines properly it seems. I mean these functions
```cpp
    void setCategoryFilterWithSave(const QSet<QString>& categories);

    void resetCategoryFilter();

    void updateCategoryFilter();

    ....

    void saveAndPreserveDeactivatedCategories(const QSet<QString>& currentActivated = {});
```

src/resources/icons/filterActive.svg
jensschmeink: The icon was an idea of @rainman110. For me the icon is not umportant and might be easily replaced also in a later version.

src/app/dock_widgets/output/gt_styledlogmodel.cpp
jensschmeink: As both are normal enums would it not be dangerous as the Qt::roles and the loglevelRoles might iterfere?

jensschmeink:
@mariusalexander I did a huge cleanup to rearrange the filter models to one class. I hope this is what you expected

src/app/dock_widgets/output/gt_styledlogmodel.cpp
mariusalexander: The LogRole-enum is defined in `GtLogModel` as such:
```cpp    
    /// custom roles to access the data directly
    enum LogRole
    {
        InvalidRole = Qt::UserRole, // invalid role
        LevelRole,                  // role to access level as int
        TimeRole,                   // role to access time as string
        IdRole,                     // role to access logging id
        MessageRole                 // role to access message
    };
```
It uses `Qt::UseRule` as a starting point (see https://doc.qt.io/qt-6/qt.html#ItemDataRole-enum). Qt-Roles are "below" Qt::UserRole (<256) and any user-defined roles should be "above"/equal to Qt::UserRole (>=256)

src/app/dock_widgets/output/gt_filteredlogmodel.h
```diff
@@ -12,61 +12,134 @@
...
+const QSet<int> allLevels = {
+    gt::log::TraceLevel,
+    gt::log::DebugLevel,
+    gt::log::InfoLevel,
+    gt::log::WarningLevel,
+    gt::log::ErrorLevel,
+    gt::log::FatalLevel
+};
+
+const QString emptyIDText = "EmptyID";
```
mariusalexander: Can you make it a static member of `GtFilteredLogModel` and name `EmptyIdText` instead (upper case for the first letter)? I think that is more suitable here. If you integrate the other comments you can remove the namespace here too.

src/app/dock_widgets/output/gt_filteredlogmodel.h
```diff
@@ -12,61 +12,134 @@
...
+ * @brief Log Column Constants
+ * Sammelt alle Magic Numbers für Spalten-Indizes
+ */
+enum class LogColumn
+{
+    Level    = 0,
+    Time     = 1,
+    Category = 2,
+    Message  = 3
+};
```
mariusalexander: I highly suggest moving this definition into `GtLogModel` instead. I suggest making it a non-class enum too and use "XColumn" naming scheme. That way it implicitly converts into an int.

Please update the `GtLogModel::columnFromLevel` and `GtLogModel::columnToRole` functions too.

```cpp
class GtLogModel : ...
{
...
public:

    enum LogColumn : int
    {
        LevelColumn    = 0,
        TimeColumn     = 1,
        CategoryColumn = 2,
        MessageColumn  = 3
    };
...
```

src/app/dock_widgets/output/gt_filteredlogmodel.h
```diff
@@ -12,61 +12,134 @@
...
+enum class LogFilterLevel
 {
-    friend class GtOutputDock;
+    Trace   = 1 << 0,  // 0x01 (1)
+    Debug   = 1 << 1,  // 0x02 (2)
+    Info    = 1 << 2,  // 0x04 (4)
+    Warning = 1 << 3,  // 0x08 (8)
+    Error   = 1 << 4,  // 0x10 (16)
+    Fatal   = 1 << 5   // 0x20 (32)
+};
```
mariusalexander: This is the perfect use case for QFlags, no need for  `QSet<int>`! I highly suggest using `QFlags` instead. However it requires that the enum is implcitly convertable to int (thus it cant be a enum class).

Please move this to the `gt_logmodel.h` (not into the class) and make it a normal enum (not enum class) instead.
I think this enum will be useful for other classes/modules too, but then I'd give it a more general name.
I suggest naming it `LogLevelFlag` instead.

```cpp
namespace gt
{
    /// Enum indicating which log levels are active
    enum LogLevelFlag : int
    {
        /// no log level is active
        NoLogLevelFlag = 0,

        TraceLevelFlag   = 1 << 0,
        DebugLevelFlag   = 1 << 1,
        InfoLevelFlag    = 1 << 2, 
        WarningLevelFlag = 1 << 3,
        ErrorLevelFlag   = 1 << 4,
        FatalLevelFlag   = 1 << 5,

        /// all log level are active
        AllLogLevels = TraceLevelFlag | DebugLevelFlag | InfoLevelFlag | WarningLevelFlag | FatalLevelFlag
    };
    using LogLevelFlags = QFlags<LogLevelFlag>;

}; // namespace gt

class GtLogModel { ... }
```

Benefits: 
- You can now use `LogLevelFlags` (or `QFlags<LogLevelFlag>`) instead of `QSet<int>` and it is much more clear to the read what argument is needed. Plus, it is the Qt-way of doing such things.
- `QFlags` implements a nice API to test if flags are active (`QFlags::testFlag`) and set flags (`QFlags::setFlag`).
- `QFlags` is under the hood a simple int instead of heap allocated QSet (nice to know) -> fast access times, fast to copy
- No need for `const QSet<int> allLevels = { ... }` -> simple use `AllLogLevels` instead

src/app/dock_widgets/output/gt_filteredlogmodel.cpp
```diff
@@ -9,112 +9,494 @@
...
+
+bool
+GtFilteredLogModel::matchesLevelFilter(int source_row,
+                                       const QModelIndex& source_parent) const
+{
+    if (m_filterState.levels.isEmpty()) return true;
+
+    const QModelIndex index = sourceModel()->index(source_row, 
+        static_cast<int>(gt::LogColumn::Level), source_parent);
+    const int level = sourceModel()->data(index, Qt::UserRole).toInt();
```
mariusalexander: Please also address the comment about the `GtLogModel::LevelRole` thingy from the other batch of change requests and update this line to use `GtLogModel::LevelRole`.

mariusalexander (changes requested):
Awesome that you managed to condense the two classes, it seems to have worked nicely!

Last change request. I'd suggest you use the `QFlags` instead of `QSet<int>` and move some definitions around. This will require some changes, but I think this would greatly reduce complexity and improve readability further . 

Also note, I pushed the new icon 👀 I think @rainman110 would agree that this is nicer haha

Once your incorporate the changes I'll do few manual tests and then we can merge it I think.

src/app/dock_widgets/output/gt_filteredlogmodel.h
jensschmeink: I'll replace the usage of this enum by existing functions from logmodel (columnFromRole) to avoid duplication.

src/app/dock_widgets/output/gt_filterheaderview.h
```diff
@@ -0,0 +1,115 @@
...
+ * Id (column 2) columns. Clicking a filter button opens a popup menu
+ * for multi-select filtering.
+ *
+ * The header view maintains minimum column widths (70px) for Level and Id
+ * columns to accommodate the filter buttons and ensure good usability.
+ *
+ * @note Level column uses integer-based filtering (log levels)
+ * @note Id column uses string-based filtering (log IDs/categories)
+ */
+class FilterHeaderView : public QHeaderView
```
mariusalexander: Why do you use a namespace here instead of `GtFilterHeaderView`?

src/app/dock_widgets/output/gt_matchdelegate.h
```diff
@@ -0,0 +1,50 @@
...
+class QModelIndex;
+
+namespace gt
+{
+
+/**
+ * @brief The GtMatchDelegate class is a style delegate to define how to visualize
+ * search matches in the output dock search.
+ */
+class GtMatchDelegate : public QStyledItemDelegate
```
mariusalexander, line 19: Here you use a namespace and the `Gt` prefix

mariusalexander (commented):
Just a few follow-up questions regarding the coding style

mariusalexander (approved):
LGTM, thanks four your work :)

PR #1500, opened 2026-07-21
add optional process execution info to GtCoreProcessExecutor::runTask()

### LINKED ISSUE #1499: 💡 [REQUEST] - Add ProcessExecutor info to process execution in preparation for HPC jobs
### Summary

A map of the a new GtProcessExecutionInfo is integrated into GtCoreProcessExecutor. The new class stores basic information required for running jobs via workload managers like SLURM.
Implementation of jobs (job cards) logic is developed in an own module until readiness. The implementation here is the minimal required change for GTlab core to facilitate further development.

### Improvements

allow logic of jobcards for HPC jobs

### Drawbacks

Task management through TaskRunner/ProcessExecutor needs major overhaul to integrate an abstraction level for different types how GTlab starts tasks.

### Current workarounds

Not really or too complicated

MattSchuff:
Adds class GtProcessExecutionInfo to store execution info of a task (queued/started/finished time), the GtProcessComponent::STATE and the GtObjectMementoDiff to apply to the data model.

## Description
Overloaded runTask() and queueTask() to optionally hand over a GtProcessExecutionInfo.
The process executor stores this object and writes information to it.

In a dedicated module for (HPC) jobs, the job starts in batch mode and creates a GtProcessExecutionInfo. The object pointer is handed over via runTask().
Ownership of GtProcessExecutionInfo object is within module.

Closes #1499

## Checklist:

## Summary by Sourcery

Expose task execution results and timing through an optional GtProcessExecutionInfo object.

New Features:
- Add GtProcessExecutionInfo to capture task queue, start, and completion times, process state, and resulting model changes.
- Allow callers to provide execution metadata when running or queueing tasks.

Enhancements:
- Track and clean up execution metadata associated with queued and running tasks.
- Register execution information with the core object factory for broader model integration.

Tests:
- Add unit coverage for successful, warning, and failed task execution metadata and model diffs.

### DIFF
--- a/src/core/gt_coreprocessexecutor.cpp
+++ b/src/core/gt_coreprocessexecutor.cpp
@@ -70,7 +70,7 @@ GtCoreProcessExecutor::setCoreExecutorFlags(Flags flags)
 
 bool
 GtCoreProcessExecutor::runTask(GtTask* task)
-{    
+{
     return runTask(task, nullptr);
 }
 
--- a/src/core/gt_coreprocessexecutor.cpp
+++ b/src/core/gt_coreprocessexecutor.cpp
@@ -45,7 +45,7 @@ struct GtCoreProcessExecutor::Impl
     QPointer<GtRunnable> currentRunnable;
 
     /// store proc run info per task
-    QHash<GtTask*, GtProcessExecutionInfo*> processExecInfo;
+    QHash<GtTask*, QPointer<GtProcessExecutionInfo>> processExecInfo;
 };
 
 GtCoreProcessExecutor::GtCoreProcessExecutor(QObject* parent, Flags flags) :
@@ -157,6 +157,7 @@ bool
 GtCoreProcessExecutor::terminateAllTasks()
 {
     m_queue.clear();
+    pimpl->processExecInfo.clear();
 
     return !taskCurrentlyRunning() || terminateTask(m_current);
 }
@@ -224,6 +225,7 @@ GtCoreProcessExecutor::queueTask(GtTask* task, GtProcessExecutionInfo* procExcIn
 
     m_queue.append(task);
 
+    cleanupProcessExecutionMapping();
     if (procExcInfo!=Q_NULLPTR)
     {
         procExcInfo->setProcessState(GtProcessComponent::QUEUED);
@@ -244,7 +246,15 @@ GtCoreProcessExecutor::removeFromQueue(GtTask *task)
     {
         m_queue.removeAll(task);
 
-        if (task) task->setStateRecursively(GtTask::NONE);
+        if (task)
+        {
+            task->setStateRecursively(GtTask::NONE);
+        }
+
+        if(processExecutionInfo(task))
+        {
+            pimpl->processExecInfo.remove(task);
+        }
 
         emit queueChanged();
     }
@@ -324,9 +334,10 @@ GtCoreProcessExecutor::execute()
 
         connect(runner, &QObject::destroyed, &eventLoop, &QEventLoop::quit);
 
-        if(pimpl->processExecInfo.contains(m_current))
+        auto procExecInfo = processExecutionInfo(m_current);
+        if(procExecInfo)
         {
-            pimpl->processExecInfo[m_current]->setStartTimeNow();
+            procExecInfo->setStartTimeNow();
         }
 
         // run
@@ -410,12 +421,10 @@ GtCoreProcessExecutor::handleTaskFinishedHelper(
         }
 
 
-        if(pimpl->processExecInfo.contains(task))
+        auto procExecInfo = processExecutionInfo(task);
+        if(procExecInfo)
         {
-            auto _x = pimpl->processExecInfo[task];
-            _x->setProcessState(task->currentState());
-            _x->setDataDiffToMerge(sumDiff);
-            _x->setEndTimeNow();
+            procExecInfo->setDataDiffToMerge(sumDiff);
         }
 
     }
@@ -429,10 +438,6 @@ GtCoreProcessExecutor::handleTaskFinishedHelper(
     }
 
 
-    if(pimpl->processExecInfo.contains(task))
-    {
-        pimpl->processExecInfo.remove(task);
-    }
 
     gtDebugId(GT_EXEC_ID).medium() << __FUNCTION__ << "end";
 }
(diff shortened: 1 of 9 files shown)


src/core/gt_coreprocessexecutor.cpp
```diff
@@ -42,6 +43,9 @@ struct GtCoreProcessExecutor::Impl
 
     /// Pointer to current runnable
     QPointer<GtRunnable> currentRunnable;
+
+    /// store proc run info per task
+    QHash<GtTask*, GtProcessExecutionInfo*> processExecInfo;
```
rainman110: This hash owns neither object, but it stores raw pointers to two QObject-derived types. The PR states that the calling module owns GtProcessExecutionInfo, so it may be destroyed while queued/running and later be dereferenced here. Please make the observed value lifetime-safe (for example QPointer<GtProcessExecutionInfo>) and ensure entries are removed when either object is destroyed. The non-owning contract should also be documented on the public API.

src/core/gt_coreprocessexecutor.cpp
```diff
@@ -379,6 +408,16 @@ GtCoreProcessExecutor::handleTaskFinishedHelper(
             gtWarningId(GT_EXEC_ID) << tr("Failed to apply memento diff!");
             ok = false;
         }
+
+
+        if(pimpl->processExecInfo.contains(task))
```
rainman110: Final state/end time and map cleanup happen only in this helper, but onTaskRunnerFinished() calls it only for successful tasks when result saving is enabled. Failed, terminated, setup-failed, and DryExecution tasks therefore remain recorded as QUEUED without an end time and leave stale entries. Please finalize and erase tracking on every terminal path; attaching the result diff can remain conditional.

src/core/gt_coreprocessexecutor.cpp
```diff
@@ -208,6 +223,15 @@ GtCoreProcessExecutor::queueTask(GtTask* task)
     task->resetMonitoringProperties();
 
     m_queue.append(task);
+
+    if (procExcInfo!=Q_NULLPTR)
+    {
+        procExcInfo->setProcessState(GtProcessComponent::QUEUED);
+        procExcInfo->setQueuedTimeNow();
+        pimpl->processExecInfo[task] = procExcInfo;
```
rainman110, line 233: This association is not erased when removeFromQueue() removes a queued task or when terminateAllTasks() clears the queue. Requeueing the same task without info then reuses the stale mapping, potentially updating an old or already-deleted object. Please centralize cleanup for removal, cancellation, setup failure, and normal completion.

src/core/process_management/gt_processexecutioninfo.h
```diff
@@ -0,0 +1,50 @@
+/* GTlab - Gas Turbine laboratory
+ *
+ * SPDX-License-Identifier: MPL-2.0+
+ * SPDX-FileCopyrightText: 2023 German Aerospace Center (DLR)
```
rainman110: This is a new file created in 2026 (also reflected by the creation date below), so its SPDX copyright year should be 2026. The new .cpp file has the same issue.

rainman110 (changes requested):
The new execution tracking has lifetime and cleanup problems that can leave dangling pointers or incomplete/stale job information. Please address the inline findings and add focused Core tests before merging.
 - Codecov currently reports 6.67% patch coverage (84 changed lines uncovered)
 - and git diff --check reports trailing whitespace in gt_coreprocessexecutor.cpp:73 and gt_processexecutioninfo.h:33.
 - The new files also need their copyright year corrected to 2026. CI build/test jobs otherwise pass.
 
 For future review requests, please mark the PR ready for review first; draft PRs are normally treated as unfinished.

src/core/gt_coreprocessexecutor.cpp
MattSchuff: Considering that the current implementation is only a bridge technology, pulling the lifetime of GtProcessExecutionInfo fully outside of GtCoreProcessExecutor was my minimal invasive approach here.
To address the current issue: I can wrap access to pimpl->procExecInfo into a function which first checks if the pointers are still available, otherwise doing nothing. This is not "perfect", but would avoid SEGFAULTs.

For future remodelling of the task/process management:

My other idea was: I could implement it so that GtCoreProcessExecutor owns the GtProcessExecutionInfo about a task by making a special entry point like `GtProcessExecutionInfo* runTaskAndGetProcessExectionInfo(GtTask* task)` (maybe a better name). This would leave ownership within in GtCoreProcessExecutor. My thought against that approach was, that lifetime issues remain, because after task completion, the GtCoreProcessExecutor forgets about the task completely (the point where it could destroy the GtProcessExecutionInfo). However, the actual usage of GtProcessExecutionInfo is after task clean up, in order to save its memento to a file.

For future remodelling of the task/process management:
Best approach would be to return a handle (UUID or otherwise) when submitting a task for running (under which ID GTlab allows tracking of this exact task execution).
GtCoreProcessExecutor than could be queried for a process execution info. This transfers the lifetime questions to how long the objects (foremost the GtProcessComponent::STATE and the memento diff) are kept in memory after task completion, or if they are stored to disk or otherwise.

src/core/gt_coreprocessexecutor.cpp
MattSchuff: moved partially to onTaskRunnerFinished()

src/core/gt_coreprocessexecutor.cpp
MattSchuff: added handling of pimpl->processExecInfo to removeFromQueue() and terminateAllTasks()

src/core/process_management/gt_processexecutioninfo.h
MattSchuff: done

MattSchuff:
> The new execution tracking has lifetime and cleanup problems that can leave dangling pointers or incomplete/stale job information. Please address the inline findings and add focused Core tests before merging.
> 
>     * Codecov currently reports 6.67% patch coverage (84 changed lines uncovered)
> 
>     * and git diff --check reports trailing whitespace in gt_coreprocessexecutor.cpp:73 and gt_processexecutioninfo.h:33.
> 
>     * The new files also need their copyright year corrected to 2026. CI build/test jobs otherwise pass.
> 
> 
> For future review requests, please mark the PR ready for review first; draft PRs are normally treated as unfinished.

Tests will be added if the implementation is OK.

src/core/process_management/gt_processexecutioninfo.h
```diff
@@ -0,0 +1,50 @@
...
+    const double queuedTime();
+    void setQueuedTime(double queuedTime);
+    void setQueuedTimeNow();
+
+    const double startTime();
+    void setStartTime(double startTime);
+    void setStartTimeNow();
+
+    const double endTime();
+    void setEndTime(double endTime);
```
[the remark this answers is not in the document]
MattSchuff: done

src/core/process_management/gt_processexecutioninfo.cpp
```diff
@@ -0,0 +1,157 @@
...
+}
+
+const double GtProcessExecutionInfo::endTime()
+{
+    return pimpl->endTime.getVal();
+}
+
+void GtProcessExecutionInfo::setEndTime(double endTime)
+{
+    pimpl->endTime.setVal(endTime);
```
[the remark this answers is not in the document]
MattSchuff: done

MattSchuff:
On hold, cross check with #1525

MattSchuff:
Issue withdrawn, using the run_task_from_memento instead.

PR #1506, opened 2026-07-22
Reduce deprecation warnings related to use of old property constructors

### LINKED ISSUE #1505: [CHANGE] - Reduce deprecation warnings related to use of old property constructors
### Summary

The GtIntProperty and GTDoubleProperty have new interfaces for bound definitions. But the old are still used in the code which can be reduced to reduce build warnings

### Impact on the codebase

Cleaner code base

### Current workarounds

-

jensschmeink:
The internal usage of some deprecated functions related to properties should be reduced

## Description
For Int- and IntMonitoring- and DoubleMonitoringProperty some deprecated functions are still used internally. To cleanup the build warnings these usages are removed.

As the code now uses always the new versions this code extensions lead to more lines of code which need more tests to fullfill coverage.

## How Has This Been Tested?
Unittests had been extended to check the new and still the old code.
The tests for the functions which are marked as deprecated are in separated test functions to be removed easier in a later version. 

## Checklist:

## Summary by Sourcery

Reduce internal usage of deprecated property constructors and adjust monitoring properties to rely on non-deprecated base constructors while preserving behavior.

Enhancements:
- Normalize integer property boundary handling directly in constructors and clamp initial values to configured bounds.
- Refactor int and double monitoring property constructors to construct via the underlying base property types and explicitly enable monitoring.
- Suppress deprecation warnings within the int property factory function without changing its existing interface or behavior.

## Summary by Sourcery

Reduce internal usage of deprecated integer and double property constructors and monitoring property helpers while keeping behavior consistent and tightening boundary handling.

Bug Fixes:
- Normalize and validate integer property boundaries so misordered limits now result in well-defined min/max ranges instead of falling back to unbounded defaults.

Enhancements:
- Route deprecated integer property constructors and integer monitoring factory through the new Boundaries-based APIs to centralize boundary logic and clamping.
- Mark several int, double, and monitoring property constructors and factory helpers with versioned deprecation macros for scheduled removal.
- Update monitoring-related codepaths to create standard properties and enable monitoring via flags instead of relying on dedicated monitoring property types.

Tests:
- Extend unit tests for int and double properties and monitoring properties to cover both new Boundaries-based constructors and legacy deprecated paths, including clamping behavior and factory functions.
- Adjust calculator and process test fixtures to use non-deprecated property APIs and monitoring factories while preserving test semantics.

### DIFF
--- a/src/core/process_management/gt_intmonitoringproperty.cpp
+++ b/src/core/process_management/gt_intmonitoringproperty.cpp
@@ -21,7 +21,7 @@ GtIntMonitoringProperty::GtIntMonitoringProperty(const QString& ident,
 
 GtIntMonitoringProperty::GtIntMonitoringProperty(const QString& ident,
                                                  const QString& name) :
-    GtIntProperty(ident, name)
+    GtIntProperty(ident, name, QString())
 {
     setMonitoring(true);
 }
--- a/src/core/process_management/gt_intmonitoringproperty.cpp
+++ b/src/core/process_management/gt_intmonitoringproperty.cpp
@@ -21,9 +21,9 @@ GtIntMonitoringProperty::GtIntMonitoringProperty(const QString& ident,
 
 GtIntMonitoringProperty::GtIntMonitoringProperty(const QString& ident,
                                                  const QString& name) :
-    GtIntMonitoringProperty(ident, name, QString())
+    GtIntProperty(ident, name)
 {
-
+    setMonitoring(true);
 }
 
 gt::PropertyFactoryFunction
--- a/src/core/process_management/gt_doublemonitoringproperty.cpp
+++ b/src/core/process_management/gt_doublemonitoringproperty.cpp
@@ -21,9 +21,9 @@ GtDoubleMonitoringProperty::GtDoubleMonitoringProperty(const QString& ident,
 
 GtDoubleMonitoringProperty::GtDoubleMonitoringProperty(const QString& ident,
                                                        const QString& name) :
-    GtDoubleMonitoringProperty(ident, name, QString())
+    GtDoubleProperty(ident, name)
 {
-
+    setMonitoring(true);
 }
 
 gt::PropertyFactoryFunction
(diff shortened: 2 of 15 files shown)


src/core/process_management/gt_intmonitoringproperty.cpp
```diff
@@ -21,9 +21,9 @@ GtIntMonitoringProperty::GtIntMonitoringProperty(const QString& ident,
 
 GtIntMonitoringProperty::GtIntMonitoringProperty(const QString& ident,
                                                  const QString& name) :
-    GtIntMonitoringProperty(ident, name, QString())
+    GtIntProperty(ident, name)
```
jensschmeink:
```suggestion
    GtIntProperty(ident, name, QString())
```

jensschmeink:
- Build warnings in the core code had been  removed by using only non-deprecated functions
- The changed calls via the new functions needed code extensions which needed tests so the unittests had been extended to fullfill code coverage
- There are more build warnings for the unittests now as the warnings are currently not suppressed there.  Otherwise the later removal might be forgotten. 
- The deprecation warnings for the properties might be extended with a version for the removal. Then the warnings in the test could be suppressed as the compiler would remind us to remove the old tests. A discussion would be needed to clarify which version should be used to remove the support of the old bound-definition and the usage of units for int-properties.

rainman110:
@jensschmeink Please remove draft status

jensschmeink:
Used deprecations will get an exipre date for the old code of GTlab version 2.2

jensschmeink:
Latest work:
- The property related deprecations got an expiration version (2-2).
- Small code cleanup and test extension as suggested by bot review
- The unittests of this function uses the macro to suppress build warnings
- Unittests using the deprecated functions, but only as sideeffect, are now using the new functions to avoid build warnings
- The test modules now uses the new functions to avoid build warnings and to keep them up to date

jensschmeink:
@rainman110 now it should be ready for review

rainman110 (approved):
Thanks for the nice work. Great cleanup. Hope, that we will have no deprecation warnings anymore at some point.

But it is now much less than before!

Issue #1513, opened 2026-07-26
Prove the compatibility-first multi-project slice end to end


rainman110:
Parent epic: #1507

Depends on #1508, #1509, #1510, #1511, and #1512.

## Goal

Validate that the first migration slice provides useful multi-project behavior without requiring a GUI rewrite or immediate module migration.

## Test scenario

Use two projects with distinct marker data and a representative legacy calculator that accesses the project via `gtApp->currentProject()`.

The test must demonstrate:

- execution for project A sees only project A,
- execution for project B sees only project B,
- changing the GUI-selected project does not redirect a running job,
- same-project writes are serialized,
- the existing desktop GUI can still open and execute a project as before.

## Tasks

- Add an integration test covering two project contexts.
- Add a desktop-GUI smoke/regression test where practical.
- Add a batch/headless regression test.
- Document the recommended API for new calculators:
  - explicit execution context/project accessor,
  - legacy compatibility behavior,
  - threading limitations.
- Add a short follow-up section listing deferred architecture work.

## Acceptance criteria

- [ ] The two-project integration scenario passes reliably.
- [ ] A legacy calculator requires no source changes.
- [ ] Existing GUI execution behavior is preserved.
- [ ] Batch/headless execution behavior is preserved.
- [ ] Documentation clearly distinguishes active GUI project from execution project.
- [ ] Deferred items are captured without expanding this epic's scope.

## Deferred follow-ups

- Execution-project snapshots.
- Context-aware async helpers.
- Per-context externalization and state services.
- Full in-process multi-project datamodel.
- Web API and worker orchestration.

Issue #1512, opened 2026-07-26
Serialize mutating executions per project


rainman110:
Parent epic: #1507

Depends on #1511.

## Goal

Provide a safe first concurrency model for a multi-project service: different projects may be scheduled independently, but mutating executions for the same project must not overlap.

## Tasks

- Introduce a stable project execution key based on canonical project identity/path.
- Add a project-scoped execution guard or lock at the service/executor boundary.
- Reject or queue a second mutating execution according to a documented policy.
- Ensure locks are released on success, failure, cancellation, and worker shutdown.
- Prevent project close/save operations from racing with an active execution.
- Expose a structured busy result that a future backend can map to an API response.

## Acceptance criteria

- [ ] Two mutating jobs for the same project cannot execute concurrently.
- [ ] Jobs for different projects are not blocked by the same project lock.
- [ ] A failed or cancelled job releases the lock.
- [ ] Save/close behavior during execution is deterministic and tested.
- [ ] The locking policy and its limitations are documented.

## Non-goals

- Concurrent writes with conflict resolution.
- Distributed locking across service instances.
- Transactional project storage.

Issue #1511, opened 2026-07-26
Install the project context for calculator and task execution


rainman110:
Parent epic: #1507

Depends on #1510.

## Goal

Ensure every calculator execution performed by the Core process engine runs with the correct project context, and stop deriving the execution project path from the GUI-selected project.

## Tasks

- Determine the owning project from the queued task/source before execution.
- Construct the execution context in the process executor or task runner.
- Install the context around the complete calculator execution boundary.
- Pass the project path as job/context data instead of falling back to `gtApp->currentProject()->path()` in `GtRunnable::projectPath()`.
- Ensure context cleanup on success, failure, cancellation, and setup errors.
- Keep the existing executor API working for GUI callers.
- Add tests for sequential jobs belonging to different projects.

## Acceptance criteria

- [ ] A calculator sees the project belonging to its task, regardless of the active GUI project.
- [ ] `GtRunnable::projectPath()` uses execution-specific data during a job.
- [ ] Failed or cancelled runs do not leak the previous context.
- [ ] Existing GUI task execution continues to work without caller changes.
- [ ] Batch execution remains functional.
- [ ] No calculator module changes are required for the tested legacy path.

## Non-goals

- Parallel writes to one project.
- A redesigned scheduler.
- Context propagation into threads created directly by calculators.

Issue #1510, opened 2026-07-26
Resolve currentProject() through the active execution context


rainman110:
Parent epic: #1507

Depends on #1509.

## Goal

Keep existing calculator modules source-compatible by making `gtApp->currentProject()` context-aware.

## Required behavior

1. If an execution context is active, return its project.
2. Otherwise, return the GUI/session active project exactly as today.
3. Outside both contexts, return `nullptr` as today.

`GtCoreDatamodel::currentProject()` and `GtCoreApplication::currentProject()` must have one clearly documented resolution policy and must not diverge accidentally.

## Tasks

- Implement the compatibility resolution policy.
- Preserve the existing GUI fallback.
- Add diagnostics for execution-worker access without an installed context where feasible.
- Add regression tests using the legacy calculator introduced by the baseline issue.
- Document `currentProject()` as a compatibility API; do not deprecate it yet.

## Acceptance criteria

- [ ] Existing source code using `gtApp->currentProject()` continues to compile.
- [ ] A scoped execution observes its associated project.
- [ ] The GUI thread observes its selected project when no execution context is active.
- [ ] Leaving a scope restores the previous result.
- [ ] No blanket deprecation warning is introduced for existing modules.

## Non-goals

- Migrating all Core or module call sites.
- Supporting arbitrary asynchronous callbacks after an execution scope has ended.

PR #1514, opened 2026-07-26
Implement project execution context and per-project mutation locking (#1508-#1513)

rainman110:
## Motivation and scope

GTlab Core currently assumes a single globally selected project in many code paths. This works for the desktop application, but prevents the process engine from safely serving multiple independent projects in a headless runtime or backend service.

The long-term goal is a multi-project runtime in which tasks for different projects can be submitted and executed independently, including through an asynchronous service API. The existing desktop GUI and legacy modules must remain compatible during this migration.

This pull request is the first compatibility-focused step toward that goal. It does not implement the complete runtime or service API. Instead, it:

- separates the execution project from the project selected in the desktop UI;
- introduces an execution-scoped project context;
- preserves existing `currentProject()`-based calculators;
- prevents concurrent mutations of the same project; and
- allows executions for different projects to proceed independently.

Runtime lifecycle management, asynchronous task handles, worker orchestration, multi-process coordination, and the web/backend API remain follow-up work.

## Description

This pull request implements the complete first slice of Epic #1507 (#1508-#1513): compatibility-first project context for multi-project services.

Existing calculators can continue to use `gtApp->currentProject()` without source changes. Without an active execution context, `currentProject()` resolves to the selected GUI/session project. During Core execution, an active `GtExecutionContext` takes precedence: it resolves to the execution project, while an active path-only context deliberately returns `nullptr` and does not fall back to the GUI-selected project.

The change also introduces per-project mutation serialization and deterministic protection for project save/close operations. Projects with a path are identified by their canonical path; pathless projects use their UUID, with a process-local object identity as the final fallback. Batch/headless behavior is preserved.

The compatibility-baseline document is intentionally temporary. It documents the current transition and recommended explicit context API; integration into the regular developer documentation is tracked as follow-up issue #1518.

API and ABI compatibility:

- Existing `currentProject()` call sites and legacy calculator source remain source-compatible.
- The new public `GtExecutionContext`, `GtExecutionContextScope`, and `GtProjectExecutionGuard` types are exported through the Core export mechanism.
- The public `GtExecutionContext` API was intentionally reduced to a borrowed project pointer and project path.
- Existing public methods remain available; no existing module-facing API is removed or renamed.
- The legacy `GtRunnable(QString)` constructor remains available, while the context-aware constructor is additive.
- The change is not presented as drop-in ABI-compatible for already compiled consumers: adding execution-context state to `GtRunnable` changes its class layout, so dependent binaries must be rebuilt. The existing Core executor keeps its state behind its private implementation boundary.

The end-to-end regression scenario uses two projects and a legacy calculator. It verifies that a running execution keeps its project context even when the GUI-selected project changes, while same-project writes remain serialized and different projects can execute independently.

Related issues: #1507, #1508, #1509, #1510, #1511, #1512, #1513

## How Has This Been Tested?

- Focused context, runnable, executor, and project-guard tests: 25 passed, including the path-only context and guard-cleanup regression tests.
- Full `GTlabUnitTest --gtest_shuffle --gtest_random_seed=1514`: completed without test failures; the existing Windows file-locking test remains skipped.
- Developer documentation built successfully.
- Final CI status for commit `53a3aeae`: Ubuntu, Windows, Linux coverage, documentation, user documentation, license compliance, and Sourcery review passed. Qt6/Windows and the GitLab mirror checks were still pending when this description was updated.

## Checklist:

- [x] The changelog has been extended, if this MR contains important changes from the user's point of view.

## Summary by Sourcery

Introduce execution-scoped project resolution and per-project mutation serialization as the compatibility foundation for multi-project Core execution.

New Features:
- Add execution-scoped project contexts that let Core code resolve the active project independently of the GUI-selected project.
- Add process-local per-project execution guards that serialize mutations and permit concurrent execution of different projects.
- Expose execution results so callers can distinguish started, queued, busy, and invalid tasks.

Bug Fixes:
- Prevent project saves and closes from occurring while the project is being executed.
- Ensure execution guards and contexts are released or restored when runs finish, fail, or unwind.

Enhancements:
- Preserve existing currentProject()-based calculator behavior while prioritizing an active execution context and supporting path-only contexts without GUI fallback.
- Maintain the legacy runnable constructor and custom project-path behavior while adding context-aware execution support.

Documentation:
- Add a temporary developer migration guide describing project-context compatibility and explicit context usage.
- Document execution-context project resolution in the changelog and developer documentation index.

Tests:
- Add regression coverage for context scoping, nesting, thread isolation, runnable compatibility, project locking, executor lifecycle cleanup, and independent multi-project execution.

### DIFF
--- a/src/core/gt_coreapplication.cpp
+++ b/src/core/gt_coreapplication.cpp
@@ -948,14 +948,7 @@ GtCoreApplication::settings()
 GtProject*
 GtCoreApplication::currentProject()
 {
-    GtSession* w = session();
-
-    if (w)
-    {
-        return w->currentProject();
-    }
-
-    return nullptr;
+    return m_dataModel ? m_dataModel->currentProject() : nullptr;
 }
 
 GtProject*
--- a/src/core/gt_coredatamodel.h
+++ b/src/core/gt_coredatamodel.h
@@ -56,8 +56,11 @@ public:
     GtSession* session();
 
     /**
-     * @brief Returns current project. If no current project exists NULL will
-     * be returned.
+     * @brief Returns the execution-context project or the session project.
+     *
+     * If a GtExecutionContext is active on the current thread, its project is
+     * returned. Otherwise the currently selected session project is returned;
+     * if neither exists, NULL is returned.
      * @return Current project pointer
      */
     GtProject* currentProject();
--- a/src/core/gt_coreprocessexecutor.cpp
+++ b/src/core/gt_coreprocessexecutor.cpp
@@ -499,15 +499,7 @@ GtCoreProcessExecutor::setupTaskRunner()
     // create new task runner
     auto* runner = new GtTaskRunner{m_current};
 
-    GtProject* project = qobject_cast<GtProject*>(m_source.data());
-    if (!project)
-    {
-        project = m_source->findParent<GtProject*>();
-    }
-    if (!project)
-    {
-        project = m_current->findParent<GtProject*>();
-    }
+    GtProject* project = projectForTask(m_current, m_source);
 
     const QString projectPath = pimpl->customProjectPath.isEmpty() && project ?
                                      project->path() :
@@ -543,8 +535,9 @@ GtCoreProcessExecutor::onTaskRunnerFinished()
 {
     gtTrace() << __FUNCTION__;
 
-    const auto guardCleanup = gt::finally(
-        [this]() { pimpl->projectGuard.reset(); });
+    const auto _ = gt::finally([this]() {
+        pimpl->projectGuard.reset();
+    });
 
     // create timer
     QElapsedTimer timer;
--- a/src/core/gt_coreprocessexecutor.cpp
+++ b/src/core/gt_coreprocessexecutor.cpp
@@ -11,6 +11,7 @@
 #include <QDir>
 #include <QElapsedTimer>
 #include <QEventLoop>
+#include <QScopeGuard>
 
 #include "gt_logging.h"
 #include "gt_runnable.h"
@@ -514,10 +515,7 @@ GtCoreProcessExecutor::setupTaskRunner()
                                      pimpl->customProjectPath;
     GtExecutionContext executionContext(
         project,
-        {},
-        m_source->objectName(),
-        projectPath,
-        m_current->objectName());
+        projectPath);
 
     // create new runnable
     pimpl->currentRunnable = new GtRunnable{
@@ -546,6 +544,10 @@ GtCoreProcessExecutor::onTaskRunnerFinished()
 {
     gtTrace() << __FUNCTION__;
 
+    const auto guardCleanup = qScopeGuard(
+        [this]() { pimpl->projectGuard.reset(); });
+    Q_UNUSED(guardCleanup);
+
     // create timer
     QElapsedTimer timer;
 
@@ -566,7 +568,6 @@ GtCoreProcessExecutor::onTaskRunnerFinished()
     if (!m_current)
     {
         gtFatalId(GT_EXEC_ID) << tr("Current task corrupted!");
-        pimpl->projectGuard.reset();
         taskRunner->deleteLater();
         return;
     }
@@ -591,8 +592,6 @@ GtCoreProcessExecutor::onTaskRunnerFinished()
         }
     }
 
-    pimpl->projectGuard.reset();
-
     gtInfoId(GT_EXEC_ID).medium()
         << tr("----> Task finished (took %1 ms to merge) <----")
                .arg(timer.elapsed());
@@ -603,5 +602,7 @@ GtCoreProcessExecutor::onTaskRunnerFinished()
     // delete task runner
     taskRunner->deleteLater();
 
+    // Release before starting a queued task so it can acquire its own key.
+    pimpl->projectGuard.reset();
     executeNextTask();
 }
--- a/src/core/gt_coreprocessexecutor.cpp
+++ b/src/core/gt_coreprocessexecutor.cpp
@@ -11,8 +11,6 @@
 #include <QDir>
 #include <QElapsedTimer>
 #include <QEventLoop>
-#include <QScopeGuard>
-
 #include "gt_logging.h"
 #include "gt_runnable.h"
 #include "gt_project.h"
@@ -22,6 +20,7 @@
 #include "gt_objectmementodiff.h"
 #include "gt_coreapplication.h"
 #include "gt_executioncontext.h"
+#include "gt_finally.h"
 #include "gt_projectexecutionguard.h"
 #include "gt_taskrunner.h"
 
@@ -544,9 +543,8 @@ GtCoreProcessExecutor::onTaskRunnerFinished()
 {
     gtTrace() << __FUNCTION__;
 
-    const auto guardCleanup = qScopeGuard(
+    const auto guardCleanup = gt::finally(
         [this]() { pimpl->projectGuard.reset(); });
-    Q_UNUSED(guardCleanup);
 
     // create timer
     QElapsedTimer timer;
--- a/src/core/gt_coreprocessexecutor.cpp
+++ b/src/core/gt_coreprocessexecutor.cpp
@@ -20,6 +20,7 @@
 #include "gt_task.h"
 #include "gt_objectmementodiff.h"
 #include "gt_coreapplication.h"
+#include "gt_executioncontext.h"
 #include "gt_taskrunner.h"
 
 #include "gt_coreprocessexecutor.h"
@@ -446,8 +447,29 @@ GtCoreProcessExecutor::setupTaskRunner()
     // create new task runner
     auto* runner = new GtTaskRunner{m_current};
 
+    GtProject* project = qobject_cast<GtProject*>(m_source.data());
+    if (!project)
+    {
+        project = m_source->findParent<GtProject*>();
+    }
+    if (!project)
+    {
+        project = m_current->findParent<GtProject*>();
+    }
+
+    const QString projectPath = pimpl->customProjectPath.isEmpty() && project ?
+                                     project->path() :
+                                     pimpl->customProjectPath;
+    GtExecutionContext executionContext(
+        project,
+        {},
+        m_source->objectName(),
+        projectPath,
+        m_current->objectName());
+
     // create new runnable
-    pimpl->currentRunnable = new GtRunnable{pimpl->customProjectPath};
+    pimpl->currentRunnable = new GtRunnable{
+        pimpl->customProjectPath, std::move(executionContext)};
 
     // setup task runner
     if (!runner->setUp(pimpl->currentRunnable, m_source))
--- a/src/core/gt_coreprocessexecutor.cpp
+++ b/src/core/gt_coreprocessexecutor.cpp
@@ -85,15 +85,8 @@ GtCoreProcessExecutor::setCoreExecutorFlags(Flags flags)
     pimpl->detached =  flags.testFlag(gt::NonBlockingExecution);
 }
 
-bool
-GtCoreProcessExecutor::runTask(GtTask* task)
-{
-    const auto result = runTaskWithResult(task);
-    return result == RunTaskResult::Started || result == RunTaskResult::Queued;
-}
-
 GtCoreProcessExecutor::RunTaskResult
-GtCoreProcessExecutor::runTaskWithResult(GtTask* task)
+GtCoreProcessExecutor::startTask(GtTask* task)
 {
     gtDebugId(GT_EXEC_ID).medium() << __FUNCTION__;
 
@@ -107,17 +100,11 @@ GtCoreProcessExecutor::runTaskWithResult(GtTask* task)
         return RunTaskResult::Queued;
     }
 
-    return executeNextTaskWithResult();
-}
-
-bool
-GtCoreProcessExecutor::executeNextTask()
-{
-    return executeNextTaskWithResult() == RunTaskResult::Started;
+    return startNextTask();
 }
 
 GtCoreProcessExecutor::RunTaskResult
-GtCoreProcessExecutor::executeNextTaskWithResult()
+GtCoreProcessExecutor::startNextTask()
 {
     // check whether a task is already running
     if (taskCurrentlyRunning())
@@ -519,7 +506,7 @@ GtCoreProcessExecutor::setupTaskRunner()
         delete pimpl->currentRunnable;
         delete runner;
         clearCurrentTask();
-        executeNextTask();
+        startNextTask();
         return nullptr;
     }
 
@@ -595,5 +582,5 @@ GtCoreProcessExecutor::onTaskRunnerFinished()
 
     // Release before starting a queued task so it can acquire its own key.
     pimpl->projectGuard.reset();
-    executeNextTask();
+    startNextTask();
 }
--- a/src/core/gt_coreprocessexecutor.cpp
+++ b/src/core/gt_coreprocessexecutor.cpp
@@ -85,39 +85,39 @@ GtCoreProcessExecutor::setCoreExecutorFlags(Flags flags)
     pimpl->detached =  flags.testFlag(gt::NonBlockingExecution);
 }
 
-GtCoreProcessExecutor::RunTaskResult
+GtCoreProcessExecutor::TaskExecState
 GtCoreProcessExecutor::startTask(GtTask* task)
 {
     gtDebugId(GT_EXEC_ID).medium() << __FUNCTION__;
 
     if (!queueTask(task))
     {
-        return RunTaskResult::Invalid;
+        return TaskExecState::Invalid;
     }
 
     if (taskCurrentlyRunning())
     {
-        return RunTaskResult::Queued;
+        return TaskExecState::Queued;
     }
 
     return startNextTask();
 }
 
-GtCoreProcessExecutor::RunTaskResult
+GtCoreProcessExecutor::TaskExecState
 GtCoreProcessExecutor::startNextTask()
 {
     // check whether a task is already running
     if (taskCurrentlyRunning())
     {
         gtErrorId(GT_EXEC_ID) << tr("A Task is already running!");
-        return RunTaskResult::Busy;
+        return TaskExecState::Busy;
     }
 
     // check whether queue is empty
     if (m_queue.isEmpty())
     {
         emit allTasksCompleted();
-        return RunTaskResult::Invalid;
+        return TaskExecState::Invalid;
     }
 
     // checkout next task in queue
@@ -128,7 +128,7 @@ GtCoreProcessExecutor::startNextTask()
     {
         gtErrorId(GT_EXEC_ID) << tr("Cannot execute an invalid Task!");
         clearCurrentTask();
-        return RunTaskResult::Invalid;
+        return TaskExecState::Invalid;
     }
 
     // setup source
@@ -142,7 +142,7 @@ GtCoreProcessExecutor::startNextTask()
         {
             gtErrorId(GT_EXEC_ID) << tr("Source corrupted!");
             clearCurrentTask();
-            return RunTaskResult::Invalid;
+            return TaskExecState::Invalid;
         }
     }
 
@@ -157,7 +157,7 @@ GtCoreProcessExecutor::startNextTask()
             m_current->setState(GtProcessComponent::NONE);
             m_current = nullptr;
             emit queueChanged();
-            return RunTaskResult::Busy;
+            return TaskExecState::Busy;
         }
         if (result == GtProjectExecutionGuard::Result::InvalidProject)
         {
@@ -169,7 +169,7 @@ GtCoreProcessExecutor::startNextTask()
 
     execute();
 
-    return RunTaskResult::Started;
+    return TaskExecState::Started;
 }
 
 bool
--- a/src/core/gt_coreprocessexecutor.cpp
+++ b/src/core/gt_coreprocessexecutor.cpp
@@ -566,6 +566,7 @@ GtCoreProcessExecutor::onTaskRunnerFinished()
     if (!m_current)
     {
         gtFatalId(GT_EXEC_ID) << tr("Current task corrupted!");
+        pimpl->projectGuard.reset();
         taskRunner->deleteLater();
         return;
     }
--- a/src/core/gt_coreprocessexecutor.cpp
+++ b/src/core/gt_coreprocessexecutor.cpp
@@ -21,6 +21,7 @@
 #include "gt_objectmementodiff.h"
 #include "gt_coreapplication.h"
 #include "gt_executioncontext.h"
+#include "gt_projectexecutionguard.h"
 #include "gt_taskrunner.h"
 
 #include "gt_coreprocessexecutor.h"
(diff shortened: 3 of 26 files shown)


rainman110:
## Update: 
Done 

## Review: remaining work before merge

The implementation is close to merge-ready: the current head is mergeable, CI succeeds, and there are no unresolved review threads. I recommend addressing or explicitly resolving the following items before merging.

### Required

1. **Make save/close coordination atomic**

   `saveProject()` and `closeProject()` currently call `GtProjectExecutionGuard::isBusy(project)` before continuing. This is a check-then-act sequence: another thread can acquire the execution guard after the check and before save/close finishes.

   Please either:
   - acquire the same exclusive project guard for the complete save/close operation; or
   - document and enforce the invariant that save/close and execution startup can only occur serially on the same event-loop thread.

   For the intended service use, acquiring the same guard is preferable.

2. **Decide how projects without a path are serialized**

   `projectKey()` returns an empty key for pathless projects, and the executor then runs without a guard. The test `AllowsExecutionForProjectWithoutPath` currently confirms this behavior, although the PR describes mutation serialization per project.

   Prefer a process-local stable fallback identity, e.g. a guaranteed project UUID/runtime ID (and only as a last local fallback, object identity). If pathless projects intentionally remain unguarded, state this explicitly as a limitation and adjust the claims in the PR/documentation.

3. **Clarify `GtExecutionContext::isValid()`**

   `isValid()` currently means only `project() != nullptr`, while `GtRunnable` installs a context when any context field is set, including a path-only context.

   Please make the semantics consistent:
   - either define validity for the actual minimum usable context; or
   - rename the method to `hasProject()`.

   `GtRunnable` should ideally use one context method instead of independently inspecting all five fields.

4. **Guarantee guard cleanup on every completion/error path**

   In `onTaskRunnerFinished()`, the missing-`m_current` path releases the guard, but the invalid-`sender()` early return does not. Please centralize or scope cleanup so a future error path cannot leave a project permanently busy.

5. **Complete the minimum public API documentation**

   The full documentation restructuring belongs to #1518, but the public API introduced here should already document its contract.

   For `GtExecutionContext` / `GtExecutionContextScope`:
   - thread-local scope and synchronous lifetime;
   - borrowed ownership and prohibition on retaining context/project pointers;
   - no automatic propagation to child threads;
   - project access during calculator execution must not directly mutate the persistent project;
   - `projectPath()` can be available without `project()`;
   - `source` and `jobId` are not stable identifiers unless this is guaranteed;
   - precise meaning of `executionDataRoot` (currently the executor supplies it empty).

   For `GtProjectExecutionGuard`:
   - purpose, process-local scope, and thread safety;
   - project identity/key behavior;
   - meaning of each `Result`;
   - RAII release behavior;
   - that reacquiring first releases the previously held key.

   For `GtCoreProcessExecutor`:
   - meaning of every `RunTaskResult`;
   - `Busy` means rejected/removed, not automatically retried;
   - compatibility behavior of `runTask()`: both `Started` and `Queued` return `true`.

   For `GtRunnable`:
   - context lifetime and installation behavior;
   - context-first resolution of `projectPath()`;
   - fix the existing `Constrcutor` typo.

6. **Add the changelog entry**

   The PR checklist still has this open. A short entry should mention context-aware project resolution, compatibility for existing calculators, and the explicit context API for new execution code. The detailed module guidance will follow in #1518.

### Recommended

7. **Cover the remaining executor error/cleanup lines**

   Codecov reports three uncovered changed lines in `gt_coreprocessexecutor.cpp`. The overall patch coverage is already high, but cleanup and error paths are exactly where a leaked project guard would be most damaging. Please add focused regression coverage if those lines belong to such paths.

8. **Keep the baseline explicitly temporary**

   The complete integration into the regular module and Core documentation is tracked in #1518. For this PR, mark the compatibility baseline as a temporary implementation/transition note and link #1518; avoid expanding this PR into the full documentation restructuring.

### API-scope consideration

Before merging a new exported API, please also confirm that `executionDataRoot`, `source`, and `jobId` already have sufficiently stable semantics to become public contract. If the first slice only needs `project` and `projectPath`, reducing the initial API surface now will be easier than changing it after release.

rainman110:
## Update
Done 

## Follow-up review after `11f8282`

The previous merge-readiness findings are substantially addressed:

- save/close now acquire the same guard for the complete operation;
- pathless projects receive a UUID/address fallback identity;
- `GtExecutionContext::isValid()` now matches project-or-path contexts and `GtRunnable` uses it;
- guard cleanup is scope-based and the invalid-sender path has regression coverage;
- the public API surface was reduced to project + project path;
- API comments, changelog, and the temporary link to #1518 were added;
- current CI succeeds.

I do not see a remaining functional merge blocker. Two small documentation/API precision points are worth fixing before merge if possible:

1. **Clarify guard thread-safety wording**

   The `Result` comment currently says “Guards are process-local, thread-safe”. The shared registry is protected and coordinates different guard instances across threads, but an individual guard instance is not safe for concurrent calls: `m_key`, `isHeld()`, and `key()` are not synchronized.

   Suggested wording: “The process-local guard registry is thread-safe. A guard instance must be owned and used by one thread.”

2. **Make path-only `currentProject()` resolution explicit**

   The compatibility page correctly says that a path-only active context makes `currentProject()` return `nullptr`. The numbered resolution list can still be read as falling back to the selected session project when the active context contains no project pointer.

   Please state explicitly that an active context takes precedence even when `context->project()` is null; session fallback occurs only when no execution context is active. This is important because silently falling back to the GUI project would target the wrong project.

Codecov currently reports six uncovered changed lines (95.65% patch coverage), four in the executor and two in the guard. CI is green, so I do not consider the percentage itself a blocker. If these are additional cleanup or fallback branches, focused tests would still be valuable; otherwise the existing test set covers the important behavior introduced by the follow-up.

rainman110:
## Follow-up: active migration diagnostic

The compatibility behavior in this PR intentionally keeps existing calculator calls to `gtApp->currentProject()` working and does not globally deprecate the method, because it remains a valid API for operations targeting the project selected in the desktop GUI.

Active guidance for module developers is tracked separately in #1519. That follow-up will detect `currentProject()` access originating from an executing calculator/process component and emit a deduplicated Developer-Mode warning with the preferred migration alternatives, without changing execution behavior or warning for legitimate GUI use.

Together with the permanent documentation work in #1518, this provides a migration path without expanding the scope of #1514:

- #1514: compatibility and execution-context foundation;
- #1518: canonical module/Core documentation;
- #1519: developer-facing runtime migration diagnostic.

rainman110:
@mariusalexander Can I ask for a review from you? 😃

src/core/gt_coredatamodel.h
```diff
@@ -56,8 +56,11 @@ class GT_CORE_EXPORT GtCoreDatamodel : public QAbstractItemModel
...
-     * @brief Returns current project. If no current project exists NULL will
-     * be returned.
+     * @brief Returns the execution-context project or the session project.
+     *
+     * If a GtExecutionContext is active on the current thread, its project is
+     * returned. Otherwise the currently selected session project is returned;
+     * if neither exists, NULL is returned.
      * @return Current project pointer
      */
     GtProject* currentProject();
```
mariusalexander, line 66: Just noting my thoughts: did not know we have two functions for accessing the project, somehow I (and I assume most others) default to `gtApp->currentProject`.

mariusalexander:
@rainman110 I thought you are on vacation? It seems you are much more productive when you are not working :P I'll take a look next week, sorry, did not got to it this week.

rainman110:
> @rainman110 I thought you are on vacation? It seems you are much more productive when you are not working :P I'll take a look next week, sorry, did not got to it this week.

When I am not working, I have much more mental capacity to work 😁

src/core/gt_coreprocessexecutor.cpp
```diff
@@ -66,37 +88,50 @@ GtCoreProcessExecutor::setCoreExecutorFlags(Flags flags)
 
 bool
 GtCoreProcessExecutor::runTask(GtTask* task)
+{
+    const auto result = runTaskWithResult(task);
+    return result == RunTaskResult::Started || result == RunTaskResult::Queued;
+}
+
+GtCoreProcessExecutor::RunTaskResult
+GtCoreProcessExecutor::runTaskWithResult(GtTask* task)
```
mariusalexander: Not a particular fan of this naming scheme. Same for `executeNextTaskWithResult` below. If I read the code correctly the task is only triggered? Maybe we could use a function like 
- `triggerTaskExecution` and `triggerNextTaskExecution`
- or just `triggerTask` and `triggerNextTask`
- or `tryExecuteTask` and `tryExecuteNextTask`
- or `runTaskEx` and `executeNextTaskEx` (Ex for "extended")
I also don't know if we should deprecate the old method (returning a bool instead)?
Just some ideas, I'd like your input.

src/core/gt_coreprocessexecutor.cpp
```diff
@@ -446,12 +500,31 @@ GtCoreProcessExecutor::setupTaskRunner()
     // create new task runner
     auto* runner = new GtTaskRunner{m_current};
 
+    GtProject* project = qobject_cast<GtProject*>(m_source.data());
```
mariusalexander: This looks quite similar to `::projectForTask(...)`. Can these code blocks be merged?

src/core/gt_coreprocessexecutor.cpp
```diff
@@ -525,5 +602,7 @@ GtCoreProcessExecutor::onTaskRunnerFinished()
     // delete task runner
     taskRunner->deleteLater();
 
+    // Release before starting a queued task so it can acquire its own key.
+    pimpl->projectGuard.reset();
```
mariusalexander, line 584: isnt the guard reset by the finally clause at the start of this function?

src/core/gt_coreprocessexecutor.cpp
```diff
@@ -471,6 +544,10 @@ GtCoreProcessExecutor::onTaskRunnerFinished()
 {
     gtTrace() << __FUNCTION__;
 
+    const auto guardCleanup = qScopeGuard(
```
mariusalexander: You could also use "our" function:
```suggestion
    const auto guardCleanup = gt::finally(
```

src/core/gt_coreprocessexecutor.cpp
```diff
@@ -446,12 +500,31 @@ GtCoreProcessExecutor::setupTaskRunner()
...
+
     // create new runnable
-    pimpl->currentRunnable = new GtRunnable{pimpl->customProjectPath};
+    pimpl->currentRunnable = new GtRunnable{
+        pimpl->customProjectPath, std::move(executionContext)};
 
     // setup task runner
     if (!runner->setUp(pimpl->currentRunnable, m_source))
     {
+        pimpl->projectGuard.reset();
```
mariusalexander, line 505: why does the guard need to be reset here? (And where is it set?) Maybe we can move the "reset" call to the code section where the guard is set? (or assert that the guard is not set?)

src/core/gt_coreprocessexecutor.h
```diff
@@ -50,6 +51,18 @@ class GT_CORE_EXPORT GtCoreProcessExecutor : public QObject
...
+    enum class RunTaskResult
+    {
+        /// Execution was started immediately.
+        Started,
+        /// Task was accepted into this executor's queue.
+        Queued,
+        /// Task was rejected because the project or executor is busy.
+        Busy,
+        /// Task or execution setup was invalid.
+        Invalid
```
mariusalexander: Make "Invalid" default to `0`? IMO you could also think about dropping the `enum class` and just use `enum` here, since the "scope" of the enum is already limited to this class. That way one could also "implicitly" convert the enum to a bool? But that is just an idea.

```suggestion
        /// Task or execution setup was invalid.
        Invalid = 0,
        /// Execution was started immediately.
        Started,
        /// Task was accepted into this executor's queue.
        Queued,
        /// Task was rejected because the project or executor is busy.
        Busy
```

src/core/gt_executioncontext.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+#include "gt_executioncontext.h"
+
+#include <utility>
+
+#include "gt_project.h"
+
+namespace {
+
+thread_local GtExecutionContext const* currentExecutionContext = nullptr;
```
mariusalexander, line 15: Did not know about `thread_local`. This means every thread has an execution context?

src/core/gt_executioncontext.h
```diff
@@ -0,0 +1,82 @@
...
+    bool isValid() const noexcept;
+
+    /**
+     * @brief Returns the context installed on the current thread.
+     * @return Borrowed current context, or nullptr if no scope is active.
+     */
+    static GtExecutionContext const* current() noexcept;
+
+    /// Alias for current(), for callers that prefer the explicit name.
+    static GtExecutionContext const* currentContext() noexcept;
```
mariusalexander: I'd drop one IMO:
```suggestion
    static GtExecutionContext const* current() noexcept;
```

src/core/gt_executioncontext.h
```diff
@@ -0,0 +1,82 @@
...
+ */
+class GT_CORE_EXPORT GtExecutionContextScope
+{
+public:
+    explicit GtExecutionContextScope(GtExecutionContext const& context);
+    GtExecutionContextScope(GtExecutionContext&&) = delete;
+    ~GtExecutionContextScope();
+
+    GtExecutionContextScope(GtExecutionContextScope const&) = delete;
+    GtExecutionContextScope& operator=(GtExecutionContextScope const&) = delete;
```
mariusalexander, line 73: what about the move assignment? It is implictly deleted I think but best to mark it deleted too?
```suggestion
    GtExecutionContextScope& operator=(GtExecutionContextScope const&) = delete;
    GtExecutionContextScope& operator=(GtExecutionContextScope&&) = delete;
```

src/core/gt_executioncontext.h
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * @brief Installs an execution context for the lifetime of a scope.
+ *
+ * Scopes are thread-local, nestable, non-copyable, and restore the previous
+ * context in their destructor. The referenced context must outlive the scope.
+ * Neither the scope nor the context extends the lifetime of the borrowed
+ * project. Do not retain the context or project pointer for asynchronous work.
+ */
+class GT_CORE_EXPORT GtExecutionContextScope
```
mariusalexander, line 65: I am unsure if it makes a difference here, but maybe just export the constructor and destructor instead of the entire class?

src/core/gt_projectexecutionguard.h
```diff
@@ -0,0 +1,65 @@
...
+     *
+     * The process-local guard registry is thread-safe. A guard instance must
+     * be owned and used by one thread and is held for its lifetime. Acquiring
+     * a new project first releases a previously held key.
+     */
+    enum class Result
+    {
+        Acquired,
+        Busy,
+        InvalidProject
```
mariusalexander: I'd suggest making one enum have a fixed value:
```suggestion
        InvalidProject = 0
```
Also: since the enum is in a class, we may not need a "enum class"? But thats just my preference.

src/core/gt_projectexecutionguard.h
```diff
@@ -0,0 +1,65 @@
...
+     * @brief Returns the project identity used by the process-local guard.
+     *
+     * Existing project paths are canonicalized. Pathless projects use their
+     * stable object UUID and finally their in-process object address.
+     */
+    static QString projectKey(GtProject const* project);
+    static bool isBusy(GtProject const* project);
+
+private:
+    QString m_key;
```
mariusalexander, line 79: Maybe use a pimpl instead?

src/core/gt_projectexecutionguard.h
```diff
@@ -0,0 +1,65 @@
...
+     * @return Acquired on success, Busy if another guard holds the project,
+     *         or InvalidProject if no stable project identity is available.
+     */
+    Result tryAcquire(GtProject* project);
+
+    /// Releases the currently held project key, if any.
+    void release();
+
+    bool isHeld() const;
+    QString key() const;
```
mariusalexander: What does key refer to? Either add a description or use a more descriptive function name IMO

src/core/gt_projectexecutionguard.h
```diff
@@ -0,0 +1,65 @@
...
+     * @brief Tries to acquire exclusive access for @p project.
+     * @return Acquired on success, Busy if another guard holds the project,
+     *         or InvalidProject if no stable project identity is available.
+     */
+    Result tryAcquire(GtProject* project);
+
+    /// Releases the currently held project key, if any.
+    void release();
+
+    bool isHeld() const;
```
mariusalexander: Maybe a more standard naming scheme? Just a thought
```suggestion
    bool isLocked() const;
```

mariusalexander (changes requested):
@rainman110 Sorry for the holdup. I have not looked at the test to be honest, but I added my feedback to the other files. I have added lots of thoughts without an explicit request to incorporate a suggestion or not. If you decide not apply a suggestion, please write a short comment and resolve the corresponding thread :)

src/core/gt_coreapplication.cpp
```diff
@@ -948,14 +948,7 @@ GtCoreApplication::settings()
 GtProject*
 GtCoreApplication::currentProject()
 {
-    GtSession* w = session();
-
-    if (w)
-    {
-        return w->currentProject();
```
mariusalexander, line 955: One more thing: what about `GtSession::currentProject?` Do we need to adapt the context here too? If I see this correctly we have four ways of accessing a project: `GtCoreApplication::currentProject`, `GtCoreDataModel::currentProject`, `GtSession::currentProject` and `findParent<GtProject*>()`. The latter would be near impossible to incorporate, that just leaves `GtSession`

src/core/gt_executioncontext.h
rainman110: I agree!

src/core/gt_executioncontext.cpp
rainman110: `thread_local` means, that each thread owns its own global variable. 

In this case, each thread has a own execution context variable, which though could also be a nullptr. So each thread does not automatically have an execution context, but can have it.

src/core/gt_coreprocessexecutor.h
rainman110: I agree dropping `class` to make it less verbose.
What is your particular idea of making `Invalid` 0?

src/core/gt_executioncontext.h
rainman110: That would be an alternative. But I downvote it due to two reasons:
 1. It won't change anything
 2. It is pretty uncommon in the GTlab code base. The reason why the export macros are used here so often in the class declaration (and not in front of the members) is the fact that `MOC`ed slots from `QObject`-derived classes will not be exported otherwise, which will cause linker error.

src/core/gt_executioncontext.h
rainman110: I agree!

src/core/gt_executioncontext.h
mariusalexander: I see :) thanks for your input

src/core/gt_projectexecutionguard.h
rainman110: Pretty much same as before. Why do you think invalid is best suited for the first (and not last) position?

src/core/gt_projectexecutionguard.h
rainman110: Matter of taste. I could live with both. So I agree changing it.

src/core/gt_projectexecutionguard.h
rainman110: Agree

src/core/gt_coreprocessexecutor.cpp
rainman110: Agree!

src/core/gt_coreprocessexecutor.h
mariusalexander: I am unsure how enums are initialized when using  `return {}` or `RunTaskResult result = {}`.
I assume the "lowest" enum value is used, here this could be `Started`.
Therefore, I feel like `Invalid` would be a good default value for initializing `RunTaskResult` and like with integers `0` is a good default value.

Plus one could write:
```cpp
RunTaskResult result = ...
if (!result) // error handling
```
while it might not be desired to use it like this it feels intuitive and would work here.

One follow-up question: Why no `Finished` or `Valid` state? It seems this is only for triggering a task so maybe it is not needed

src/core/gt_coreapplication.cpp
rainman110: Good point. Currently, `GtCoreApplication::currentProject()` and `GtCoreDatamodel::currentProject()` are context-aware. `GtCoreApplication` ultimately delegates to the data model. `GtSession::currentProject()`, on the other hand, still exclusively returns the project selected in the GUI.

`findParent<GtProject*>() `is something different again: it performs a structural parent lookup and is not a global “current project” accessor.

So there are indeed two possible semantics:

“project of the current execution context”
“project currently selected in the session/GUI”

We should decide whether `GtSession::currentProject()` should also take the execution context into account. I would not change this implicitly as part of this PR, because that would alter the existing behavior of session-related code. At a minimum, however, this should be documented or addressed as follow-up work in #1518 (documentation if the execution context).

src/core/gt_coredatamodel.h
rainman110: You are right and this is not ideal. Now at least, one delegates to the other and we are not duplication code.

src/core/gt_coreprocessexecutor.cpp
rainman110: Yes, all "runTask" related methods are async. Since the old methods start with "runTask", i wanted to keep the naming scheme, but wanted to change the return value to something more useful.

Since this though will cause API changes, i needed to change the method name. So what do you suggest?

We could also use the `tryExecuteTask` scheme or maybe even better `startTask`, which reflects more what it is doing and maybe also deprecated the "normal" `runTask` methods.

src/core/gt_coreprocessexecutor.h
rainman110: You are right! So lets use 0 for `invalid`.  `Finished` or `Valid` does not make sense at this point when executing `runTask`, since it is an async operation. But i would make sense for other methods that aim to query the current task state.

So it is probably a good idea to add it LATER, when we need it.

After i thought about the pros and cons of enum vs enum class: This is not only about syntax but also about type safety. Enum can automatically be converted into `int`s  or even worse `bool`s, which make them inherently more unsafe. Hence, i'd like to stick with `enum class`.

src/core/gt_projectexecutionguard.h
rainman110: We well use `Invalid = 0` but stick with `enum class` due to the arguments written above.

src/core/gt_projectexecutionguard.h
rainman110: Why do you want one? Do you think that the internal representation will change over time?

src/core/gt_coreprocessexecutor.cpp
rainman110: Yes, the logic in `setupTaskRunner()` and `projectForTask(...)` should be consolidated. Both currently determine the project from:

 1. source, if it is itself a project,
 2. the parent of source,
 3. the parent of the task

src/core/gt_coreprocessexecutor.cpp
rainman110: @mariusalexander What do you think?

src/core/gt_coreprocessexecutor.cpp
rainman110: Yes, it will be reset, but too late. It needs to be reset before executing the next task! Thats why we call it manually here.

src/core/gt_coreprocessexecutor.cpp
rainman110: Same as in the other comment. We need to reset it before starting the next task

src/core/gt_coreprocessexecutor.cpp
mariusalexander: use `startTask`, deprecate `runTask`.

src/core/gt_projectexecutionguard.h
mariusalexander: Keep as is :)

rainman110:
@mariusalexander I addressed all of your review comments. Please check again

mariusalexander (approved):
LGTM

Issue #1515, opened 2026-07-27
Introduce a headless single-project runtime with asynchronous executable-operation support


rainman110:
Parent epic: #1526
Depends on: #1528, #1529
Normative contract: ADR 0001 (`docs/dev/architecture/decisions/0001-executable-operations-one-shot-worker.rst`)
Implementation input: PR #1520 (task-specific prototype; not normative)

## Goal

Implement `GtHeadlessProjectRuntime` as the **single-project execution-side application/runtime boundary** in GTlab Core and generalize the useful lifecycle/async mechanics from PR #1520 from tasks to arbitrary `GtExecutableOperation`s.

A runtime owns exactly one project lifecycle and runtime-owned operation execution states. It is broader than a task executor: short project reads/mutations remain explicit runtime/project APIs and are not automatically forced through asynchronous operation handles.

The runtime must be reusable by both the first one-shot batch worker (#1530) and a later resident single-project worker/session without changing operation semantics.

## Required execution boundary

```text
Project Worker / headless consumer
            |
            v
GtHeadlessProjectRuntime
    |-- exactly one project lifecycle
    |-- explicit project context / mutation coordination
    `-- asynchronous operation submission
             |
             v
      generic execution handle
             |
             v
GtExecutableOperation::execute()
```

Submission is asynchronous from the caller's perspective, while `GtExecutableOperation::execute(GtOperationExecutionContext&)` remains a synchronous execution-side contract.

Conceptually:

```cpp
auto handle = runtime.submitOperation(operation, data); // returns before heavy execute() blocks
```

The runtime/execution infrastructure owns scheduling, state transitions, cancellation, event connection and terminal capture. This is explicitly invalid:

```cpp
operation->execute(context);
return alreadyFinishedHandle;
```

Submission transfers/establishes execution-side ownership of the execution-local operation instance and optional detached invocation data for the lifetime of the invocation. The concrete C++ ownership spelling may use `std::unique_ptr` or an equivalent explicit ownership type, but borrowed originating/client operation state must not leak into the runtime.

## Runtime completion boundary

This runtime represents **execution-side completion only**:

```text
runtime success
    = execute() returned successfully
      + detached terminal result captured
```

It does **not** perform originating-side result transport/reconstruction or `applyResult()`. A later client-side `GtOperationExecutor` owns the broader prepare/transport/commit lifecycle and must not report client-visible success before `applyResult()` succeeds.

## Execution handle contract

[...]

Issue #1517, opened 2026-07-29
[CHANGE] - Make the implementation of GtProject private (pimpl-idiom)


mariusalexander:
### Summary

`GtProject` is an important data class that currently contains public members. These should be hidden using a pimpl (private-implementation) structure. This helps avoid future ABI changes. 

This is a follow-up of #1494

### Impact on the codebase

Avoid future ABI changes

### Current workarounds

-

PR #1520, opened 2026-07-31
Introduce headless single-project runtime

rainman110:
## Motivation

Introduce a headless, single-project Core runtime for asynchronous task services as described in #1515. This is stacked on the execution-context foundation from #1514 and is intended to support the later Python bindings and GTlab webservice migration.

## Scope

- add lifecycle operations for initialization, project open/save/close, task discovery and submission;
- reuse `GtExecutionContext`, `GtProjectExecutionGuard` and `runTaskWithResult()` from the Core executor;
- expose value-based task descriptors, stable handles, terminal status and structured operation results;
- support task-id, qualified group/task-id and UUID lookup without unwanted group-selection side effects;
- preserve pre-existing executor flags and enforce owner-thread and bounded-shutdown rules;
- add focused executor/runtime integration tests and developer documentation.

## Prototype migration

The implementation reuses the prototype's task descriptors, stable task handles, lifecycle polling and non-blocking execution model. It redesigns the boundary as a value-based Core API with an explicit project, execution context, project guard, structured operation/cancellation results and owner-thread lifecycle. Prototype-specific webservice and worker orchestration, as well as private implementation details, are intentionally not part of this API.

## Out of scope

Multi-project orchestration, worker-thread marshalling and the Python adapter itself remain follow-up work. The PR currently targets the not-yet-merged #1514 branch and can be retargeted to `master` after #1514 lands.

## Summary by Sourcery

Add a headless single-project Core runtime for safely executing asynchronous operations with explicit lifecycle and structured operation APIs.

New Features:
- Introduce a headless Core runtime that manages one project and supports initialization, project lifecycle operations, and asynchronous task submission.
- Expose stable operation handles with synchronized status, cancellation, waiting, event streams, detached results, and structured runtime outcomes.

Bug Fixes:
- Preserve configured core executor flags when they are updated and make them queryable.

Enhancements:
- Enforce owner-thread lifecycle rules, project execution guards, and bounded shutdown behavior for headless operations.

Documentation:
- Add developer documentation for using the headless project runtime.

Tests:
- Add integration coverage for operation execution, events, cancellation, shutdown, waiting, failures, project lifecycle, and scoped execution contexts.
- Add coverage confirming core executor flags are preserved and exposed.

### DIFF
--- a/src/core/gt_headlessprojectruntime.cpp
+++ b/src/core/gt_headlessprojectruntime.cpp
@@ -95,6 +95,11 @@ bool GtHeadlessTaskStatus::isDone() const
            state == State::Cancelled || state == State::Shutdown;
 }
 
+bool GtHeadlessTaskCancellationResult::succeeded() const
+{
+    return code == Code::Accepted;
+}
+
 GtHeadlessTaskHandle::GtHeadlessTaskHandle() = default;
 GtHeadlessTaskHandle::~GtHeadlessTaskHandle() = default;
 
@@ -137,16 +142,25 @@ GtHeadlessTaskStatus GtHeadlessTaskHandle::status() const
     if (m_state->runtimeClosed)
     {
         result.state = GtHeadlessTaskStatus::State::Shutdown;
+        result.result = GtHeadlessTaskStatus::Result::RuntimeShutdown;
         result.error = QStringLiteral("Runtime was shut down before task completion");
         return result;
     }
 
     if (!m_state->task)
     {
-        result.state = GtHeadlessTaskStatus::State::Failed;
+        const bool executorFinished = !m_state->executor ||
+                                      !m_state->executor->taskCurrentlyRunning();
+        result.state = executorFinished ? GtHeadlessTaskStatus::State::Failed :
+                                          GtHeadlessTaskStatus::State::Running;
+        result.result = executorFinished ? GtHeadlessTaskStatus::Result::TaskUnavailable :
+                                           GtHeadlessTaskStatus::Result::None;
         result.error = QStringLiteral("Task object is no longer available");
-        m_state->terminalStatus = result;
-        m_state->hasTerminalStatus = true;
+        if (executorFinished)
+        {
+            m_state->terminalStatus = result;
+            m_state->hasTerminalStatus = true;
+        }
         return result;
     }
 
@@ -167,12 +181,18 @@ GtHeadlessTaskStatus GtHeadlessTaskHandle::status() const
     }
     if (result.state == GtHeadlessTaskStatus::State::Failed)
     {
+        result.result = GtHeadlessTaskStatus::Result::ExecutionFailed;
         result.error = QStringLiteral("Task execution failed (%1)")
                            .arg(static_cast<int>(result.processState));
     }
     else if (result.state == GtHeadlessTaskStatus::State::Cancelled)
     {
+        result.result = GtHeadlessTaskStatus::Result::Cancelled;
         result.error = QStringLiteral("Task execution was cancelled");
+    }
+    else if (result.state == GtHeadlessTaskStatus::State::Finished)
+    {
+        result.result = GtHeadlessTaskStatus::Result::Succeeded;
     }
 
     if (result.isDone() && executorFinished)
@@ -183,20 +203,53 @@ GtHeadlessTaskStatus GtHeadlessTaskHandle::status() const
     return result;
 }
 
-bool GtHeadlessTaskHandle::cancel() const
+GtHeadlessTaskCancellationResult GtHeadlessTaskHandle::cancel() const
 {
-    if (!m_state || !isRuntimeOwnerThread() || m_state->runtimeClosed ||
-        m_state->hasTerminalStatus || !m_state->task || !m_state->executor)
+    if (!m_state)
     {
-        return false;
+        return {GtHeadlessTaskCancellationResult::Code::TaskUnavailable,
+                QStringLiteral("Task handle is invalid")};
+    }
+
+    if (!isRuntimeOwnerThread())
+    {
+        return {GtHeadlessTaskCancellationResult::Code::WrongThread,
+                QStringLiteral("Task handles must be used from the GTlab owner thread")};
+    }
+
+    if (m_state->runtimeClosed)
+    {
+        return {GtHeadlessTaskCancellationResult::Code::RuntimeShutdown,
+                QStringLiteral("Runtime was shut down")};
+    }
+
+    if (m_state->hasTerminalStatus || status().isDone())
+    {
+        return {GtHeadlessTaskCancellationResult::Code::AlreadyCompleted,
+                QStringLiteral("Task has already completed")};
+    }
+
+    if (!m_state->task)
+    {
+        return {GtHeadlessTaskCancellationResult::Code::TaskUnavailable,
+                QStringLiteral("Task object is no longer available")};
+    }
+
+    if (!m_state->executor)
+    {
+        return {GtHeadlessTaskCancellationResult::Code::ExecutorUnavailable,
+                QStringLiteral("Task executor is no longer available")};
     }
 
     const bool accepted = m_state->executor->terminateTask(m_state->task);
     if (accepted)
     {
         m_state->cancellationRequested = true;
+        return {GtHeadlessTaskCancellationResult::Code::Accepted, {}};
     }
-    return accepted;
+
+    return {GtHeadlessTaskCancellationResult::Code::ExecutorRejected,
+            QStringLiteral("Task executor rejected the cancellation request")};
 }
 
 GtHeadlessTaskStatus GtHeadlessTaskHandle::wait(int timeoutMs) const
@@ -417,16 +470,16 @@ GtHeadlessRuntimeResult GtHeadlessProjectRuntime::openProject(const QString& pro
                        QStringLiteral("Runtime must be used from the GTlab owner thread"));
     }
 
-    if (m_private->state != State::Initialized)
-    {
-        return failure(GtHeadlessRuntimeResult::Code::InvalidState,
-                       QStringLiteral("Runtime is not initialized"));
-    }
-
     if (m_private->project)
     {
         return failure(GtHeadlessRuntimeResult::Code::ProjectAlreadyLoaded,
                        QStringLiteral("A project is already loaded"));
+    }
+
+    if (m_private->state != State::Initialized)
+    {
+        return failure(GtHeadlessRuntimeResult::Code::InvalidState,
+                       QStringLiteral("Runtime is not initialized"));
     }
 
     if (gtDataModel->currentProject())
--- a/src/core/gt_headlessprojectruntime.cpp
+++ b/src/core/gt_headlessprojectruntime.cpp
@@ -676,7 +676,6 @@ GtHeadlessRuntimeResult GtHeadlessProjectRuntime::closeProject()
                            QStringLiteral("Project execution started during close"));
         }
 
-        m_private->state = State::CloseFailed;
         return failure(GtHeadlessRuntimeResult::Code::CloseFailed,
                        QStringLiteral("Project could not be closed"));
     }
--- a/src/core/gt_headlessprojectruntime.cpp
+++ b/src/core/gt_headlessprojectruntime.cpp
@@ -84,6 +84,7 @@ struct GtHeadlessTaskHandle::State
     QPointer<GtTask> task;
     QPointer<GtCoreProcessExecutor> executor;
     bool runtimeClosed{false};
+    bool cancellationRequested{false};
     mutable bool hasTerminalStatus{false};
     mutable GtHeadlessTaskStatus terminalStatus;
 };
@@ -142,13 +143,28 @@ GtHeadlessTaskStatus GtHeadlessTaskHandle::status() const
 
     if (!m_state->task)
     {
-        result.state = GtHeadlessTaskStatus::State::Invalid;
+        result.state = GtHeadlessTaskStatus::State::Failed;
         result.error = QStringLiteral("Task object is no longer available");
+        m_state->terminalStatus = result;
+        m_state->hasTerminalStatus = true;
         return result;
     }
 
     result.processState = m_state->task->currentState();
     result.state = taskState(result.processState);
+    const bool executorFinished = !m_state->executor ||
+                                  (!m_state->executor->taskCurrentlyRunning() &&
+                                   !m_state->executor->taskQueued(m_state->task));
+    if (result.isDone() && !executorFinished)
+    {
+        result.state = GtHeadlessTaskStatus::State::Running;
+    }
+    if (m_state->cancellationRequested && executorFinished &&
+        (result.state == GtHeadlessTaskStatus::State::Failed ||
+         result.state == GtHeadlessTaskStatus::State::Running))
+    {
+        result.state = GtHeadlessTaskStatus::State::Cancelled;
+    }
     if (result.state == GtHeadlessTaskStatus::State::Failed)
     {
         result.error = QStringLiteral("Task execution failed (%1)")
@@ -159,7 +175,7 @@ GtHeadlessTaskStatus GtHeadlessTaskHandle::status() const
         result.error = QStringLiteral("Task execution was cancelled");
     }
 
-    if (result.isDone())
+    if (result.isDone() && executorFinished)
     {
         m_state->terminalStatus = result;
         m_state->hasTerminalStatus = true;
@@ -175,7 +191,12 @@ bool GtHeadlessTaskHandle::cancel() const
         return false;
     }
 
-    return m_state->executor->terminateTask(m_state->task);
+    const bool accepted = m_state->executor->terminateTask(m_state->task);
+    if (accepted)
+    {
+        m_state->cancellationRequested = true;
+    }
+    return accepted;
 }
 
 GtHeadlessTaskStatus GtHeadlessTaskHandle::wait(int timeoutMs) const
@@ -191,7 +212,7 @@ GtHeadlessTaskStatus GtHeadlessTaskHandle::wait(int timeoutMs) const
                (!m_state->executor->taskCurrentlyRunning() &&
                 !m_state->executor->taskQueued(m_state->task));
     };
-    if (current.isDone() && executorFinished())
+    if (current.isDone() && (!m_state->task || executorFinished()))
     {
         return current;
     }
@@ -371,6 +392,21 @@ void GtHeadlessProjectRuntime::restoreExecutorFlags()
     m_private->executorCompletionConnection = {};
     m_private->configuredExecutor.clear();
     m_private->executorFlagsOverridden = false;
+}
+
+void GtHeadlessProjectRuntime::cleanupCompletedTasks()
+{
+    if (!isRuntimeOwnerThread())
+    {
+        return;
+    }
+
+    auto isCompleted = [](const auto& task) {
+        return GtHeadlessTaskHandle(task).status().isDone();
+    };
+    m_private->tasks.erase(
+        std::remove_if(m_private->tasks.begin(), m_private->tasks.end(), isCompleted),
+        m_private->tasks.end());
 }
 
 GtHeadlessRuntimeResult GtHeadlessProjectRuntime::openProject(const QString& projectPath)
@@ -791,7 +827,10 @@ GtHeadlessTaskHandle GtHeadlessProjectRuntime::submitTask(
             executor,
             &GtCoreProcessExecutor::allTasksCompleted,
             this,
-            [this]() { restoreExecutorFlags(); });
+            [this]() {
+                restoreExecutorFlags();
+                cleanupCompletedTasks();
+            });
     }
 
     auto flags = executor->coreExecutorFlags();
--- a/src/core/gt_headlessprojectruntime.cpp
+++ b/src/core/gt_headlessprojectruntime.cpp
@@ -12,6 +12,7 @@
 #include <QFileInfo>
 #include <QDir>
 #include <QPointer>
+#include <QTimer>
 #include <QThread>
 #include <QUuid>
 
@@ -30,8 +31,12 @@
 #include "gt_taskgroup.h"
 #include "provider/gt_projectprovider.h"
 
+bool isRuntimeOwnerThread();
+
 namespace
 {
+constexpr int shutdownTimeoutMs = 5000;
+
 GtHeadlessTaskStatus::State taskState(GtProcessComponent::STATE state)
 {
     using ProcessState = GtProcessComponent::STATE;
@@ -79,6 +84,8 @@ struct GtHeadlessTaskHandle::State
     QPointer<GtTask> task;
     QPointer<GtCoreProcessExecutor> executor;
     bool runtimeClosed{false};
+    mutable bool hasTerminalStatus{false};
+    mutable GtHeadlessTaskStatus terminalStatus;
 };
 
 bool GtHeadlessTaskStatus::isDone() const
@@ -114,9 +121,22 @@ GtHeadlessTaskStatus GtHeadlessTaskHandle::status() const
     }
 
     result.id = m_state->id;
+    if (!isRuntimeOwnerThread())
+    {
+        result.state = GtHeadlessTaskStatus::State::Invalid;
+        result.error = QStringLiteral("Task handles must be used from the GTlab owner thread");
+        return result;
+    }
+
+    if (m_state->hasTerminalStatus)
+    {
+        return m_state->terminalStatus;
+    }
+
     if (m_state->runtimeClosed)
     {
         result.state = GtHeadlessTaskStatus::State::Shutdown;
+        result.error = QStringLiteral("Runtime was shut down before task completion");
         return result;
     }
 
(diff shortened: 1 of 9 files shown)


rainman110:
## Remaining work before review

- Rebase onto #1514 and use its `GtExecutionContext`, project guard, and structured executor result instead of leaving a future integration comment.
- Clarify executor ownership; do not overwrite/reset flags on the globally selected executor without preserving existing state.
- Fix `closeProject()` failure handling: handles must not become `Shutdown` before close/delete succeeds, and partial close failures must leave a consistent lifecycle state.
- Make shutdown bounded; the destructor must not wait forever when cancellation is ignored.
- Store a stable terminal status/result in the handle, including structured execution errors and a defined retention/cleanup policy.
- Enforce the documented owner-thread rule for handle `status()`, `cancel()`, and `wait()`; avoid a busy-loop implementation of `wait()`.
- Make `listTasks()` side-effect free, or correctly restore both the previous task-group ID and type.
- Add UUID-based task submission and document how ambiguous/unqualified task names are resolved.
- Add real integration tests for open/list/submit/wait/cancel/save/close, legacy `currentProject()`, context/guard behavior, shutdown, and failure cleanup. The current tests cover only the API skeleton.
- Replace the generated-only PR description with a migration summary linked to #1515; the current Sourcery reference to #1517 is incorrect.

rainman110:
## Follow-up findings

The runtime is now substantially closer to the intended design. Remaining items:

- **Do not freeze terminal status before executor completion.** A task may report `FINISHED` before result merging completes; a later merge failure would currently remain cached as `Finished`. Persist the final snapshot only after the executor has fully completed the task.
- **Make a lost task object terminal.** If the submitted `GtTask` disappears, `status()` returns non-terminal `Invalid`; `wait(-1)` can then block forever without a future signal. Store a terminal failure/shutdown result instead.
- **Verify UUID lookup across inactive task groups.** `project->findChildren<GtTask*>()` likely only sees the currently loaded group. Add a test submitting the UUID of a task in a non-selected custom group and adapt lookup if needed.
- **Define completed-handle retention.** `m_private->tasks` currently retains every submitted task state until runtime destruction. Remove completed entries after preserving their shared terminal snapshot, or document and bound job-history retention.
- **Document the global-executor assumption.** Flag preservation is improved, but restoring a saved flag snapshot can still overwrite changes from another consumer. State that the first runtime assumes one runtime/executor consumer per worker process, or use a runtime-owned executor instance.
- Add tests for merge failure, lost task object, custom-group UUID lookup, successful close, `CloseFailed` retry, real cancellation, shutdown timeout, executor-flag restoration, and completed-state cleanup.
- Minor: change “dtlab webservice” to “GTlab webservice” in the PR description.

rainman110:
Follow-up after reviewing the latest changes: the previously reported core issues appear to be addressed. Before marking the PR ready for review, I would still suggest:

- Run/verify the full Core test suite for the current head; there is currently no GitHub Actions run associated with it.
- Rebase/retarget onto the final `master` state after #1514 has been merged, then rerun the tests.
- Add coverage for the remaining lifecycle/error paths where practical:
  - failure during executor post-processing/result merge,
  - `CloseFailed` followed by a successful retry,
  - shutdown timeout resulting in terminal `Shutdown`,
  - restoration of the original executor flags,
  - cleanup of completed runtime state while an external handle remains usable.
- Clarify in the `status()` documentation that `Finished` means the complete executor processing, including post-processing/result merge, has completed—not merely the calculator execution itself.

I do not see a new conceptual blocker in the current delta.

rainman110:
## Remaining work against #1515

The main runtime architecture and happy path are implemented, but the following acceptance criteria from #1515 still need to be addressed before the issue can be considered complete:

- **Return a structured cancellation result.** `GtHeadlessTaskHandle::cancel()` currently returns only `bool`, although #1515 explicitly requires replacing boolean-only cancellation outcomes. Callers should be able to distinguish at least accepted, already completed, runtime shut down, unavailable task/executor, wrong thread, and executor rejection.
- **Expose and test the structured final execution result.** The current status contains a state and error string, but the final executor outcome—especially post-processing/result-merge failure—is not demonstrably propagated. Add a test for this failure path and ensure it becomes a structured terminal failure.
- **Add the explicit legacy compatibility test required by #1515.** Execute an unchanged calculator that resolves its project via `gtApp->currentProject()`. The existing execution-context test only verifies `GtExecutionContext::current()->project()` and therefore does not cover this acceptance criterion.
- **Add the prototype migration note to the PR description.** Briefly document which prototype concepts were reused, which parts were redesigned and why, and which prototype behavior was deliberately not migrated. Do not disclose private source.
- **Complete the missing lifecycle/error coverage**, in particular:
  - initialize and clean shutdown without a project;
  - reject a second project;
  - `CloseFailed` followed by a successful retry;
  - shutdown timeout resulting in terminal `Shutdown`;
  - executor-flag restoration;
  - guard release after failure and cancellation.
- **Rebase/retarget onto `master` after #1514 is merged and verify the complete test suite.**

If some of the rare lifecycle tests are intentionally deferred, please create and link explicit follow-up issues and document the reduced scope in both #1515 and this PR.

src/core/gt_headlessprojectruntime.h
```diff
@@ -0,0 +1,211 @@
...
+        Cancelled,
+        TaskUnavailable,
+        ExecutionFailed,
+        RuntimeShutdown
+    };
+
+    QString id;
+    State state{State::Invalid};
+    GtProcessComponent::STATE processState{GtProcessComponent::NONE};
+    int progress{-1};
```
rainman110, line 55: `progress` is currently part of the public status contract but is never populated, so it always remains `-1`. Please propagate `GtProcessComponent::currentProgress()` (using `-1` only when progress is genuinely unavailable) and add a test with a calculator/task that reports an intermediate progress value. This is required by #1515's “Progress is exposed where available” acceptance criterion.

src/core/gt_headlessprojectruntime.cpp
```diff
@@ -0,0 +1,933 @@
...
+    return code == Code::Success;
+}
+
+GtHeadlessProjectRuntime::GtHeadlessProjectRuntime(QObject* parent) :
+    QObject(parent),
+    m_private(new Private)
+{
+}
+
+GtHeadlessProjectRuntime::~GtHeadlessProjectRuntime()
```
rainman110, line 354: Please make the destruction contract explicit and safe. When destruction occurs off the owner thread while the GTlab thread is no longer running or `QCoreApplication::closingDown()` is true, this path returns without `shutdown()`: the project may remain registered, executor flags may remain overridden, and handles receive no terminal shutdown snapshot. For this first worker scope, it is acceptable to require destruction on the owner thread, but then document that precondition on the public destructor/class contract and enforce it (for example with a diagnostic assertion). Otherwise implement a deterministic fallback that restores all runtime-owned state without invoking owner-thread-only APIs.

tests/unittests/core/test_gt_headlessprojectruntime.cpp
```diff
@@ -0,0 +1,515 @@
...
+    });
+    foreignThread.join();
+    EXPECT_EQ(foreignThreadCancellation.code,
+              GtHeadlessTaskCancellationResult::Code::WrongThread);
+
+    const auto cancellation = handle.cancel();
+    ASSERT_TRUE(cancellation.succeeded());
+    EXPECT_EQ(cancellation.code,
+              GtHeadlessTaskCancellationResult::Code::Accepted);
+    InterruptibleCalculator::release.store(true);
```
rainman110: This test releases the calculator through a separate test flag after requesting cancellation. It therefore proves that the request is accepted and later mapped to `Cancelled`, but not that `cancel()` actually stops a cooperative calculation. Please make the test calculator react to the same interruption/termination mechanism used by `terminateTask()` and verify that the handle request alone causes it to finish. Keep the existing guard-release and `AlreadyCompleted` assertions.

tests/unittests/core/test_gt_headlessprojectruntime.cpp
```diff
@@ -0,0 +1,515 @@
...
+{
+    ASSERT_NE(openProject(), nullptr);
+
+    EXPECT_TRUE(m_runtime->closeProject());
+    EXPECT_EQ(m_runtime->state(), GtHeadlessProjectRuntime::State::Closed);
+    EXPECT_TRUE(m_runtime->projectPath().isEmpty());
+    EXPECT_EQ(gtDataModel->currentProject(), nullptr);
+}
+
+TEST_F(TestGtHeadlessProjectRuntime, CompletedHandleRemainsUsableAfterRuntimeCleanup)
```
rainman110, line 855: Please add focused coverage for the remaining contract-critical paths before merge: (1) calculator/executor failure, including post-processing or result-merge failure, must produce a structured terminal `Failed` result and release `GtProjectExecutionGuard`; (2) `CloseFailed` must preserve a retryable runtime/project state and a subsequent retry must reach `Closed`; (3) shutdown timeout must leave copied handles terminal as `Shutdown`; and (4) the original executor flags must be restored after close/destruction. A save-success assertion should also be added so every public lifecycle operation has a positive-path test.

src/core/gt_headlessprojectruntime.h
```diff
@@ -0,0 +1,211 @@
...
+        ProjectLoaded,
+        CloseFailed,
+        Closed
+    };
+    Q_ENUM(State)
+
+    explicit GtHeadlessProjectRuntime(QObject* parent = nullptr);
+    ~GtHeadlessProjectRuntime() override;
+
+    GtHeadlessRuntimeResult initialize();
```
rainman110, line 226: Please add Doxygen documentation for every exported operation (`initialize`, open/save/close, task listing/submission, status/cancellation/wait) and for the public result/status fields. In particular, specify that `Closed` is terminal, `CloseFailed` is retryable, `wait()` timeout returns the latest non-terminal snapshot rather than a timeout result, and `Finished` means complete executor processing including result merge—not only calculator completion. The RST guide is useful but should not be the sole contract for this exported Core API.

rainman110 (commented):
The single-project architecture and the main happy path are in good shape. The points below should still be addressed before merge because they are part of #1515's public runtime contract or cover newly introduced lifecycle/recovery behavior. Multi-project orchestration, broker/cluster integration, and a public transport event API should remain out of this PR. After #1514 is merged, please rebase/retarget this PR onto `master` and rerun the full Core test matrix before marking it ready.

src/core/gt_headlessprojectruntime.h
rainman110: Implemented in `19947255`. `GtHeadlessTaskStatus::progress` now aggregates the task and its process components, returning `-1` only when no component exposes progress. `ReportsTaskProgress` verifies an intermediate value of 42% while the task is running.

src/core/gt_headlessprojectruntime.cpp
rainman110: Addressed in `19947255`. The public class contract now documents owner-thread destruction. Destruction is marshalled while the owner thread is available; otherwise a diagnostic assertion is emitted when runtime-owned state exists, preventing silent loss of project/flag/handle cleanup.

tests/unittests/core/test_gt_headlessprojectruntime.cpp
rainman110: Addressed in `19947255`. `InterruptibleCalculator` now exits solely when the interruption requested by `terminateTask()` reaches its parent task; the test no longer releases it through an independent test flag. The test still verifies guard release and `AlreadyCompleted`.

tests/unittests/core/test_gt_headlessprojectruntime.cpp
rainman110: Addressed in `19947255`. Added focused coverage for structured execution failure and guard release, `CloseFailed` retry, shutdown timeout with terminal `Shutdown`, executor-flag restoration, successful save, progress, and cancellation. Final failure/shutdown results are now structured and retained after executor completion.

src/core/gt_headlessprojectruntime.h
rainman110: Addressed in `19947255`. Added Doxygen documentation for exported status/result fields and runtime/handle operations, including owner-thread requirements, retryable `CloseFailed`, terminal `Closed`, timeout semantics, progress, and the post-processing/result-merge boundary for `Finished`.

rainman110 (commented):
There are still three contract-critical gaps before merge. The comments below separate the required runtime-state correction from the two missing test proofs. These requests stay within #1515's single-project runtime scope; no worker, broker, or multi-project functionality is required.

src/core/gt_headlessprojectruntime.cpp
```diff
@@ -0,0 +1,963 @@
...
+
+    if (!gtDataModel->closeProject(project))
+    {
+        if (GtProjectExecutionGuard::isBusy(project))
+        {
+            return failure(GtHeadlessRuntimeResult::Code::ProjectBusy,
+                           QStringLiteral("Project execution started during close"));
+        }
+
+        m_private->state = State::CloseFailed;
```
rainman110: `CloseFailed` currently conflates two different phases. If `closeProject(project)` fails here, the project may still be open, but the next `closeProject()` call enters the `State::CloseFailed` branch above and calls `deleteProject(project)` directly without retrying the close operation. Please keep the runtime in `ProjectLoaded` when the close operation itself fails (while returning `Code::CloseFailed` for that call), so a retry invokes `gtDataModel->closeProject()` again. Enter the persistent `State::CloseFailed` state only after closing succeeded and `deleteProject()` failed. Please add separate tests for both phases: a close failure must remain retryable through another close attempt, while a removal failure must retry removal and reach `Closed`.

tests/unittests/core/test_gt_headlessprojectruntime.cpp
```diff
@@ -0,0 +1,732 @@
...
+    });
+    foreignThread.join();
+    EXPECT_EQ(foreignThreadCancellation.code,
+              GtHeadlessTaskCancellationResult::Code::WrongThread);
+
+    const auto cancellation = handle.cancel();
+    ASSERT_TRUE(cancellation.succeeded());
+    EXPECT_EQ(cancellation.code,
+              GtHeadlessTaskCancellationResult::Code::Accepted);
+    InterruptibleCalculator::release.store(true);
```
rainman110: This line still invalidates the cancellation proof: the calculator exits when either the release flag or `isInterruptionRequested()` becomes true, so the test may pass even if `terminateTask()` never propagates interruption. Please remove this release from the success path and let `handle.cancel()` be the sole cause of termination. Keep a release mechanism only as a failure/cleanup fallback that runs after the assertion path (for example via an RAII guard), so a regression cannot hang the suite. The test should continue to verify terminal `Cancelled`, guard release, and `AlreadyCompleted`.

tests/unittests/core/test_gt_headlessprojectruntime.cpp
```diff
@@ -0,0 +1,732 @@
...
+
+    delete task;
+
+    const auto status = handle.wait(100);
+    EXPECT_EQ(status.state, GtHeadlessTaskStatus::State::Failed);
+    EXPECT_EQ(status.result, GtHeadlessTaskStatus::Result::TaskUnavailable);
+    EXPECT_TRUE(status.isDone());
+}
+
+TEST_F(TestGtHeadlessProjectRuntime, FailedTaskReleasesProjectGuard)
```
rainman110, line 533: This test covers only a calculator whose `run()` fails; it does not verify the documented executor boundary after successful calculation. Please add a distinct test in which calculator execution succeeds but post-processing/result merging fails. Assert that the handle never snapshots `Finished` prematurely, ends as `Failed` with `Result::ExecutionFailed`, releases `GtProjectExecutionGuard`, and retains the same terminal failure in a copied handle. This is needed because #1515's async result contract depends on completion of the full executor operation, not only calculator completion.

src/core/gt_headlessprojectruntime.cpp
rainman110: Addressed in `14743155`. A failure of `gtDataModel->closeProject()` now leaves the runtime in `ProjectLoaded`, so the next call retries the close phase. `CloseFailed` is entered only after close succeeded and removal failed. Separate `CloseFailureKeepsProjectLoadedForRetry` and `RemovalFailureRetriesRemoval` tests cover both recovery paths.

tests/unittests/core/test_gt_headlessprojectruntime.cpp
rainman110: Addressed in `14743155`. The success path no longer sets the release flag; cancellation and task interruption are the sole cause of calculator termination. A `QScopeGuard` provides release only during test cleanup after assertions. The test still verifies `Cancelled`, guard release, and `AlreadyCompleted`.

tests/unittests/core/test_gt_headlessprojectruntime.cpp
rainman110: Addressed in `14743155`. `MergeFailureBecomesTerminalExecutionFailure` runs a calculator successfully, removes its merge target before executor post-processing, and verifies that the handle never exposes `Finished`, ends at `Failed`/`ExecutionFailed`, releases the project guard, and retains the same result through a copied handle.

rainman110:
One architectural clarification should be added before merge, without expanding the implementation scope of this PR.

`GtHeadlessProjectRuntime` is intended to be the **single-project application boundary** of a future Project Worker. Asynchronous task execution is its first implemented capability, but it is not the complete long-term purpose of the runtime.

Future short-lived project queries and mutations—including Memento-based read/apply operations—should use the same loaded project and owner-thread execution context, but they will not necessarily be represented as asynchronous task handles.

Please clarify this in the class-level documentation of `gt_headlessprojectruntime.h` and, consistently, in `headless_project_runtime.rst`. The documentation should avoid establishing the runtime as a task-only abstraction.

No query/command API, Memento API, project revisions, project events, or IPC contracts need to be implemented in this PR; those should remain follow-up work.

rainman110:
## Architecture update: adapt this draft from async tasks to async executable operations

The target architecture of #1515 has changed since this draft was created. Please treat the current PR as a useful implementation prototype, **not** as the final public API shape.

The important change is that `GtHeadlessProjectRuntime` remains the single-project application/runtime boundary, but its asynchronous execution capability is now **generic** and operates on `GtExecutableOperation`, as defined by #1526 and the operation foundation in #1528. `GtTask` becomes one integration case through `ProcessTaskOperation` (#1531).

Please update this PR accordingly rather than continuing to harden a task-specific runtime API.

### 1. Preserve the runtime/application-boundary work

The following concepts in this PR are still valuable and should be kept/refactored rather than discarded:

- `GtHeadlessProjectRuntime` as the owner of exactly one loaded project;
- initialization/open/save/close lifecycle;
- explicit `GtExecutionContext` use for the loaded project;
- project mutation serialization/guard integration;
- owner-thread / Qt event-loop rules where required by current Core internals;
- immediate asynchronous submission;
- runtime-owned execution state rather than a process-global handle registry;
- value-based handles with no public `GtTask*`, `GtProject*` or other Qt/GTlab object pointers;
- status polling;
- cancellation;
- optional blocking `wait()` with bounded shutdown behavior;
- stable terminal result/error snapshots;
- deterministic behavior on project close/runtime destruction;
- structured errors/results;
- the thin Python-adapter philosophy.

The runtime documentation should continue to describe `GtHeadlessProjectRuntime` as the **single-project application boundary**, not as a task executor.

### 2. Generalize the public async API from tasks to executable operations

The architectural async boundary is now:

```text
caller
   |
   | submit operation
   v
GtHeadlessProjectRuntime
   |
   +-- runtime-owned execution state
   +-- generic execution handle
   +-- cancellation
   +-- execution event stream
   |
   v
GtExecutableOperation::execute(GtOperationExecutionContext&)
```

Submission is asynchronous from the caller's perspective and must return a handle immediately.

`GtExecutableOperation::execute(...)` itself remains **synchronous on the execution side**. Do not make operation implementations spawn their own threads merely to satisfy the async API. The runtime provides the asynchronous lifecycle around the synchronous `execute()` call.

The exact final public names/signatures should follow #1528 / the operation ADR, but conceptually the runtime should evolve toward something like:

```cpp
auto handle = runtime.submitOperation(operation, data);

handle.status();
handle.cancel();
handle.wait();
handle.result();
```

where `data` is the optional detached per-invocation `GtObject` defined by the operation architecture and is separate from project state.

### 3. Generalize the task-specific handle/status types

The current public types such as:

- `GtHeadlessTaskHandle`
- `GtHeadlessTaskStatus`

should not remain the generic runtime execution contract.

Refactor the reusable state/lifetime implementation behind them into a generic execution/operation handle and generic execution status (exact names to be finalized with #1528).

Generic state must not fundamentally depend on storing a `GtTask*` or `GtCoreProcessExecutor*` as the identity of an execution.

States such as queued/running/finished/failed/cancelled/shutdown may remain useful, but task/process-specific fields must not define the generic API.

### 4. Remove the hard dependency `runtime -> GtCoreProcessExecutor`

This is the most important implementation change.

The generic runtime must **not** require every `GtExecutableOperation` to execute through `GtCoreProcessExecutor`.

The desired dependency is:

```text
GtHeadlessProjectRuntime
        |
        v
GtExecutableOperation
        |
        +-- arbitrary operation implementation
        |
        +-- ProcessTaskOperation
                |
                v
         GtCoreProcessExecutor
                |
                v
             GtTask
```

The existing task/executor integration code in this PR is still useful reference material, but it belongs behind `ProcessTaskOperation` (#1531), not in the generic runtime execution path.

Do **not** generalize or replace `GtCoreProcessExecutor` itself as part of this PR.

### 5. Move task lookup/discovery out of the generic execution contract

The following parts of the current draft are task-domain concerns, not generic runtime execution concerns:

- `TaskSpec` parsing;
- `TaskId` / `GroupName/TaskId` lookup;
- UUID task lookup;
- `listTasks()` as a prerequisite for generic execution;
- task-group selection/restoration logic.

They should not drive the new runtime API.

If some task discovery helper is worth preserving for compatibility with the existing Python backend, keep it clearly separated as a task-specific convenience API, or defer it to the `ProcessTaskOperation` integration work. Do not couple `submitOperation()` to task discovery/addressing.

### 6. Integrate the generic operation execution context

Each submitted execution should create/use the generic `GtOperationExecutionContext` from #1528.

That context is invocation-local and is distinct from `GtExecutionContext`:

```text
GtExecutionContext
    = loaded/provisioned project execution scope

GtOperationExecutionContext
    = one operation invocation
      - detached operation data
      - execution event sink
      - cancellation
      - execution identity
```

For project-bound operations, establish the normal `GtExecutionContext` / project guard around execution. Do not add the project pointer to `GtOperationExecutionContext` merely to make project access possible.

### 7. Replace task progress coupling with the generic event stream

The runtime execution state should integrate with the generic execution event stream from #1529.

Operations publish through the event sink in `GtOperationExecutionContext`; local observation uses Qt signals/slots. Task-specific progress/state can later be translated into generic operation events by `ProcessTaskOperation`.

Do not make generic runtime progress depend on `GtTask::currentProgress()` or `GtProcessComponent` state.

### 8. Generalize cancellation

`handle.cancel()` remains part of the async contract, but cancellation should target the generic execution/cancellation state exposed through `GtOperationExecutionContext`.

A concrete operation decides how cancellation reaches its underlying computation. For `ProcessTaskOperation`, that adapter can translate generic cancellation into the existing task/executor termination mechanism.

The generic runtime must not assume task cancellation semantics.

### 9. Generalize results

Successful runtime completion means:

```text
GtExecutableOperation::execute()
        finished
           +
 detached GtObject result available
```

The result is a detached serializable `GtObject` according to #1526/#1528.

This runtime must **not** call `applyResult()` on the originating authoritative project. That is a later client-side `GtOperationExecutor` responsibility.

This distinction is important:

```text
runtime execution completion
    = execute() completed + result available

client invocation completion (later)
    = result transported/deserialized + applyResult() completed
```

Do not collapse these two lifecycle boundaries into one simply because both may expose handles.

### 10. Keep project lifecycle and short project APIs broader than async operations

`GtHeadlessProjectRuntime` remains broader than its async computation facility.

Project lifecycle and future short read/apply APIs do not all have to become asynchronous operation handles. Computational `GtExecutableOperation` submission is async; lightweight project/application operations may remain synchronous where appropriate.

### 11. Suggested refactoring order for this PR

Please refactor incrementally so the useful current implementation can be retained:

1. keep `GtHeadlessProjectRuntime` project lifecycle code and tests;
2. extract/generalize the current task handle/state implementation into generic execution state/handle types;
3. remove `GtTask*` / `GtCoreProcessExecutor*` as defining public/generic execution state;
4. introduce generic async `submitOperation(...)` based on the APIs from #1528;
5. create `GtOperationExecutionContext` per invocation and establish `GtExecutionContext` separately when project state is required;
6. wire generic cancellation and the execution event stream (#1529);
7. return the detached operation result through the generic handle;
8. move/defer task lookup, task progress mapping and direct `GtCoreProcessExecutor` execution to `ProcessTaskOperation` (#1531);
9. update documentation/tests from "async task runtime" to "single-project runtime with async executable-operation support".

It is fine for this PR to remain draft/stacked on the operation-foundation work while the exact operation types are introduced. Please avoid creating temporary public task-specific APIs merely to make the branch compile if they would immediately have to be removed again.

### 12. Scope that still stays out of this PR

Do not add:

- client-side `GtOperationExecutor`;
- worker CLI/stdio framing;
- remote/control-plane transport;
- GUI interactors;
- resident-session synchronization/protocol;
- broker/queue/cluster concepts;
- a new operation serializer/factory/registry;
- a replacement for the existing task executor hierarchy.

The one-shot batch worker (#1530) should later consume this runtime rather than owning an independent operation-execution lifecycle.

Please use the updated #1515 as the source of truth for this refactoring. The current code is valuable implementation material, but the public execution abstraction must now be operation-generic.

rainman110:
The generic operation API is now implemented and pushed in `d28b5f09`. One architectural point needs direction before this can be considered complete: the current implementation queues synchronous `execute()` on the GTlab owner thread to preserve project/Qt affinity. This means a long-running operation blocks that event loop. Status and cancellation remain accessible from another thread, but an uncooperative operation prevents the owner thread from processing project lifecycle or shutdown requests, so bounded shutdown cannot be guaranteed.

How should we proceed with synchronous `execute()`?

1. Keep execution on the owner thread and explicitly accept/document that shutdown is cooperative and cannot be bounded for an uncooperative operation; or
2. Run `execute()` on a dedicated execution thread and define how project-bound operations receive a thread-safe execution-local project/context without moving arbitrary GTlab objects across threads.

The current patch follows option 1 for thread-affinity safety, but it does not meet the bounded-shutdown requirement. Which direction should this PR take?

PR #1521, opened 2026-08-03
Bool property activation in property connector

jensschmeink:
The introduced changes activate bool properties in the connection editor.

There is needed some discussion how to handle this in detail as some default properties, like "Skip" are then added to the property connection editor as well.

Now it looks like this:

image

## Checklist:

## Summary by Sourcery

Activate boolean properties for selection in the process property connection editor.

New Features:
- Enable boolean properties in the process connection editor and identify them with a dedicated icon.

Enhancements:
- Simplify connection item data access and empty-value handling without changing behavior.

### DIFF
--- a/src/app/mdi_items/process_env/entities/gt_processpropertyportentity.cpp
+++ b/src/app/mdi_items/process_env/entities/gt_processpropertyportentity.cpp
@@ -109,10 +109,16 @@ GtProcessPropertyPortEntity::connectPort(
         GtProcessPropertyConnectionEntity* connection)
 {
     // check connection
-    if (!connection) return false;
+    if (!connection)
+    {
+        return false;
+    }
 
     // check whether connection already exists
-    if (m_connections.contains(connection)) return false;
+    if (m_connections.contains(connection))
+    {
+        return false;
+    }
 
     // check port type
     if (m_type == GtProcessPropertyPortEntity::INPUT_PORT)
@@ -274,7 +280,10 @@ GtProcessPropertyPortEntity::canConnect(GtProcessPropertyPortEntity* port)
 QVariant
 GtProcessPropertyPortEntity::propertyValue()
 {
-    if (!m_item) return {};
+    if (!m_item)
+    {
+        return QString();
+    }
 
     return m_item->propertyValue();
 }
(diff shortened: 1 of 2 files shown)


src/app/mdi_items/process_env/gt_processconnectionitem.cpp
```diff
@@ -263,55 +245,39 @@ GtProcessConnectionItem::data(int column, int role)
...
     {
-        if (m_component)
-        {
-            return m_component->uuid();
-        }
+        if (m_component) return m_component->uuid();
     }
 
-    return QString();
+    return {};;
```
jensschmeink:
```suggestion
    return {};
```

src/app/mdi_items/process_env/gt_processconnectionitem.cpp
```diff
@@ -263,55 +245,39 @@ GtProcessConnectionItem::data(int column, int role)
...
-            {
-                return false;
-            }
+            else return false;
         }
     }
     }
 
-    return QVariant();
+    return {};;
```
jensschmeink:
```suggestion
    return {};
```

src/app/mdi_items/process_env/gt_processconnectionitem.cpp
```diff
@@ -335,11 +301,9 @@ GtProcessConnectionItem::propertyId()
         {
             return GtPropertyReference(m_property->ident()).toString();
         }
-
-
     }
 
-    return QString();
+    return QString{};
```
jensschmeink:
```suggestion
    return {};
```

src/app/mdi_items/process_env/entities/gt_processpropertyportentity.cpp
```diff
@@ -282,7 +282,7 @@ GtProcessPropertyPortEntity::propertyValue()
 {
     if (!m_item)
     {
-        return QVariant();
+        return QString();
```
jensschmeink:
```suggestion
        return QVariant();
```

jensschmeink:
@rainman110 For the bot everything is fine but I would like to have a human review aswell.

rainman110:
@jensschmeink Looks good, but i think we should explicitly remove the  skip property of process elements.

rainman110 (changes requested):
As written before, the skip property should be removed.

rainman110 (approved):
After discussion, lets keep the skip property as it allows a dynamic skipping of calculators

jensschmeink:
> As written before, the skip property should be removed.

As discussed in a recent call we keep SKIP active as it can have benefits in process control

Issue #1526, opened 2026-09-03
[Epic] Introduce transport-neutral executable operations


rainman110:
## Status and normative role

This epic is the **normative architecture source** for the executable-operation model.

Implementation issues and ADRs must preserve the decisions marked as **decided** below. They may refine details explicitly marked **open**, but must not introduce competing execution, serialization, registration, result-application, or project-lifecycle models.

Related work:

- #1515 — single-project headless runtime with asynchronous executable-operation support
- PR #1520 — older task-specific runtime draft; useful implementation input, but not normative
- #1527 — materialize this architecture into ADR + developer documentation

If PR #1520 or older task/runtime concepts conflict with this epic, **this epic and the updated #1515 take precedence**.

## Motivation

GTlab shall support outsourcing computational work without coupling executable domain logic to a particular execution location or transport.

Examples include tasks/calculators, Intelligraph computations, and module-provided computational operations.

The same operation contract shall ultimately support:

- in-process execution;
- isolated local one-shot workers;
- remote one-shot workers;
- later, resident single-project sessions.

Snapshot/one-shot execution and Resident Project Execution may share operation, event, result, and runtime infrastructure, but remain different execution profiles with different project/session lifecycles.

The initial implementation deliberately focuses on a thin one-shot worker slice. Broker, queue, cluster scheduling, and resident-session synchronization are not part of the first milestone.

## Core architecture principle — decided

> **Executable domain logic is independent of execution location and transport.**

Tasks, calculators, Intelligraph components, and other domain implementations must not contain logic for process creation, local-vs-remote selection, control-plane submission, project transfer, IPC/network transport, execution-event transport, or result transport.

The existing task-specific `GtCoreProcessExecutor` remains task-specific and shall **not** be generalized into the arbitrary-operation abstraction.

Existing tasks are integrated through an adapter:

```text
existing GtTask / Calculator
        |
        v
ProcessTaskOperation
        |
        v
generic operation/runtime infrastructure
```

## Executable operation — decided

An executable operation is a normal GTlab domain object:

```cpp
class GtExecutableOperation : public GtObject
{
public:
    virtual bool requiresProject() const = 0;

    // originating/client side
    virtual std::unique_ptr<GtObject>
    createData(const GtExecutionContext& context) const = 0;

    // execution side
    virtual std::unique_ptr<GtObject>
    execute(GtOperationExecutionContext& context) = 0;

    // originating/client side
    virtual void
[...]

Issue #1530, opened 2026-09-03
[Operations] Add generic executable-operation command to the GTlab batch worker


rainman110:
Parent epic: #1526
Depends on: #1528, #1529, #1515
Normative contract: ADR 0001 (`docs/dev/architecture/decisions/0001-executable-operations-one-shot-worker.rst`)

## Goal

Add the generic one-shot executable-operation command to the GTlab batch/console application as a **boundary adapter around `GtHeadlessProjectRuntime`**.

The command must prove the process boundary without introducing a second execution lifecycle. It reconstructs serialized boundary inputs, configures the stdio adapter and delegates execution to `GtHeadlessProjectRuntime::submitOperation(...)`.

The command must remain generic and contain no `GtTask`-specific execution path.

## Required flow

```text
serialized operation
+ optional detached data
+ optional serialized project
        |
        v
GTlab batch command
    |-- load required modules
    |-- reconstruct + validate operation through GtObjectFactory
    |-- reconstruct optional detached data through normal object infrastructure
    |-- initialize GtHeadlessProjectRuntime
    |-- open/provision project when supplied/required
    |-- attach stdio encoder to the runtime execution event stream
    |-- submit operation to GtHeadlessProjectRuntime
    |-- observe/wait for runtime terminal state through the handle
    |-- encode terminal detached result OR structured failure
    |-- close/shutdown runtime deterministically
        |
        v
       exit
```

The batch command must **not** construct its own competing execution lifecycle. In particular it must not architecturally:

```text
construct GtOperationExecutionContext
establish execution scope itself
call operation.execute() directly
manage operation state/cancellation/result ownership itself
```

Those responsibilities belong to `GtHeadlessProjectRuntime` from #1515.

## Boundary inputs

Support the minimum Worker Slice 1 inputs required by ADR 0001:

- serialized `GtExecutableOperation` Memento/XML;
- optional serialized detached invocation data (`GtObject` Memento/XML);
- optional serialized/provisioned project input according to the existing batch/project-loading facilities.

Requirements:

- load modules/type metadata before reconstructing module-provided operations;
- reconstruct operation/data exclusively through existing Memento/ObjectFactory infrastructure;
- reject malformed/unknown operation types and reconstructed non-`GtExecutableOperation` objects before submission;
- if `requiresProject() == true`, reject submission when no valid project is provisioned;
- if `requiresProject() == false`, allow execution without a loaded project;
- do not treat project state as operation data.

[...]

Issue #1529, opened 2026-09-03
[Operations] Add execution events for executable operations


rainman110:
Parent epic: #1526
Depends on: #1528
Architecture decision: ADR 0001 (`docs/dev/architecture/decisions/0001-executable-operations-one-shot-worker.rst`)

## Motivation

Executable operations currently use Qt signals and slots to report progress and other observations to code in the same process. This works well for local execution, but Qt signals cannot carry these updates across a process or machine boundary. A remote worker therefore needs a separate and reliable way to make its progress available.

This issue proposes one event system that operations use in both cases. Operations publish events without knowing whether another part of GTlab receives them locally, writes them to a file, or forwards them later. The first worker integration writes events to a separate file, so normal logs stay readable.

## Scope

- Add `GtExecutionEventStream` as the event source for one operation execution. Operations publish through `context.events().publish(...)`.
- Give every execution one shared ID. The operation context and its event stream must expose that same ID.
- Give each event a sequence number, a non-empty event type, and an optional JSON payload. Event payloads are JSON values only; they are not `GtObject` or Memento data.
- Use Qt signals and slots for local observers. A file writer and optional compatibility adapters are ordinary observers.
- Provide a worker-owned `events.ndjson` file. It contains one compact JSON event per line. Normal stdout, stderr, and worker logs stay in separate files or streams.
- Keep the stdio V1 encoder only as a compatibility/fallback adapter. It must not claim to control unrelated stdout output.

## Required behavior

Events from one execution are numbered `0, 1, 2, ...`. Events published at the same time, or from an observer while another event is handled, must still reach every observer in one consistent order. Observer code must not run while the stream holds its internal lock.

The worker or integration layer chooses the event-file path. Core must not depend on `QProcess`, GUI code, or a remote-control component.

Terminal results and failures are lifecycle output, not ordinary events.

## Out of scope

- Runtime scheduling and state management (#1515)
- Batch-worker and task integration (#1530, #1531)
- Remote transport, reconnect, replay, or a persistent event store
- Artifact or large-data transfer
- A generic event-sink or observer-interface hierarchy

## Acceptance criteria

- [ ] An executable operation publishes through `GtExecutionEventStream`; no `GtExecutionEventSink` interface is added.
- [ ] Context and stream use one shared execution ID.
- [ ] Events have a stable ID, gap-free sequence number, event type, and JSON-only payload.
- [ ] Qt observers can receive events locally, including across threads.
[...]

Issue #1528, opened 2026-09-03
[Operations] Introduce executable-operation core foundation


rainman110:
Parent epic: #1526
Normative contract: ADR 0001 (`docs/dev/architecture/decisions/0001-executable-operations-one-shot-worker.rst`), introduced by #1527 / PR #1533

## Goal

Add the minimal Core API for a transport-neutral `GtExecutableOperation`. It must be possible to register, serialize, reconstruct, and execute an operation in-process using existing GTlab mechanisms.

This issue defines the operation contract and its invocation-local execution context. It does not implement an asynchronous runtime, worker transport, event framing, batch command, or task adapter.

## Operation contract

`GtExecutableOperation` derives from `GtObject` and follows ADR 0001: prepare → compute → commit.

- `requiresProject()` states whether execution also needs a project in addition to detached data.
- `createData()` runs on the originating side. It is lightweight, may read the authoritative project, and returns independently owned detached data or `nullptr`.
- `execute()` is the synchronous, potentially expensive compute step. It may modify detached data and an execution-local project, but never the authoritative originating project. It returns independently owned detached, serializable result data or `nullptr`.
- `applyResult()` runs on the originating side. It is lightweight and is the only step that applies a result to the authoritative project. It receives a nullable, read-only, non-owning result pointer and must report structured success or failure.

Do not add an operation-data hierarchy, result-applicator registry, or another execution model.

## Execution context

`GtOperationExecutionContext` is valid for exactly one `execute()` call. It provides:

- mutable, non-owning access to optional detached data; the operation must not retain the pointer;
- an immutable, opaque, globally unique execution ID with stable text form;
- direct access to the invocation-local `GtExecutionEventStream`;
- cooperative cancellation state that `execute()` can read while another thread requests cancellation.

It does not own, load, or synchronize project state. Project-bound execution obtains its project separately through `GtExecutionContextScope`.

Do not add a `GtExecutionEventSink` interface. The concrete stream is both the operation-facing publisher and the local observable source; local consumers use Qt signals and slots. Event envelopes, ordering, transport, and file/process adapters belong to #1529.

## Module registration and serialization

Add the optional Core/headless `GtOperationInterface` using the existing module mechanism. Modules declare operation `QMetaObject`s; the normal module loader validates that each one is a `GtExecutableOperation` and registers it through `GtObjectFactory`. Module authors need no second registration step.

[...]

PR #1533, opened 2026-09-03
docs: finalize executable operation worker contract

### LINKED ISSUE #1527: [Operations] Finalize the one-shot worker slice design and developer contract
Parent epic: #1526
Related runtime issue: #1515
Related draft implementation: PR #1520

## Implementation instruction — read this first

This issue is **not a green-field architecture task**.

Before changing documentation or code, read **#1526 completely**. The epic is the normative source of truth for the executable-operation architecture. The decisions marked there as **decided** must be transferred into the ADR/developer documentation and must not be redesigned.

Also read the updated #1515 for the `GtHeadlessProjectRuntime` role. PR #1520 is useful implementation/reference material only. If PR #1520 conflicts with #1526/#1515, the issues take precedence.

Do **not** introduce new competing abstractions merely because an exact name or implementation detail is still open. In particular, do not introduce a second operation factory/registry, a task-specific runtime execution model, a second project lifecycle, direct worker-owned execution semantics, or an operation-specific metadata/type-ID layer.

If an implementation detail is explicitly listed as open in #1526 and cannot be resolved from existing GTlab conventions, document the smallest concrete choice needed by Worker Slice 1 rather than broadening the architecture.

## Goal

Materialize the architecture already decided in #1526 into an ADR and developer documentation precise enough that the subsequent implementation issues can be executed without reopening architecture.

The first target is the one-shot worker slice:

```text
serialized project / operation / optional data
                 |
                 v
GTlab one-shot batch worker
    |
    +-- reconstruct process-boundary input
    +-- configure stdio protocol adapter
    |
    v
GtHeadlessProjectRuntime
    |
    +-- project lifecycle/context
    +-- async operation lifecycle
    +-- execution handle/cancellation/events
    |
    v
GtExecutableOperation::execute()
```

The worker is an adapter around the runtime. It does not own an independent execution lifecycle and does not architecturally call `execute()` directly.

## Decisions that must appear in the ADR

The ADR must faithfully capture the decided contracts in #1526, including at least:

- `GtExecutableOperation` is a `GtObject`;
- normal GTlab Memento/ObjectFactory infrastructure is reused for operation/data/result serialization and reconstruction;
- reconstructed objects are validated as `GtExecutableOperation` before execution;
- `createData() -> execute() -> applyResult()` responsibilities and thread/performance rules;
- project state is separate from detached operation data;
- `GtExecutionContext` remains the project context;
- `GtOperationExecutionContext` carries invocation-local data/events/cancellation/execution identity;
- `GtOperationInterface` is the authoritative module declaration of operation types;
- no second operation-specific factory/global registry is introduced initially;
- Worker Slice 1 follows the existing GTlab module metadata convention and may expose `QMetaObject` directly for operation declaration/registration;
- no new `GtTypeId`, `GtTypeDescriptor`, operation-specific metadata wrapper, or placeholder `GtMetaObject` is introduced for this slice;
- future migration of module interfaces from `QMetaObject` to the planned `GtMetaObject` abstraction belongs to separate GTlab 3/type-system work;
- operation invocation/submission is asynchronous from the caller's perspective;
- `GtExecutableOperation::execute()` itself remains synchronous;
- execution-local operation/data ownership lasts through terminal execution-side state;
- the originating operation instance remains alive through client-side `applyResult()`;
- `GtHeadlessProjectRuntime` is the single-project execution-side runtime/application boundary;
- runtime submission returns a generic execution handle before expensive execution blocks the submitting control path;
- status, cancellation, and event observation remain meaningful while execution is running;
- execution-side completion differs from future client-visible completion after transport + `applyResult()`;
- tasks enter the generic runtime through `ProcessTaskOperation`;
- `GtCoreProcessExecutor` remains task-specific and may be reused inside `ProcessTaskOperation`, not as the generic operation runtime;
- event production is transport-neutral: operation -> context -> event sink/stream -> adapter;
- Qt signals/slots are the preferred local observation mechanism;
- the one-shot worker delegates execution lifecycle to `GtHeadlessProjectRuntime`.

## Async/threading constraint

The ADR must explicitly prevent a fake asynchronous implementation such as:

```cpp
operation->execute(context); // potentially long running
return alreadyFinishedHandle;
```

At the same time, do not blindly require arbitrary GTlab project objects to be moved to a background thread. Project/thread-affinity constraints must be preserved.

The required invariant is:

> Long-running `execute()` must not make status, cancellation, and event observation unusable until `execute()` returns.

The detailed scheduler/marshalling implementation may remain open if it belongs to #1515, but the invariant and ownership boundary must be explicit.

## Runtime completion vs. client completion

Document these as separate lifecycle boundaries:

```text
Execution-side runtime completion
    = execute() finished
      + detached result/failure available

Future client invocation completion
    = execution-side completion
      + transport/reconstruction
      + applyResult() completed on originating side
```

Worker Slice 1 implements the first boundary. A future `GtOperationExecutor` owns the second.

## Stdio V1 framing

Finalize the minimal V1 stdio framing needed by #1529/#1530.

`stdout` is **not** an exclusive protocol channel. Normal GTlab log/diagnostic/status output may be interleaved with protocol output.

The framing must define concretely:

- how a protocol record is recognized unambiguously;
- record boundaries;
- how non-protocol stdout content is separated/ignored/forwarded by a decoder;
- how domain events, terminal success/result, and terminal failure records are distinguished.

A decoder must never assume every stdout line is protocol data.

Keep this protocol minimal for a one-shot worker. Do not design reconnect, replay, broker, queue, or a general remote protocol.

## Relationship to PR #1520

The ADR/developer docs should explain how the existing draft evolves:

### Reuse/generalize

- single-project runtime lifecycle;
- runtime-owned asynchronous execution state;
- opaque/value-based handles;
- status/cancel/wait concepts;
- terminal result/error snapshots;
- bounded shutdown;
- owner-thread/event-loop integration where required;
- explicit `GtExecutionContext` and mutation coordination.

### Replace/generalize

- `GtHeadlessTaskHandle` -> generic execution handle/equivalent;
- `GtHeadlessTaskStatus` -> generic execution status/equivalent;
- `submitTask()` -> generic operation submission as the architectural execution API;
- execution state fundamentally tied to `GtTask*` -> operation/execution state independent of task types;
- runtime hard-wiring to `GtCoreProcessExecutor` -> generic runtime, with task machinery behind `ProcessTaskOperation`.

Task lookup/addressing helpers may remain later as task-specific convenience/compatibility APIs, but are not the generic execution contract.

## Deliverables

- ADR containing the decided architecture from #1526;
- developer documentation for operation authors, runtime implementers, and worker implementers;
- compact diagrams for ownership/lifecycle, project context, runtime submission, event flow, and result flow;
- explicit execution-side vs. client-side completion documentation;
- module-author registration example using the existing `QMetaObject`-based module convention for Worker Slice 1;
- concrete minimal V1 stdio framing contract;
- a short migration note explaining how PR #1520 is to be generalized.

## Non-goals

- implementing the operation/runtime/worker APIs in this issue;
- client-side `GtOperationExecutor` implementation;
- client-side `applyResult()` orchestration;
- remote/control-plane integration;
- GUI interactor implementation;
- resident-session synchronization;
- broker/queue/cluster design;
- event replay/reconnect;
- complete future `GtMetaObject` design;
- introducing a placeholder `GtMetaObject`, `GtTypeId`, `GtTypeDescriptor`, or operation-specific metadata abstraction for Worker Slice 1;
- preserving task-specific runtime APIs from PR #1520 as the generic execution contract.

## Acceptance criteria

- [ ] The implementation explicitly states that #1526 is the normative architecture source.
- [ ] ADR and developer documentation reproduce all relevant **decided** contracts from #1526 without introducing competing abstractions.
- [ ] The ADR explicitly uses the current `QMetaObject`-based module registration model for Worker Slice 1 and defers `GtMetaObject` migration to separate type-system work.
- [ ] No operation-specific type-ID/metadata compatibility layer is introduced merely to anticipate GTlab 3.
- [ ] The async boundary is explicit: submission returns a handle before expensive execution blocks the submitting control path; `execute()` remains synchronous.
- [ ] Ownership/lifetime of originating operation, execution-local operation, detached data, and detached result are documented.
- [ ] Async status/cancellation/event usability while `execute()` runs is documented as an invariant without violating project thread-affinity constraints.
- [ ] `GtHeadlessProjectRuntime` is documented as the execution-side owner of project and operation lifecycle.
- [ ] The worker is documented as an adapter around the runtime, not an independent execution runtime.
- [ ] Execution-side completion and future client-side completion after `applyResult()` are clearly distinguished.
- [ ] The V1 stdout framing is concrete enough for #1529 to implement encoder/decoder tests and separates protocol records from normal stdout messages.
- [ ] The ADR explains precisely how PR #1520 should be generalized from task-specific async execution to generic operation execution.
- [ ] No unresolved architecture question blocks #1528, #1529, #1515, or #1530; remaining open points are implementation details explicitly allowed by #1526.

rainman110:
## Description

Implements #1527 as the documentation-only foundation for Worker Slice 1.

- Adds the normative executable-operation ADR derived from #1526.
- Adds developer guidance for operation/module/runtime/worker implementers.
- Defines the minimal V1 stdout framing contract and explains the generalization of PR #1520.

No operation, runtime, event, or worker API is implemented in this PR.

## How Has This Been Tested?

- `source docs/docenv/bin/activate && cd docs/dev && make`

## Checklist:

## Summary by Sourcery

Document the executable-operation and one-shot worker contract as the foundation for Worker Slice 1.

Enhancements:
- Define the normative contract for executable operations and one-shot workers, including minimal V1 stdout framing and guidance for implementers.

Documentation:
- Add architecture documentation and an ADR covering executable operations, worker behavior, and the planned generalization of the operation contract.


rainman110:
The ADR is a good consolidation of the large architectural boundaries, but I think it still leaves too many contracts open for #1527 to be considered complete. The purpose of #1527 is precisely to avoid pushing architectural decisions into #1528, #1515, #1529, or #1530.

Please tighten the ADR where we already have a clear design direction instead of delegating these points to follow-up implementation issues.

In particular:

1. **Execution handle semantics**
   - `submitOperation(...)` returns a generic asynchronous execution handle immediately.
   - The handle is copyable/value-like and refers to runtime-owned execution state.
   - It exposes no `GtObject*`, `GtProject*`, `GtTask*`, or other process-local object pointers.
   - It provides at least execution identity, status, cancellation request, optional blocking wait, and terminal result/error access.
   - Runtime execution state must remain valid at least as long as handles refer to it; shutdown/close must result in a defined terminal state rather than dangling handles.
   - `cancel()` requests cancellation; acceptance of the request is distinct from the execution already being cancelled/terminated.
   - The detached `GtObject` returned by `GtExecutableOperation::execute()` is the runtime execution result and becomes available only in the appropriate successful terminal state.

2. **Threading / cancellation contract**
   The current invariant that status/cancellation/events remain usable while `execute()` is running is important, but it needs an API-level consequence. At minimum, please specify that status inspection and cancellation request state are thread-safe / callable independently of the thread currently blocked in `execute()`. The cancellation token/request state exposed to the operation must therefore also be safe to observe across that boundary.

   Do not prescribe blindly moving GTlab project objects to arbitrary worker threads. The runtime may use owner-thread execution, existing non-blocking machinery, safe worker threads, or explicit marshalling as appropriate. But a design in which cancellation can only be requested after the synchronous `execute()` call returns does not satisfy the asynchronous contract.

3. **Minimal `GtOperationExecutionContext` contract**
   #1528 must implement this type, so it should not have to invent its semantics. Please define the minimal contract now:
   - detached invocation data access and ownership/lifetime;
   - execution identity type/semantics;
   - event sink access;
   - cancellation-token access;
   - explicit statement that project context is provided separately through `GtExecutionContextScope` and is not owned by this context.

   Exact class spelling may still change, but the semantics should be normative.

4. **`applyResult()` failure/result contract**
   The ADR currently sketches `applyResult()` as `void`, while client-visible completion is defined as successful only after `applyResult()` has completed successfully. That leaves no structured way to represent a failed authoritative commit.

   Please resolve the semantic contract here: either `applyResult()` returns a structured success/failure result, or document another explicit mechanism by which result-application failure becomes part of the client invocation lifecycle. The exact concrete result type may follow existing GTlab conventions, but the failure path must not remain undefined.

5. **Event envelope**
   #1529 is expected to implement the event type/envelope. Please define the minimal logical fields in this ADR instead of leaving the complete event model to #1529. At minimum we need:
   - execution identity;
   - per-execution ordering/sequence;
   - domain event key/type;
   - payload contract / allowed transportable value shape.

   Qt signals/slots remain the local observation mechanism, but the logical event model must stay transport-neutral.

6. **`GtTypeId` -> current factory/type infrastructure**
   The ADR requires `GtOperationInterface` to expose `GtTypeId`, but the current Core object factory registers concrete Qt metadata (`QMetaObject`) and there is no existing public `GtTypeId` abstraction yet. #1528 must not be forced to invent a second metadata system just to make operation registration work.

   Please specify the minimal bridge for the first implementation: the public operation interface uses a GTlab-owned stable type identifier, while the module-loader/type-infrastructure layer resolves that identifier to today's concrete metadata/factory representation internally. If this requires introducing a small `GtTypeId` value type or a descriptor that contains both stable identity and current factory metadata internally, define that direction explicitly. The important constraint remains that `QMetaObject` must not leak into the new public operation extension contract.

7. **Stdio V1 framing**
   The prefix + one-line compact JSON approach is good. Please make the V1 contract concrete enough that #1529 does not invent protocol semantics independently:
   - exact required common fields;
   - minimal fields for `event`, `result`, and `failure`;
   - how Memento/XML or other payload bytes are represented so a record cannot contain an unescaped newline;
   - exactly one terminal record per execution;
   - protocol writes must be serialized so an ordinary/log write from another thread cannot interleave bytes into the middle of a protocol record.

   Non-prefixed stdout remains ordinary output and must be ignored/forwarded separately by a decoder.

8. **Worker/runtime dependency**
   Please make it explicit that Worker Slice 1 depends on the generic `GtHeadlessProjectRuntime` from #1515. The worker command delegates lifecycle to `GtHeadlessProjectRuntime::submitOperation(...)`; it should not construct an execution context and call `operation.execute()` itself as an architectural path.

These points do not require locking down every implementation detail. Internal scheduler classes, shared-state implementation, exact enum spelling, GUI APIs, remote replay/reconnect, resident-session synchronization, and the future full client-side `GtOperationExecutor` can remain open.

The goal is that after this ADR, the agents implementing #1528, #1515, #1529, and #1530 should be choosing implementation details, not inventing the execution model.

rainman110:
Architecture update: the type-registration requirement has changed after further discussion and the related issues (#1526, #1527, #1528) have been updated accordingly.

For Worker Slice 1, please **do not introduce `GtTypeId`, `GtTypeDescriptor`, an operation-specific metadata wrapper, or a placeholder `GtMetaObject`** for operation registration.

Instead, follow the existing GTlab module/type-registration convention and let `GtOperationInterface` expose `QMetaObject` directly, e.g. conceptually:

```cpp
class GtOperationInterface
{
public:
    virtual QList<QMetaObject> operations() const { return {}; }
};
```

The module loader should validate the declared operation types and register them through the existing `GtObjectFactory` / current GTlab type infrastructure.

This intentionally supersedes point 6 of my earlier review comment and the `GtTypeId` bridge added in the latest commit.

Rationale: the planned future `GtMetaObject` abstraction will be part of broader GTlab 3 / type-system work and is expected to represent richer GTlab runtime type metadata, not merely act as a thin compatibility wrapper around `QMetaObject`. We therefore do not want to design or pre-implement a placeholder version here. When `GtMetaObject` is ready, operation registration should migrate together with the other module interfaces.

Please update the ADR and developer documentation accordingly while keeping the rest of the latest execution-contract clarifications intact.

rainman110:
The latest simplification around operation type registration looks good for Worker Slice 1: using the existing `QMetaObject`/`GtObjectFactory` convention is preferable to introducing a one-off `GtTypeId` bridge before the broader GTlab type-system migration.

I only see two small points left before #1527 is complete:

1. **Execution-handle result ownership/access**
   The ADR now says that the handle is copyable/value-like, exposes no process-local object pointers, and provides terminal result access to the detached `GtObject` returned by `execute()`. Please make the ownership semantics explicit so #1515 does not have to invent them.

   At minimum, define that the terminal result is retained by the runtime-owned execution state and that reading it through one handle copy must not consume, move, or invalidate the result for other handle copies. The concrete API may still be chosen by #1515 (for example a serialized/value snapshot, explicit clone, or another non-dangling access form), but a one-shot ownership transfer from a copyable handle should not be the implicit contract.

2. **Truncated sentence in the developer guide**
   In `docs/dev/howtos/executable_operations.rst`, the module-registration section currently ends with:

   > `A future GtMetaObject migration belongs to the broader GTlab type-system work and`

   and then jumps to `Runtime implementers`. Please complete/fix that sentence.

With these two points addressed, I do not currently see another architectural blocker for #1527.

rainman110:
One final documentation-structure request before merge: the new ADR and executable-operations guide should not be linked directly from the top-level documentation hierarchy.

The current developer docs are organized around broad navigation areas such as Start here, Module development, Concepts, Maintenance and support, and Core contributors. The operation ADR is currently placed directly under Concepts, and the executable-operations guide directly under Maintenance and support. That makes both pages appear at the same level as general concepts or user-facing maintenance guides, which does not reflect their role.

Please introduce an explicit architecture hierarchy under the Core contributors area instead.

Suggested structure:

```text
Core contributors
└─ Architecture
   ├─ Executable operations
   └─ Technology decisions
      └─ ADR 0001 — Executable operations and one-shot worker
```

Concretely, a structure like this would fit well:

```text
docs/dev/
├─ architecture/
│  ├─ index.rst
│  ├─ executable_operations.rst
│  └─ decisions/
│     ├─ index.rst
│     └─ 0001-executable-operations-one-shot-worker.rst
```

`architecture/index.rst` should be the architecture landing page for Core contributors and link to both the executable-operations architecture guide and the technology-decision collection.

`architecture/decisions/index.rst` should be a short ADR overview explaining what ADRs are and listing accepted decisions through a toctree. This gives us a proper home for future ADRs instead of exposing individual ADRs directly in the root navigation.

The executable-operations page should be moved out of `howtos/`: its current content is not really a task-oriented how-to, but an architecture/developer guide covering operation authors, module authors, runtime implementers, worker implementers, task integration, etc.

At the root `docs/dev/index.rst`, please only add the Architecture landing page under `Core contributors`, e.g.:

```rst
.. toctree::
   :caption: Core contributors
   :maxdepth: 2
   :hidden:

   basics/building
   architecture/index
   api
   genindex
```

The root navigation should describe documentation areas, not individual architecture topics or ADRs.

No content-level architecture changes are requested here; this is only about integrating the new material cleanly into the existing documentation hierarchy.

PR #1534, opened 2026-09-03
[remote execution] Add executable operation core foundation

### LINKED ISSUE #1528: [Operations] Introduce executable-operation core foundation

rainman110:
## Description

Implements #1528.

- Adds the transport-neutral `GtExecutableOperation` contract and structured `GtOperationApplyResult`.
- Adds the invocation-local execution context with detached data access, execution identity, event-sink boundary, and cooperative cancellation token.
- Adds optional `GtOperationInterface` discovery through the existing process-module loader and `GtObjectFactory`; no operation registry, serializer, or secondary factory is introduced.
- Adds normal Memento/ObjectFactory roundtrip coverage, execution-context coverage, and rejection of reconstructed non-operation objects.
- Updates executable-operation architecture guidance to distinguish this Core foundation from follow-up runtime/event/worker work.

API note: `applyResult()` receives a nullable non-owning `GtObject const*`, and the result type exposes `succeeded()` plus `errorMessage()`. The event sink is intentionally the minimal event-type boundary for this slice; ordered envelopes and streaming remain #1529 scope.

## How Has This Been Tested?

- `cmake --build build --target GTlabUnitTest -j$(nproc)`
- `./build/GTlabUnitTest --gtest_filter="GtExecutableOperation.*:GtOperationApplyResult.*"`
- `source docs/docenv/bin/activate && cd docs/dev && make`

## Checklist:

## Summary by Sourcery

Establish the Core foundation for transport-neutral executable operations and their module-based discovery.

New Features:
- Introduce the transport-neutral executable operation contract with detached data preparation, execution, and result application phases.
- Provide invocation execution context support for execution identity, shared cancellation, detached data, and event publication.
- Enable optional executable-operation discovery and registration through process modules and the object factory.

Enhancements:
- Add structured apply-result status reporting with success and error details.
- Define Core architectural guidance and module interface documentation for executable operations.

Build:
- Include the new operation interfaces and context sources in the Core build and public include paths.

Documentation:
- Update executable-operation architecture decisions and add operation module interface guidance.

Tests:
- Add coverage for operation round trips, execution context value semantics, cancellation, event publication, apply statuses, module registration, duplicate declarations, and invalid operation rejection.

Chores:
- Update the Cppcheck suppression baseline.


src/core/operations/gt_executableoperation.h
```diff
@@ -0,0 +1,42 @@
...
+class GT_CORE_EXPORT GtExecutableOperation : public GtObject
+{
+    Q_OBJECT
+
+public:
+    using GtObject::GtObject;
+    ~GtExecutableOperation() override = default;
+
+    virtual bool requiresProject() const = 0;
+
```
rainman110, line 40: Please document the semantics of the public operation methods directly in the header, not only at class level / in the ADR. In particular, callers and operation authors need to know for `requiresProject()`, `createData()`, `execute()` and `applyResult()` which side/context they run on, ownership/lifetime of returned or passed objects, whether `nullptr` is valid, and that only `applyResult()` performs the authoritative originating-side commit. These contracts are important enough that the API should be usable correctly without reading the ADR first.

src/core/gt_operationinterface.h
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * @brief Optional module declaration for executable operation types.
+ */
+class GT_CORE_EXPORT GtOperationInterface
+{
+public:
+    virtual ~GtOperationInterface() = default;
+
+    virtual QList<QMetaObject> operations() const { return {}; }
```
rainman110: Please expand the Doxygen for this interface and especially `operations()`. It should state that the returned `QMetaObject`s must describe reconstructable `GtExecutableOperation` classes, that the normal module loader validates and registers them in `GtObjectFactory`, and that module authors must not perform a second registration step. Also document the invokable-constructor requirement implied by ObjectFactory reconstruction.

src/core/gt_operationexecutioncontext.h
```diff
@@ -0,0 +1,88 @@
...
+    virtual void publish(QString const& eventType) = 0;
+};
+
+/**
+ * @brief Invocation-local services passed to one GtExecutableOperation::execute call.
+ *
+ * The context never owns project state or the detached data object. The caller
+ * owns the data for the complete invocation; operations must not retain it.
+ */
+class GT_CORE_EXPORT GtOperationExecutionContext
```
rainman110: Please document the public execution-context accessors individually. The important contracts are currently mostly implicit: `data()` is borrowed invocation data and must not be retained, `executionId()` is stable for the invocation, `events()` is a transport-neutral observation sink, and the cancellation token shares cooperative request state that may be observed while another thread requests cancellation. These lifetime/threading semantics should live at the API boundary.

rainman110 (commented):
The overall architecture is aligned with ADR 0001: the operation contract stays transport-neutral, the normal Memento/ObjectFactory infrastructure is reused, project context remains separate from the invocation context, and no parallel registry/serializer/runtime model is introduced.

Before merge, please address the public API documentation and registration-test gaps below.

In addition to the inline comments:

- Please add `GtOperationInterface` to the module-developer interface documentation (`docs/dev/modules/interfaces.rst`). It is now a public optional module capability and should be discoverable alongside the existing module interfaces, not only from the architecture guide.
- Please address the existing Sourcery finding in `GtProcessModuleLoader`: duplicate entries within one `operations()` list are currently not rejected because `containsDuplicates()` only checks against classes already present in the factory.
- Please test the actual `GtOperationInterface -> GtProcessModuleLoader -> GtObjectFactory` path. The current roundtrip test manually calls `gtObjectFactory->registerClass(...)`, so it does not prove the module-registration acceptance criterion from #1528. The new loader code is consequently currently uncovered.

The implementation itself otherwise looks consistent with the agreed architecture.

src/core/process_management/gt_processmoduleloader.cpp
```diff
@@ -87,6 +89,33 @@ GtProcessModuleLoader::check(GtModuleInterface* plugin) const
...
     }
 
+    GtOperationInterface* operationInterface =
+        checkInterface<GtOperationInterface>(plugin->ident(), plugin);
+
+    if (operationInterface)
+    {
+        const QList<QMetaObject> operations = operationInterface->operations();
+
+        if (gtObjectFactory->containsDuplicates(operations))
```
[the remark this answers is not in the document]
rainman110: Addressed in 90415c73. The loader now rejects duplicate class names within a single operations() declaration before factory registration, with focused coverage for that path.

src/core/operations/gt_executableoperation.h
rainman110: Addressed in 90415c73. The public header now documents the originating-side, lightweight prepare/commit steps; the synchronous, potentially expensive execution-side step; ownership and nullable-result semantics; and the exclusive authoritative-commit role of applyResult().

src/core/gt_operationinterface.h
rainman110: Addressed in 90415c73. GtOperationInterface::operations() now documents the GtExecutableOperation, invokable-constructor, loader-validation, GtObjectFactory-registration, and no-second-registration requirements. The module-developer interface guide now also includes this optional capability and example.

src/core/gt_operationexecutioncontext.h
rainman110: Addressed in 90415c73. GtExecutionId, GtCancellationToken, GtExecutionEventSink, and each GtOperationExecutionContext accessor now carry direct API documentation for identity stability, borrowed-data lifetime, transport-neutral event publication, and concurrent cooperative cancellation observation.

rainman110:
The interface documentation still does not follow the existing documentation structure.

`docs/dev/modules/interfaces.rst` is the overview/routing page. The existing public module capabilities use abstract documentation names there (`Module interface`, `Data-model interface`, `Process interface`, etc.) and each has its own dedicated subpage under `docs/dev/modules/interfaces/` with a `:ref:` label. For example, `process.rst` is headed `Process interface` and is linked from the overview through `:ref:`processinterface``.

Please apply the same pattern to executable operations instead of documenting the concrete C++ type directly on the overview page:

- add `docs/dev/modules/interfaces/operation.rst`;
- use a label such as `.. _operationinterface:` and the page title `Operation interface`;
- move the detailed `Executable-operation interface` section and example from `interfaces.rst` to that page;
- in the overview table, link it as `:ref:`operationinterface`` rather than displaying the concrete class name ``GtOperationInterface``;
- add `interfaces/operation` to the hidden `toctree` next to the other interface pages.

The concrete class name `GtOperationInterface` should of course be introduced and documented on the dedicated `Operation interface` page, just as the other pages introduce `GtProcessInterface`, `GtDatamodelInterface`, etc. This keeps the navigation and terminology consistent with the rest of the module-interface documentation.

rainman110:
Addressed in c38158f7. The module-interface overview is now limited to routing: the detailed operation guidance moved to docs/dev/modules/interfaces/operation.rst, labelled operationinterface and linked from the overview table and hidden toctree. The developer documentation build passes.

rainman110 (commented):
One additional API-clarity point before merge: the event sink introduced by #1528 is intentionally only a temporary/minimal boundary. Please make that explicit in the code so a human reviewer or API user does not mistake the current `publish(QString)` signature for the final event contract from ADR 0001 / #1529.

src/core/gt_operationexecutioncontext.h
```diff
@@ -0,0 +1,98 @@
...
+ * @brief Transport-neutral boundary for execution-side operation observations.
+ */
+class GT_CORE_EXPORT GtExecutionEventSink
+{
+public:
+    virtual ~GtExecutionEventSink() = default;
+
+    /// Publishes a transport-neutral execution-side observation.
+    virtual void publish(QString const& eventType) = 0;
+};
```
rainman110: This currently looks like the final public event API, but it is intentionally incomplete: ADR 0001 / #1529 defines logical events as `eventType` **plus an optional payload**, with `executionId` and `sequence` added by the event infrastructure. Please mark this clearly as a temporary #1528 boundary/TODO for #1529 (in the Doxygen and/or directly next to the method). Do **not** choose or freeze the payload C++ type in this PR; #1529 should define that together with the logical event envelope. The important thing here is to avoid implying that `publish(QString const& eventType)` is the final contract.

rainman110:
Addressed in 0e248189. GtExecutionEventSink::publish() is now parameterless, so #1528 does not imply a QString event-type contract. The public interface explicitly states that the event envelope, event type, and payload contract remain follow-up runtime work.

rainman110:
@mariusalexander Would you please review this PR? Before, please read the issue description #1528 and the corresponding epic #1526 , which is the epic for the whole remote processing design.

This issue only comes with some foundation classes like event sinks, operations, data flow etc.

The architecture is documented in https://gtlab.readthedocs.io/en/latest/architecture/index.html

docs/dev/architecture/executable_operations.rst
mariusalexander, line 1: The documentation requires deep understanding of the architecture design, even though this page should be the main documentation, right? It is also quite technical IMO. To be clear, this is not meant as a blocker but as a hint for future work.

Some examples of unlcear/technical formulations making this hard to read (for me):

> This guide applies the accepted operation ADR for module, runtime, and worker implementers.
      
I find it hard to understand what "implementers" means in this context.

> #1528 introduces only the transport-neutral operation API
      
What is a "transport-neutral operation" API? 

> Core provides ``GtExecutableOperation`` together with ``GtOperationExecutionContext``,
``GtExecutionId``, ``GtCancellationToken``, ``GtExecutionEventSink``, and the
structured ``GtOperationApplyResult``

Maybe a simple bullet list of the classes involved would be better. Also consider using a dedicated folder "src/core" or "src/core/process_management" to organize the files for the developer too.

> Use ``GtOperationInterface::operations()`` to declaratively contribute operation
``QMetaObject`` values

This too I find awkward to read/understand. Maybe just write "Use ``GtOperationInterface::operations()`` to register execution units/operation units using `QMetaObject` values".

> Operation authors

This section heading is unclear to me. What are "Operation authors"?

src/core/operations/gt_executableoperation.h
```diff
@@ -0,0 +1,73 @@
...
+    using GtObject::GtObject;
+    ~GtExecutableOperation() override = default;
+
+    /**
+     *  Returns whether execution requires a provisioned project.
+     *
+     * This concerns the execution location only. An operation may still read
+     * the originating project while preparing detached invocation data.
+     */
+    virtual bool requiresProject() const = 0;
```
mariusalexander, line 39: I'd be nice if the comments would follow the doxygen syntax like the rest in the code base.

@brief ...
@return ...

src/core/operations/gt_executableoperation.h
```diff
@@ -0,0 +1,73 @@
...
+class GT_CORE_EXPORT GtExecutableOperation : public GtObject
+{
+    Q_OBJECT
+
+public:
+    using GtObject::GtObject;
+    ~GtExecutableOperation() override = default;
+
+    /**
+     *  Returns whether execution requires a provisioned project.
```
mariusalexander: > provisioned project

I find that quite technical too. I'd go with "operates on data inside the project folder" or "operates on data outside the datamodel". Maybe one could rename the method to "requiresProjectFiles" or "operatesOnProjectFiles" but I am fine with the name as is.

src/core/gt_operationapplyresult.h
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * @brief Result of applying a detached operation result on the originating side.
+ */
+class GT_CORE_EXPORT GtOperationApplyResult
+{
+public:
+    static GtOperationApplyResult success();
+    static GtOperationApplyResult failure(QString errorMessage);
+
+    bool isSuccess() const noexcept;
```
mariusalexander: "isSuccess" is awkward. Consider renaming to "hasFailed" (changes function semantics), "isSucessful" or "succeeded". Later reads most nicely:

```cpp
if (!result.succeeded()) 
{
    // error handling
}
```

src/core/operations/gt_operationexecutioncontext.h
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * @brief Opaque, globally unique identity of one operation execution.
+ */
+class GT_CORE_EXPORT GtExecutionId
+{
+public:
+    GtExecutionId();
+
+    /// Returns the stable text representation used by logs and protocols.
+    QString const& toString() const noexcept;
```
mariusalexander, line 47: consider adding operator== and operator!=

src/core/gt_operationexecutioncontext.h
mariusalexander: Please mark comments like these as "resolved" once your done

tests/unittests/core/test_gt_executableoperation.cpp
```diff
@@ -0,0 +1,299 @@
...
+    std::unique_ptr<GtObject>
+    createData(GtExecutionContext const&) const override
+    {
+        return std::make_unique<GtObject>();
+    }
+
+    std::unique_ptr<GtObject>
+    execute(GtOperationExecutionContext& context) override
+    {
+        observedData = context.data() != nullptr;
```
mariusalexander: Consider making this more explicit
```suggestion
        observedData = (context.data() != nullptr);
```

src/core/operations/gt_operationexecutioncontext.h
```diff
@@ -0,0 +1,101 @@
...
+    virtual void publish() = 0;
+};
+
+/**
+ * @brief Invocation-local services passed to one GtExecutableOperation::execute call.
+ *
+ * The context never owns project state or the detached data object. The caller
+ * owns the data for the complete invocation; operations must not retain it.
+ */
+class GT_CORE_EXPORT GtOperationExecutionContext
```
mariusalexander, line 108: Does it make sense to pimpl this class?

src/core/operations/gt_operationexecutioncontext.h
```diff
@@ -0,0 +1,101 @@
...
+{
+public:
+    virtual ~GtExecutionEventSink() = default;
+
+    /// Publishes an execution-side observation without defining its event data.
+    virtual void publish() = 0;
+};
+
+/**
+ * @brief Invocation-local services passed to one GtExecutableOperation::execute call.
```
mariusalexander: "Invocation-local services" unsure what this should tell me. Why not "Context object used by GtExecutableOperation::execute"?

tests/unittests/core/test_gt_executableoperation.cpp
```diff
@@ -0,0 +1,299 @@
...
+    ASSERT_NE(operation, nullptr);
+
+    auto* testOperation = static_cast<TestOperation*>(operation);
+    EXPECT_EQ(testOperation->savedValue(), 42);
+
+    auto data = testOperation->createData(GtExecutionContext{});
+    TestEventSink events;
+    GtCancellationToken cancellation;
+    std::thread cancellationThread([&]() { cancellation.requestCancellation(); });
+    cancellationThread.join();
```
mariusalexander, line 225: Why is "requestCancellation" executed in a separate thread here? To test that it is thread-safe?

tests/unittests/core/test_gt_executableoperation.cpp
```diff
@@ -0,0 +1,299 @@
...
+        return std::make_unique<GtObject>();
+    }
+
+    std::unique_ptr<GtObject>
+    execute(GtOperationExecutionContext& context) override
+    {
+        observedData = context.data() != nullptr;
+        observedCancellation = context.cancellation().isCancellationRequested();
+        observedExecutionId = context.executionId().toString();
+        context.events().publish();
```
mariusalexander, line 86: Must one call `events().publish()`? Best the events would be published automatically after calling execute.

mariusalexander (commented):
I find it hard to understand and follow the API documentation/the EPIC description. The EPIC #1526 and the issue description #1528 are quite long. The amount of text and description needs to shrink for future work.

You have showed me an image of how this new API should operate. This image and a couple of sentences on top would be enough and give (me at least) a better understanding of what this EPIC/Issue is all about.

Besides I find the issue-, epic- and API documentation quite technical:

1. The abbreviation ADR defined and I did not know about this abbreviation (thanks to you I now know that it stands for "Architectural Design Review" but others might not).
2. What is a "transport-neutral operation API"? You probably mean that the execution of the "CommandObjects" may be executed within the same GTlab Instance or on a Remote Instance and that these details should be "hidden"/should not matter. Such a term IMO needs to be defined as well (or a better one needs to be found)

src/core/gt_operationinterface.h
mariusalexander: Please mark comments like these as "resolved" once your done

mariusalexander (changes requested):

docs/dev/architecture/executable_operations.rst
rainman110: The basic idea is to create a design that allows operations to be executed independent on the location or the transport method to the location. 

This works since we decouple data retrieval and result application from the execution on the on hand (neutral execution site) and choose data formats the can be serialized (transport neutral)

src/core/operations/gt_executableoperation.h
rainman110: Agree

src/core/operations/gt_executableoperation.h
rainman110: That’s a slightly different distinction. "requiresProject()" is not about project files.

It means that the execution requires a GTlab project to be instantiated in addition to the explicit operation data. How that project gets there is deliberately left open. It could be transferred beforehand, already be available at the execution site, or be provided in some other way.

An operation can also choose to put everything it needs into "createData()". In that case, no additional project is required.

Tasks are one example where a project will often be needed, but other operations  e.g. geometry generation or Intelligraph node execution may only need a much smaller set of detached input data.

So the distinction is really: does "execute()" need a GTlab project in addition to its explicit invocation data?

tests/unittests/core/test_gt_executableoperation.cpp
rainman110: Publish without args is just a placeholder since I have not decided about the types to publish in this pr. 

In the final design it needs something like key value pair, i e. the event to publish. In fact we will send QJsonObjects, as they can be serialized well without adding the complexity of gtobjects ( which would have been an alternative)

When publishing an event, it will be like emitting a signal, just a different API. This event will then be transported back to the client.

tests/unittests/core/test_gt_executableoperation.cpp
rainman110: Yes, exactly. This is mainly to test that cancellation can be requested from another thread while the operation is running.

For a separate process the request would later have to be transported to the worker first, but would end up setting the same cancellation state there.

src/core/operations/gt_executableoperation.h
mariusalexander: So would any operation that for example needs to call `currentProject()` to set `requiresProject = true`? (which we of course want to avoid in the future but just to make sure I understand)

tests/unittests/core/test_gt_executableoperation.cpp
mariusalexander: I see ✅

tests/unittests/core/test_gt_executableoperation.cpp
mariusalexander: Got it! Key-value pairs like JSON objects could be a good idea

mariusalexander:
@rainman110 Unless you have started integrating the changes, could you consider moving the source files (.h and .cpp) that are part of the "operations" interface to a separate folder (IMO not into `core/process_management/` but maybe into `core/operations/`. In my head `core/process_management/` deals with tasks and calculators only)? I think we could start to better organize the source code files and moving source files that logically tackle one feature should be grouped in the source tree too. What do you think?

src/core/operations/gt_operationexecutioncontext.h
rainman110: Good catch, i think it is a good candidate to pop some pimpls 😆

src/core/operations/gt_operationexecutioncontext.h
rainman110: Yes, lets change this to 

"The context passed to the GtExecutableOperation::execute call, including the input data"

mariusalexander: Sounds good!  Much clearer

rainman110:
Architecture documentation was clarified in 18d9b497. The Executable operations guide now introduces the motivation, the relation to GtTask, the operation lifecycle, and the three requiresProject/createData combinations in plain GTlab terms. Architecture decision 0001 was shortened and no longer documents concrete #1529 event-stream or worker-output details. The Sphinx documentation build passes.

tests/unittests/core/test_gt_executableoperation.cpp
rainman110: Okay for me

src/core/operations/gt_operationexecutioncontext.h
rainman110: Agree. At least the == operator. The other should follow automatically imho.

rainman110:
> @rainman110 Unless you have started integrating the changes, could you consider moving the source files (.h and .cpp) that are part of the "operations" interface to a separate folder (IMO not into `core/process_management/` but maybe into `core/operations/`. In my head `core/process_management/` deals with tasks and calculators only)? I think we could start to better organize the source code files and moving source files that logically tackle one feature should be grouped in the source tree too. What do you think?

Good idea.  i've just moved it.

docs/dev/architecture/executable_operations.rst
rainman110: I have completely reworked the architecture decomentation. I made it less ai slop, used more gtlab speech etc. Would be nice, if you could have a look.

mariusalexander, line 1: This is really nice to read!! 👍👍

rainman110:
@mariusalexander I think i have integrated all your comments and ideas. I also have improved docstrings and documentation and shorted the corresponding issue. I hope this is now in a good shape to merge.

rainman110:
@sourcery-ai review

docs/dev/modules/interfaces/operation.rst
```diff
@@ -0,0 +1,53 @@
...
+Register operation types
+------------------------
+
+``operations()`` returns the ``staticMetaObject`` of every executable
+operation class contributed by the module. Each declared type must derive from
+``GtExecutableOperation`` and provide an invokable constructor so normal
+``GtObjectFactory``/Memento reconstruction can create it.
+
+The module loader validates and registers these types in ``GtObjectFactory``.
+Do not add a second operation factory or a separate registration call. The
```
mariusalexander: I think this makes it a bit clearer. Feel free to apply or ignore.
```suggestion
Do not register the operation types manually in this or any other factory, as this may cause the module to fail to load. The
```

src/core/operations/gt_operationapplystatus.h
```diff
@@ -0,0 +1,52 @@
...
+     * @param errorMessage Description of the failure.
+     * @return A status indicating that the result was not applied.
+     */
+    static GtOperationApplyStatus failure(QString errorMessage);
+
+    /**
+     * @brief Returns whether the result was applied successfully.
+     * @return True if the result was applied.
+     */
+    bool succeeded() const noexcept;
```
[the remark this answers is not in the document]
rainman110: > **issue (bug_risk):** The public API introduces `GtOperationApplyStatus::succeeded()` and names the type `GtOperationApplyStatus`, but the requested contract is `GtOperationApplyResult` with `isSuccess()`. Callers implementing or consuming the documented API cannot compile against this implementation.
> 
> **Suggested fix:** Rename the type to `GtOperationApplyResult` and expose `isSuccess()` consistently in the header, implementation, and callers.

It has been renamed deliberatly. Where is the contract documented?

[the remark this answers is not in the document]
rainman110: > So, if `GtOperationApplyStatus::succeeded()` was an intentional design change, my comment was based on an outdated or inconsistent API note rather than a separately documented contract. Please update the API note—or link the authoritative issue/design document—so the intended public API is unambiguous.

Done

docs/dev/architecture/decisions/0001-executable-operations.rst
```diff
@@ -0,0 +1,94 @@
...
+
+The originating operation, an execution-local reconstructed operation, and the
+detached input/result are separate objects with separate lifetimes. Detached
+objects must not contain borrowed pointers into the originating project.
+
+``GtOperationExecutionContext`` contains invocation data, execution identity,
+cancellation, and event publication. It does not own project state. A project
+at the execution location remains available through ``GtExecutionContext``.
+
+Operation submission is asynchronous for the caller, while ``execute()`` stays
```
mariusalexander, line 48: Just to make sure, can createData be run on a different thread or is this not planned/intended as of yet? I am just thinking that copying the data may still take some time and could cause micro-stutters in the UI. But maybe I am getting ahead of things.

mariusalexander (approved):
You may apply or ignore the one suggestion. The other is just a follow-up question, no blocker. I really like the documentation now. Much easier to grasp. Thanks for the effort! LGTM

docs/dev/architecture/decisions/0001-executable-operations.rst
rainman110: Conceptually, it is more about the location (i.e. the client,or where the data is). This still could be run in a thread to avoid blocking the ui.

PR #1535, opened 2026-09-04
[remote execution] Add execution event stream and worker event output

### LINKED ISSUE #1529: [Operations] Add execution events for executable operations

rainman110:
Closes #1529

## Description

This PR builds on #1528.

It lets an executable operation report progress and other small events while `execute()` is running:

- Operations publish events through `GtOperationExecutionContext::events()`.
- `GtExecutionEventStream` adds the execution ID and a sequence number. All observers receive events in the same order, including events published from several threads or from inside an observer.
- `GtExecutionEventFileWriter` writes each event as one JSON line to `events.ndjson`. It keeps normal log output separate and reports file errors. If a write fails, it removes any incomplete event record.
- `GtStdioExecutionEventEncoder` is an optional fallback. It writes complete V1 protocol records to standard output and serializes writes from multiple encoders.

Events contain only an event type and optional JSON data. Operations cannot use the event stream to send `GtObject` or Memento data.

This PR does not add operation scheduling, worker orchestration, task integration, GUI handling, or remote infrastructure.

## How Has This Been Tested?

- Built `GTlabUnitTest` from the prepared `build/` directory.
- Ran 14 focused event tests. They cover event order, publication from several threads, publication from an observer, cross-thread Qt delivery, NDJSON output, file errors, incomplete writes, V1 records, terminal result and failure records, and concurrent standard-output writes.
- Ran Cppcheck for the changed event writer and standard-output encoder.
- Built the developer documentation with `cd docs/dev && make`.

## Checklist

- [ ] The changelog has been extended, if this PR contains important user-visible changes.
- [x] The new code complies with the GTlab style guide.
- [x] New interface methods and functions are exported via `EXPORT`. Non-interface functions are not exported.

## Summary by Sourcery

Enable executable operations to publish ordered JSON events and expose them through file-based or V1 standard-output worker channels.

New Features:
- Add JSON-based execution events with execution IDs and ordered sequence numbers for reporting progress during executable operations.
- Provide NDJSON file output for execution events and an optional V1 standard-output encoder for worker compatibility, including terminal result and failure records.

Bug Fixes:
- Ensure concurrent and reentrant event publication preserves a consistent order for all observers.
- Prevent incomplete event records from remaining in the event file after write failures.

Enhancements:
- Replace the temporary execution event sink with a thread-safe event stream exposed through the operation execution context.

Documentation:
- Update executable-operation architecture documentation to describe the execution event model and boundaries.

Tests:
- Add focused coverage for event ordering, concurrency, cross-thread delivery, NDJSON persistence and rollback, protocol encoding, terminal records, and concurrent standard-output writes.


rainman110:
@sourcery-ai please update your pool request description. It contains changes from another branch since I mistakenly have chosen the wrong base.

rainman110 (commented):
The overall direction is close, but the architecture discussion around #1529 has clarified a few boundaries that should be reflected before this PR is ready. The issue has been updated accordingly.

Two additional points that are not easy to attach to a changed line:

1. **Execution identity:** one execution must have exactly one ID. The future runtime-owned execution state is the source; the event stream is constructed from that ID, and `GtOperationExecutionContext::executionId()` should expose/delegate to the same value instead of keeping an independently generated second ID. Please make inconsistent context/stream IDs structurally impossible.
2. **Tests:** please add focused tests for queued/cross-thread Qt observation and for reentrant publication, proving that all observers see one consistent sequence order. Also add coverage for the file-backed event writer (one compact JSON line per event, no mixing with ordinary stdout/log output).

The stdio encoder may remain as a compatibility/fallback adapter, but the primary Worker Slice 1 event channel should now be the exclusive file-backed `events.ndjson`-style channel described in #1529.

src/core/gt_operationexecutioncontext.h
```diff
@@ -50,18 +51,32 @@ class GT_CORE_EXPORT GtCancellationToken
...
 /**
- * @brief Temporary transport-neutral boundary for execution-side observations.
+ * @brief Transport-neutral publication boundary for operation observations.
  *
- * This foundation intentionally does not define an event envelope, event type,
- * or payload contract. Those contracts are follow-up runtime work.
+ * An operation receives this boundary from GtOperationExecutionContext and
+ * publishes domain observations only through it. Implementations such as
+ * GtExecutionEventStream assign execution order and identity; operations must
+ * not write stdout, access a transport, or select a GUI observer themselves.
```
rainman110: Please remove `GtExecutionEventSink` as an interface. We do not need a second observer/publisher abstraction here. `GtOperationExecutionContext::events()` should refer directly to the concrete `GtExecutionEventStream`, which is both the operation-facing publisher and the Qt-observable event source. Writers/adapters then simply connect to `eventPublished(...)` via signals/slots. This is now explicit in #1528/#1529.

src/core/gt_executionevent.h
```diff
@@ -0,0 +1,87 @@
...
+public:
+    /**
+     * @brief Identifies the representation of the payload.
+     *
+     * Json stores a JSON value tree directly. MementoXmlBase64 stores detached
+     * UTF-8 Memento/XML bytes; the stdio V1 encoder represents these bytes with
+     * standard Base64.
+     */
+    enum class PayloadEncoding { Json, MementoXmlBase64 };
+
```
rainman110: Please simplify Worker Slice 1 events to **JSON payloads only** and remove Memento/GtObject payload support from the event model. Events are intended for small status/progress/domain observations; larger GTlab data should travel through the result path or a future artifact/data channel. In particular, `MementoXmlBase64` also leaks a textual wire encoding into the logical event model. #1529 has been updated accordingly.

src/core/gt_executioneventstream.cpp
```diff
@@ -0,0 +1,54 @@
...
+                             payload.toMemento().toByteArray()));
+}
+
+void GtExecutionEventStream::publish(GtExecutionEvent event)
+{
+    if (event.eventType().isEmpty()) {
+        throw std::invalid_argument("Execution event type must not be empty");
+    }
+
+    std::lock_guard<std::recursive_mutex> lock(m_mutex);
```
rainman110: Please do not hold a recursive mutex while emitting `eventPublished`. A direct observer can re-enter `publish()`, causing event N+1 to be emitted before every observer has finished receiving event N; arbitrary observer code also runs under the stream lock. Please use a small enqueue/drain (or equivalent non-reentrant dispatch) scheme: assign sequence + enqueue under the lock, let only one dispatcher drain the queue, and emit signals outside the lock. Reentrant/concurrent publishes should only append and must preserve one consistent observer order.

src/core/operations/gt_stdioexecutioneventencoder.cpp
```diff
@@ -0,0 +1,124 @@
...
+{
+}
+
+bool GtStdioExecutionEventEncoder::isTerminal() const noexcept
+{
+    std::lock_guard<std::mutex> lock(m_mutex);
+    return m_terminal;
+}
+
+void GtStdioExecutionEventEncoder::encodeEvent(GtExecutionEvent event)
```
rainman110: Please do not make stdout the authoritative event channel. We want a file-backed Qt observer as the first concrete worker event channel: connect it to `GtExecutionEventStream::eventPublished`, exclusively own an `events.ndjson`-style file, and append exactly one complete compact JSON record per event. Ordinary stdout/stderr/log output stays separate (locally we can view both with tools such as `multitail`). This avoids relying on impossible-to-guarantee interleaving semantics with arbitrary stdout writers. The stdio encoder may remain as a compatibility/fallback adapter, but its documentation must not claim stronger guarantees than it can provide.

src/core/CMakeLists.txt
```diff
@@ -259,6 +265,7 @@ target_include_directories(GTlabCore
 )
 
 set_target_properties(GTlabCore PROPERTIES 
+  AUTOMOC ON
```
rainman110: Please remove this. GTlab's standard CMake setup already enables AUTOMOC globally, so this is redundant and unrelated to #1529. The corresponding `Enable Qt automatic meta-object compilation` statement should also be removed from the PR description.

rainman110:
Please also update the PR description after the implementation changes. In particular:

- state explicitly that this is a **stacked PR on #1528**;
- remove all references to the `GtExecutionEventSink` interface;
- describe Worker Slice 1 events as JSON-only (no detached Memento event payloads);
- mention the file-backed `events.ndjson`-style channel as the primary event output and stdio only as an optional compatibility/fallback adapter, if retained;
- remove the `Enable Qt automatic meta-object compilation` item;
- avoid claiming general stdout/thread-safe record integrity against unrelated writers — the stdio adapter can only serialize its own protocol writes relative to each other;
- update the test summary to include the new reentrant/cross-thread ordering and file-writer coverage once added.

The current Sourcery summary is also stale with respect to these architecture clarifications, so please do not keep wording there that contradicts the updated #1529 contract.

rainman110:
Aktualisierung in c83c06dc:

- `GtExecutionEventStream` ist nun der konkrete, Qt-beobachtbare Publisher; der Kontext verwendet genau dessen ExecutionId.
- Logische Events enthalten ausschließlich JSON-Werte. Nebenläufige und reentrante Publikationen werden per Queue/Drain in Sequenzreihenfolge ausgeliefert, ohne Observer unter dem Mutex aufzurufen.
- `GtExecutionEventFileWriter` schreibt die primäre, exklusive NDJSON-Eventdatei; der stdio-Encoder bleibt ein kompatibler Fallback und erzeugt nur seine eigenen V1-Records.
- Öffentliche Klassen und ihre Nutzungs-/Lebensdauersemantik sind wieder dokumentiert.

Validiert mit `make -C build -j2 GTlabUnitTest` sowie `./build/GTlabUnitTest --gtest_filter=…`: 6/6 fokussierte Tests grün (Reentranz, Cross-Thread, NDJSON, exaktes V1-Encoding und nebenläufige stdio-Schreibvorgänge). Die bekannte Profiling-Ausgabe beim Testprozessende betrifft nur gcov-Artefakte und ändert das erfolgreiche Testergebnis nicht.

docs/dev/architecture/decisions/0001-executable-operations-one-shot-worker.rst
```diff
@@ -194,20 +195,31 @@ Worker Slice 1 implements only runtime completion. A future client-side
...
+A transport-neutral logical event envelope has these required fields:
 
 * ``executionId``: invocation identity;
 * ``sequence``: zero-based, strictly increasing unsigned per-execution order;
 * ``eventType``: non-empty domain event key; and
 * ``payload``: absent, a JSON value tree (null, boolean, number, string, list,
-  or string-keyed object), or a detached serializable ``GtObject`` Memento with
-  an identifying payload encoding.
+  or string-keyed object), or detached serializable ``GtObject`` Memento/XML.
+
```
rainman110: This still contradicts the updated #1529 and the implementation in this PR. Worker Slice 1 events are now deliberately **JSON-only**; Memento/GtObject payloads were removed so the event stream remains a small status/progress observation channel rather than a second object-transfer path. Please update this section accordingly. Also update the later worker/stdio section to make the file-backed exclusive `events.ndjson` channel the primary Slice-1 event output and describe stdio only as a compatibility/fallback adapter if it remains.

docs/dev/architecture/executable_operations.rst
```diff
@@ -89,6 +96,19 @@ Keep lifecycle boundaries separate:
...
+Event observation
+-----------------
+
+``GtExecutionEventStream`` is the operation-facing event publisher and local
+observable stream for one invocation. It assigns execution identity and strict
+per-execution sequence order and emits logical ``GtExecutionEvent`` values via
+Qt signals/slots. Writers and adapters connect as ordinary Qt observers; there
+is no separate ``GtExecutionEventSink`` interface.
+
+Logical event payloads remain transport-neutral. A detached object payload is
```
rainman110: Please align this guide with the updated decision as well: no detached Memento payloads in `GtExecutionEvent` for Worker Slice 1. Describe JSON-only event payloads and the file-backed `events.ndjson` writer as the first concrete worker event channel. The current text still documents the superseded Memento/stdout design.

src/core/operations/gt_executioneventfilewriter.cpp
```diff
@@ -0,0 +1,54 @@
...
+        {QStringLiteral("sequence"), static_cast<qint64>(event.sequence())},
+        {QStringLiteral("eventType"), event.eventType()},
+        {QStringLiteral("payload"), event.payload().isUndefined()
+                                    ? QJsonValue::Null
+                                    : event.payload()}
+    };
+
+    QByteArray line = QJsonDocument(record).toJson(QJsonDocument::Compact);
+    line += '\n';
+
```
rainman110, line 83: Please do not silently ignore event-channel I/O failures. Check the return value of `write()` (including short/failed writes) and `flush()`, and expose a failure state or Qt signal so the worker/integration layer can detect that structured events can no longer be persisted. Keep this small—no recovery framework is needed here—but silent event loss is not acceptable for the exclusive structured channel.

rainman110 (commented):
The refactor addresses the main architecture comments well. Please complete the remaining changes below before considering #1529 done. In addition to the inline comments, add a dedicated concurrent-publish test for `GtExecutionEventStream`: publish from multiple threads concurrently, then verify that all events are delivered exactly once with one gap-free sequence set `0..N-1` and the same execution id. The existing concurrent test exercises the stdio encoders, not concurrent `GtExecutionEventStream::publish()` itself.

rainman110 (commented):
Please also update the PR description before this leaves draft. It is still stale in several places: it says the `#1528 event-sink boundary` accepts JSON + detached Mementos, presents stdio V1 as the main event path, and lists Memento event-payload testing. The description should instead reflect the current implementation/issue: concrete `GtExecutionEventStream` (no sink interface), JSON-only event payloads, file-backed `events.ndjson` as the primary Slice-1 event channel, stdio only as compatibility/fallback if retained, and removal of the extra `AUTOMOC ON` change. Please also mention that this PR is stacked on #1528.

rainman110:
Architecture documentation revised in 642b7c6b. The Executable operations page is now a short, generic introduction covering motivation, execution in threads/processes/remote workers, the relation to GtTask, and the prepare-compute-apply lifecycle. Worker and stdio details were removed from that guide. Architecture decision 0001 was shortened to the durable technical constraints, renamed without the one-shot-worker wording, and linked from the guide without duplicating its introduction. The Sphinx documentation build passes.

rainman110:
The Events section in architecture decision 0001 was revised in 20bb9de1. It now starts with the purpose of events during a running operation and explains how a client can react before the final result is available. Publication and transport are described separately in plain language. The section also contains a two-line events.ndjson example matching the current file-writer fields, including an event without payload and a progress payload. The Sphinx documentation build passes.

rainman110:
Rebased onto the updated PR #1534 head 18d9b497. The final PR #1535 tree is unchanged from before the rebase; its concrete event-stream and events.ndjson documentation remains layered on top of the generic #1534 architecture text. Validation: Sphinx documentation build and 15 focused operation/event tests pass.

rainman110:
@sourcery-ai please review

src/core/operations/gt_stdioexecutioneventencoder.cpp
```diff
@@ -0,0 +1,131 @@
...
+    line += QJsonDocument(record).toJson(QJsonDocument::Compact);
+    line += '\n';
+
+    std::lock_guard<std::mutex> outputLock(s_outputMutex);
+    m_output.write(line.constData(), line.size());
+    m_output.flush();
+    if (!m_output)
+    {
+        return false;
+    }
```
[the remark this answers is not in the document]
rainman110: I added a fix. Please check again

rainman110:
@mariusalexander I'd like you to review the event system, that will be part of the remote execution api.

It contains the worker side of the event system, i.e. the event class, event writers (to stdout and to ndjson file) etc.

It is not yet connected to any worker code nor integrated into the gui client. 

I handoff the review now to you. Before, I made sure it is ready by

 - Reviewing it by myself
 - Closed all Agent messages
 - Made sure, the documentation and docstrings aren't too technical and verbose

I hope, this time it will be a smoother review for you than last time.

src/core/operations/gt_executioneventfilewriter.h
```diff
@@ -0,0 +1,90 @@
...
+     */
+    GtExecutionEventFileWriter(std::unique_ptr<QFile> file, QString filePath,
+                               QObject* parent = nullptr);
+
+private:
+    std::unique_ptr<QFile> m_file;
+    QString m_filePath;
+    mutable std::mutex m_mutex;
+    bool m_hasError{false};
+    QString m_errorString;
```
mariusalexander: Does it make sense to pop some pimples here too since this class is exported? (Im fine either way)

Consider "sorting" member by size
```suggestion
    std::unique_ptr<QFile> m_file;
    QString m_filePath;
    QString m_errorString;
    mutable std::mutex m_mutex;
    bool m_hasError{false};
```

src/core/operations/gt_executioneventfilewriter.cpp
```diff
@@ -0,0 +1,115 @@
...
+GtExecutionEventFileWriter::isOpen() const
+{
+    std::lock_guard<std::mutex> lock(m_mutex);
+    return m_file->isOpen();
+}
+
+QString const&
+GtExecutionEventFileWriter::filePath() const noexcept
+{
+    return m_filePath;
```
mariusalexander: Since `m_filePath` wont change it needs no guard, right? You could make `m_filePath` a "const" member to enforce this behavior.

src/core/operations/gt_executioneventstream.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+        if (m_dispatching)
+        {
+            return;
+        }
+
+        m_dispatching = true;
+    }
+
+    for (;;)
```
mariusalexander: Just my preference ;)
```suggestion
    while (true)
```

src/core/operations/gt_executioneventstream.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+#include <stdexcept>
+#include <utility>
+
+GtExecutionEventStream::GtExecutionEventStream(GtExecutionId executionId,
+                                               QObject* parent) :
+    QObject(parent),
+    m_executionId(std::move(executionId))
+{
+    qRegisterMetaType<GtExecutionEvent>();
```
mariusalexander: Maybe move this to `GtExecutionEvent`? But Im fine keeping it here too just to make sure the type is registered.

src/core/operations/gt_executioneventstream.h
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * @brief Publishes an event.
+     *
+     * @param eventType Non-empty event type.
+     * @param payload JSON payload. An undefined value means no payload.
+     *
+     * This function is safe to call from multiple threads. It assigns the next
+     * sequence number before notifying observers.
+     */
+    void publish(QString eventType, QJsonValue payload = {});
```
mariusalexander, line 55: Just a thought: Maybe we could think about predefined some event types but none come to mind currently. We can have a look once an operation actually use this interface.

src/core/operations/gt_operationexecutioncontext.cpp
```diff
@@ -17,18 +19,16 @@ struct GtCancellationToken::State
...
         data(data),
-        events(&events), cancellation(std::move(cancellation)),
-        executionId(std::move(executionId))
+        events(&events), cancellation(std::move(cancellation))
     {
     }
 
     GtObject* data;
-    GtExecutionEventSink* events;
+    GtExecutionEventStream* events;
```
mariusalexander, line 30: My preference: I like to use `QPointer`s to make sure to trip on a nullptr instead of a dangling pointer, but maybe its not needed here
```suggestion
    QPointer<GtObject> data;
    QPointer<GtExecutionEventStream> events;
```

src/core/operations/gt_stdioexecutioneventencoder.h
```diff
@@ -0,0 +1,96 @@
...
+     */
+    bool isTerminal() const noexcept;
+
+public slots:
+    /**
+     * @brief Writes an intermediate event record.
+     *
+     * Events for another execution, and events after completion, are ignored.
+     */
+    void encodeEvent(GtExecutionEvent const& event);
```
mariusalexander: `GtStdioExecutionEventEncoder` and `GtExecutionEventFileWriter ` are adapters for the event stream, right? 

In this class you have used `encodeEvent` in the other it is called `writeEvent`. Maybe It would make sense to create an abstract interface class with a virtual `writeEvent` function? 

Though maybe it is not needed since in the unittests you are "subscribing" to the Signal `eventPublished` and I would assume that the event who setups the event stream adapter knows the adapter it wants to setup and thus no inheritance may be needed. I'll let you decide if a common interface could be beneficial.

mariusalexander (changes requested):
No real blockers just some thoughts and minor suggestions. Very clean implementation.

src/core/operations/gt_executioneventfilewriter.h
rainman110: Yes, it makes sense to pimpl it as well.

src/core/operations/gt_executioneventstream.cpp
rainman110: Agree.

src/core/operations/gt_executioneventfilewriter.cpp
rainman110: What guard do you mean? what needs to be changed?

src/core/operations/gt_executioneventstream.h
rainman110: Definitely. We need to have some protocol later on. Right now it allows everything to be transmitted. This is on the one hand extensible, but too broad.

src/core/operations/gt_operationexecutioncontext.cpp
rainman110: Why not

src/core/operations/gt_stdioexecutioneventencoder.h
rainman110: I have deliberately avoided to use an interface to make it more flexible. But sure, renaming the method would not do any harm here.

src/core/operations/gt_executioneventfilewriter.cpp
mariusalexander: I meant the mutex lock-guard. I think you expect `m_filePath` to not change, that is why you have not used a mutex locker here. Thus, my thought was to make `m_filePath` in the header/member defintions const to enforce that `m_filePath` cannot be changed:

```cpp
    std::unique_ptr<QFile> m_file;
    const QString m_filePath;
    QString m_errorString;
    mutable std::mutex m_mutex;
    bool m_hasError{false};
```

src/core/operations/gt_stdioexecutioneventencoder.h
mariusalexander: Agree, renaming it is! :)

src/core/operations/gt_operationexecutioncontext.cpp
rainman110: I change my comment. The pointers are borrowed and expect to be valid (!= nullptr). Otherwise, methods like events() would result in a nullptr access. So there is no need to add a security check here, since it would violate the api contract.

src/core/operations/gt_executioneventstream.cpp
rainman110: Yes, i think it makes sense to register it there.

rainman110:
I address most of your comments. I discarded the QPointer idea though, see comment above.

src/core/operations/gt_operationexecutioncontext.cpp
mariusalexander: 👍

src/core/operations/gt_executionevent.cpp
```diff
@@ -18,6 +20,14 @@ GtExecutionEvent::GtExecutionEvent(GtExecutionId executionId, quint64 sequence,
...
     m_payload(std::move(payload))
 {
+    registerMetaType();
+}
+
+void
+GtExecutionEvent::registerMetaType()
+{
+    static std::once_flag registered;
+    std::call_once(registered, [] { qRegisterMetaType<GtExecutionEvent>(); });
```
mariusalexander, line 30: Did not now this exists. Awesome! 

I just new about immediately invoked lambdas: 
```cpp
static auto init_once = [](){ 
    qRegisterMetaType<GtExecutionEvent>(); 
    return 0; 
}()
```

mariusalexander (approved):
Sorry, forgot to redo the review... LGTM!

src/core/operations/gt_executionevent.cpp
rainman110: I think it's c++ 17. Anyway, a very practical utility and a bit safer than static initialization.

PR #1539, opened 2026-09-04
Fix concurrent module loading on 2.0.X

### LINKED ISSUE #1538: 🐛 [BUG] - Concurrent startup can incorrectly disable modules
## Summary

Starting multiple GTlab processes concurrently can incorrectly classify modules that are currently being loaded by another process as modules that crashed during a previous run. The affected process then skips those modules and dependent modules may fail to load.

The problem was reproduced with 20 concurrently started `GTlabConsole --help` processes. Nineteen of the twenty processes skipped one or more modules; only one loaded without a crash-marker warning.

## Expected Behaviour

Concurrent GTlab starts that share the same GTlab configuration must not suppress modules merely because another live process is currently loading them.

A module should only be suppressed on a later start when its own loading process actually terminated before it removed its crash marker.

The fix should preserve the existing module-loading and error-reporting behaviour otherwise. This issue is specifically about preventing concurrently running GTlab instances from interfering with each other's crash-marker state.

## Reproduction steps

```
# Remove stale crash markers.
GTlabConsole enable_modules --all

# Save the standalone reproduction script from the comment below as
# test_parallel_console_start.sh in the repository root, then make it executable.
chmod +x test_parallel_console_start.sh

# Start 20 console instances at once.
./test_parallel_console_start.sh 20

# Inspect the captured standard-error logs. The script prints its output directory.
rg -n "caused a crash in a previous run|Some modules failed to load" \
  /tmp/gtlab-parallel-console-*/stderr-*.log
```

## Screenshots

Not applicable.

## Logs

Example from one affected process:

```text
WARN  [14:37:48] [Core] Module 'Test Datamodel Interface' caused a crash in a previous run. Skipping module!
WARN  [14:37:48] [Core] Module 'Test Mdi Interface' caused a crash in a previous run. Skipping module!
ERROR [14:37:48] [Core] Cannot load module 'Test Process Interface' due to missing dependency 'Test Datamodel Interface'
ERROR [14:37:48] [Core] Cannot load module 'Test Mdi Interface Ext' due to missing dependency 'Test Mdi Interface'
ERROR [14:37:48] [Core] Some modules failed to load!
```

## Root cause

`CrashedModulesLog` uses the shared `build/GTlab.ini` key `loading_crashed` as a temporary crash marker. Before loading a module, a process appends the module path and synchronizes the INI file; it removes the entry after successful loading.

A second concurrently starting process reads this temporary entry from the same shared INI file and interprets it as a marker left by a previous crashed process. It then suppresses the module. The cached list is also written back during `pop()`, creating an unsynchronized read-modify-write race between processes.

## Suggested direction

Guard the complete critical module-loading section per shared `GTlab.ini` with an inter-process lock. The lock must cover the first read of `loading_crashed`, all temporary marker updates, and removal of the final marker. A competing process should wait rather than treating a live process's temporary marker as a previous crash.

This keeps the existing crash-recovery behaviour: if the lock owner terminates while loading, its marker remains and the next process can correctly suppress the module.

The change should remain minimal and must not alter unrelated module-loading semantics or introduce new requirements for command return values.

## GTlab environment

```text
GTlab Version: all (this bug was reported years ago and we finally found the cause)
Executable: GTlabConsole, GTlab
Test modules: Test Datamodel Interface, Test Mdi Interface, Test Module Interface, Test Process Interface, Test Mdi Interface Ext
```

## OS

Linux, Windows

rainman110:
## Description

Fixes #1538.

Serialize the shared `loading_crashed` transaction during actual module initialization with a `QLockFile` located next to `GTlab.ini`. Metadata discovery and dependency ordering remain lock-free; the lock covers only crash-marker access and plugin instantiation/initialization. A short linear retry avoids `QLockFile::lock()` backoff under concurrent startup.

The existing `GTlab.ini` format and crash-recovery behaviour remain unchanged. If a loading process terminates unexpectedly, its `loading_crashed` entry remains available to the next start.

## How Has This Been Tested?

- `cmake -S . -B build -DBUILD_TESTMODULES=ON && make -C build -j$(nproc)`
- `build/GTlabConsole enable_modules --all`
- Started ten `GTlabConsole --help` instances concurrently with the reproduction script from #1538.
  - All ten completed with exit code `0` (about 1.27 s wall-clock time).
  - No `caused a crash in a previous run`, `Cannot load module`, or `Some modules failed to load` entries occurred in captured stderr logs.

## Checklist:

- [ ] A test for the new functionality was added (not applicable; the regression is exercised through concurrent processes).

## Summary by Sourcery

Serialize concurrent module initialization to prevent false crash detection and incorrect module disabling.

Bug Fixes:
- Prevent modules from being incorrectly disabled when multiple GTlab instances start concurrently.
- Preserve crash-recovery behavior while avoiding false crash markers during concurrent module initialization.

Enhancements:
- Serialize crash-marker access and module initialization across concurrent GTlab processes while keeping metadata discovery and dependency ordering concurrent.

### DIFF
--- a/src/core/gt_moduleloader.cpp
+++ b/src/core/gt_moduleloader.cpp
@@ -194,6 +194,9 @@ getMatchedModuleIds(const QString& dependency,
 
 } // namespace
 
+QStringList getSortedModulesToLoad(const QStringList& modulesIdsToLoad,
+                                   const ModuleMetaMap& metaMap);
+
 class ModuleLoadingLock
 {
 public:
@@ -247,7 +250,7 @@ public:
      * @return
      */
     bool performLoading(GtModuleLoader& moduleLoader,
-                        const QStringList& modulesToLoad,
+                        const QStringList& sortedModuleIds,
                         const ModuleMetaMap &metaMap,
                         QStringList& failedModules);
     /**
@@ -566,6 +569,24 @@ ModuleMetaData loadModuleMeta(const QString& moduleFileName)
     return meta;
 }
 
+ModuleMetaMap
+filterCrashedModules(const ModuleMetaMap& metaData)
+{
+    const auto crashedModules = CrashedModulesLog().crashedModules();
+
+    return filterModules(metaData, [&](const ModuleMetaData& m){
+        if (crashedModules.contains(m.location()))
+        {
+            logWarnOnce(
+                QObject::tr("Module '%1' caused a crash in a previous run. "
+                            "Skipping module!").arg(m.moduleId())
+            );
+            return false;
+        }
+        return true;
+    });
+}
+
 ModuleMetaMap loadModuleMeta()
 {
     std::map<QString, ModuleMetaData> metaData;
@@ -589,21 +610,6 @@ ModuleMetaMap loadModuleMeta()
                      insertResult.first->second.location()));
         }
     }
-
-    const auto crashed_mods = CrashedModulesLog().crashedModules();
-
-    // Remove all modules, that have been crashed earlies
-    metaData = filterModules(metaData, [&](const ModuleMetaData& m){
-        if (crashed_mods.contains(m.location()))
-        {
-            logWarnOnce(
-                QObject::tr("Module '%1' caused a crash in a previous run. "
-                            "Skipping module!").arg(m.moduleId())
-            );
-            return false;
-        }
-        return true;
-    });
 
     // Remove all modules, that should be exluded
     const auto excludeList = getModulesToExclude();
@@ -632,13 +638,6 @@ GtModuleLoader::loadSingleModule(const QString& moduleLocation)
         gtError() << QObject::tr("Cannot load module. "
                                  "Module File '%1' does not exists.")
                          .arg(moduleLocation);
-        return false;
-    }
-
-    ModuleLoadingLock moduleLoadingLock;
-    if (!moduleLoadingLock.locked())
-    {
-        gtError() << QObject::tr("Cannot acquire the module loading lock.");
         return false;
     }
 
@@ -673,8 +672,39 @@ GtModuleLoader::loadSingleModule(const QString& moduleLocation)
         moduleMetaMap.insert(std::make_pair(moduleMeta.moduleId(), moduleMeta));
     }
 
+    const auto sortedModuleIds =
+        getSortedModulesToLoad(modulesToLoad, moduleMetaMap);
+
+    ModuleLoadingLock moduleLoadingLock;
+    if (!moduleLoadingLock.locked())
+    {
+        gtError() << QObject::tr("Cannot acquire the module loading lock.");
+        return false;
+    }
+
+    m_pimpl->m_metaData = filterCrashedModules(m_pimpl->m_metaData);
+    moduleMetaMap = m_pimpl->m_metaData;
+
+    const auto currentModuleIt = moduleMetaMap.find(moduleMeta.moduleId());
+    if (currentModuleIt != moduleMetaMap.end())
+    {
+        currentModuleIt->second = moduleMeta;
+    }
+    else
+    {
+        moduleMetaMap.insert(std::make_pair(moduleMeta.moduleId(), moduleMeta));
+    }
+
+    QStringList loadableModuleIds = sortedModuleIds;
+    loadableModuleIds.erase(
+        std::remove_if(loadableModuleIds.begin(), loadableModuleIds.end(),
+                       [&moduleMetaMap](const QString& moduleId){
+            return moduleMetaMap.find(moduleId) == moduleMetaMap.end();
+        }),
+        loadableModuleIds.end());
+
     QStringList failedModules;
-    if (!m_pimpl->performLoading(*this, modulesToLoad,
+    if (!m_pimpl->performLoading(*this, loadableModuleIds,
                                  moduleMetaMap, failedModules))
     {
         gtError().verbose() << QObject::tr("Some modules failed to load!");
@@ -688,6 +718,12 @@ GtModuleLoader::loadSingleModule(const QString& moduleLocation)
 void
 GtModuleLoader::load()
 {
+    m_pimpl->m_metaData = loadModuleMeta();
+
+    const auto allModulesIds = m_pimpl->getAllLoadableModuleIds();
+    const auto sortedModuleIds =
+        getSortedModulesToLoad(allModulesIds, m_pimpl->m_metaData);
+
     ModuleLoadingLock moduleLoadingLock;
     if (!moduleLoadingLock.locked())
     {
@@ -695,30 +731,29 @@ GtModuleLoader::load()
         return;
     }
 
-    m_pimpl->m_metaData = loadModuleMeta();
+    m_pimpl->m_metaData = filterCrashedModules(m_pimpl->m_metaData);
 
-    auto allModulesIds = m_pimpl->getAllLoadableModuleIds();
-    auto moduleMetaMap = m_pimpl->m_metaData;
+    QStringList loadableModuleIds = sortedModuleIds;
+    loadableModuleIds.erase(
+        std::remove_if(loadableModuleIds.begin(), loadableModuleIds.end(),
+                       [this](const QString& moduleId){
+            return m_pimpl->m_metaData.find(moduleId) ==
+                   m_pimpl->m_metaData.end();
+        }),
+        loadableModuleIds.end());
 
     QStringList failedModules;
-    if (!m_pimpl->performLoading(*this, allModulesIds,
-                                 moduleMetaMap, failedModules))
+    if (!m_pimpl->performLoading(*this, loadableModuleIds,
+                                 m_pimpl->m_metaData, failedModules))
     {
         gtError().verbose() << QObject::tr("Some modules failed to load!");
-        Impl::printDependencies(failedModules, moduleMetaMap);
+        Impl::printDependencies(failedModules, m_pimpl->m_metaData);
     }
 }
 
 QMap<QString, QString>
 GtModuleLoader::moduleEnvironmentVars()
 {
-    ModuleLoadingLock moduleLoadingLock;
-    if (!moduleLoadingLock.locked())
-    {
-        gtError() << QObject::tr("Cannot acquire the module loading lock.");
-        return {};
-    }
-
     auto moduleMeta = loadModuleMeta();
 
     QMap<QString, QString> retval;
@@ -1042,14 +1077,12 @@ getSortedModulesToLoad(const QStringList& modulesIdsToLoad,
 
 bool
 GtModuleLoader::Impl::performLoading(GtModuleLoader& moduleLoader,
-                                 const QStringList& moduleIds,
+                                 const QStringList& sortedModuleIds,
                                  const ModuleMetaMap& metaMap,
                                  QStringList& failedModules)
 {
     // initialize loading fail log
     CrashedModulesLog crashLog;
-
-    auto sortedModuleIds = getSortedModulesToLoad(moduleIds, metaMap);
 
     QStringList successfullyLoaded{};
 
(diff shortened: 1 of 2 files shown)


src/core/gt_moduleloader.cpp
```diff
@@ -606,6 +626,8 @@ GtModuleLoader::loadSingleModule(const QString& moduleLocation)
         return false;
     }
 
+    ModuleLoadingLock moduleLoadingLock;
```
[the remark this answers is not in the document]
rainman110: I agree with this finding. The lock is the correctness mechanism of this bug fix: if acquiring it fails and module loading simply continues, the original race is possible again. Please handle an unrecoverable `QLockFile::lock()` failure explicitly rather than proceeding as if synchronization had succeeded. This should stay narrowly scoped to the locking failure; no change to the general module-loading/CLI exit-code semantics is intended here.

src/core/gt_moduleloader.cpp
```diff
@@ -193,6 +194,25 @@ getMatchedModuleIds(const QString& dependency,
...
 } // namespace
 
+class ModuleLoadingLock
+{
+public:
+    ModuleLoadingLock() :
+        m_lockFile(GtCoreApplication::localApplicationIniFilePath() +
+                   QStringLiteral(".module-loading.lock"))
+    {
+        m_lockFile.lock();
```
rainman110: After reconsidering this in the context of the intended minimal 2.0.x bug fix, I don't think this should be a required change. The default 30 s stale timeout is acceptable here if we treat a module-loading critical section that remains blocked for that long as an abnormal startup condition. In that case, allowing recovery from a stale lock is preferable to potentially blocking all subsequent GTlab starts indefinitely.

So I withdraw the request to call `setStaleLockTime(0)`. The important correctness requirement for this PR remains that a normal concurrent startup must serialize access to the crash marker, and that an actual `QLockFile::lock()` failure must not silently fall back to unsynchronized module loading.

rainman110 (commented):
The overall approach is correct for #1538: serialize the existing crash-marker transaction across concurrently starting GTlab processes, without changing the existing module-loading semantics. For this 2.0.x bug-fix I would keep the scope exactly there. I consider the lock-acquisition handling and the stale-lock timeout below correctness requirements before merge; no CLI exit-code behavior change is requested.

rainman110 (commented):
Correction to my previous review: I no longer consider changing `QLockFile`'s default 30 s stale timeout a requirement for this 2.0.x bug fix. Given the intended startup semantics, a module-loading critical section that remains blocked for that long can reasonably be treated as abnormal, and allowing stale-lock recovery is preferable to risking an indefinitely blocked GTlab startup. The corresponding inline thread has been updated and resolved.

The remaining correctness point is the existing lock-acquisition finding: if `QLockFile::lock()` itself fails, module loading must not silently continue without synchronization. No broader module-loading or CLI behavior change is requested.

src/core/gt_moduleloader.cpp
```diff
@@ -647,6 +688,15 @@ GtModuleLoader::loadSingleModule(const QString& moduleLocation)
 void
 GtModuleLoader::load()
 {
+    ModuleLoadingLock moduleLoadingLock;
```
rainman110: This lock is currently acquired too early. It serializes `loadModuleMeta()`, dependency preparation, and the complete loading phase. With 10 concurrent `GTlabConsole` starts we measured ~26 s total versus ~0.6 s for one process. Please move the synchronization boundary down to the actual crash-sensitive loading path. `loadModuleMeta()` should be split so that pure plugin metadata discovery is lock-free, while reading/filtering `loading_crashed` happens only after the lock has been acquired and immediately before the protected module-loading phase. The lock must then remain held through the crash-marker push/pop and the actual `QPluginLoader::instance()` / module initialization path.

src/core/gt_moduleloader.cpp
```diff
@@ -662,6 +712,13 @@ GtModuleLoader::load()
 QMap<QString, QString>
 GtModuleLoader::moduleEnvironmentVars()
 {
+    ModuleLoadingLock moduleLoadingLock;
```
rainman110: `moduleEnvironmentVars()` should not take the module-loading lock. This path only needs module metadata/environment-variable information and does not instantiate or initialize modules. Serializing this scan adds another contended `QLockFile::lock()` acquisition to every startup and contributes directly to the observed lock convoy. Once metadata discovery is separated from crash-marker filtering, this function should be completely lock-free.

rainman110 (commented):
The synchronization should be narrowed to the actual crash-sensitive module loading path before merge. We measured ~0.6 s for a single `GTlabConsole` start, but ~26 s for 10 concurrent starts with this change. That is far beyond the ~6 s expected from pure serialization and is consistent with `QLockFile::lock()` contention/backoff creating a lock convoy.

For this 2.0.x bug fix, please keep metadata discovery, dependency preparation, and environment-variable extraction outside the inter-process lock. The lock should protect the existing `loading_crashed` protocol together with the code that can actually crash while loading/initializing a module (`QPluginLoader::instance()`, validation/registration, and `onLoad()`).

Concretely, `loadModuleMeta()` currently mixes two concerns: reading plugin metadata (safe and parallelizable) and filtering based on `CrashedModulesLog` (part of the crash-recovery protocol). Please split those concerns. Metadata collection should remain lock-free; reading/interpreting `loading_crashed` must happen inside the same critical section that subsequently writes the temporary marker and performs the dangerous module load. This preserves the correctness invariant without serializing unrelated startup work.

I would keep one lock around the complete actual module-loading phase rather than acquire/release a `QLockFile` per module, to avoid multiplying the contention/backoff problem. No change to CLI return-code semantics or other module-loading behavior is requested.

rainman110:
@rainman110 Addressed in da8583ce. The lock is now taken only around crash-marker access and actual module initialization; metadata discovery, dependency ordering, and environment-variable extraction remain lock-free.\n\nWith 10 parallel `GTlabConsole --help` starts, all processes completed with exit code `0` and no false crash-marker warnings. The measured wall-clock time dropped from about 26 s to about 1.27 s.

rainman110 (commented):
The new lock strategy and scope look good. Please keep the current `tryLock(0)` + 10 ms retry implementation and keep metadata discovery/dependency ordering lock-free. The remaining changes should be deliberately conservative for the `2.0.X` backport: prefer adding small private helpers over restructuring code that existed before this PR.

Please make the following two semantic corrections with minimal changes to existing functions.

### 1. Preserve the explicit `loadSingleModule()` semantics

`loadSingleModule()` is an explicit load/test operation. It must attempt to instantiate and initialize the requested module even if that module is already present in `loading_crashed`. The crash-marker suppression is intended to protect normal startup, not to prevent an explicit module-load test.

Please do **not** special-case this by restructuring `loadSingleModule()`. Instead, keep the existing `performLoading()` path and add a small internal policy controlling whether an already-existing crash marker is respected.

Suggested shape (exact naming is flexible, semantics are not):

```cpp
enum class PreviousCrashPolicy
{
    Skip,
    Ignore
};
```

Add this as the last `performLoading()` argument with a default of `PreviousCrashPolicy::Skip`, so the normal `load()` call can remain unchanged. Only `loadSingleModule()` should pass `PreviousCrashPolicy::Ignore` explicitly.

The existing check added by this PR should then effectively become:

```cpp
if (previousCrashPolicy == PreviousCrashPolicy::Skip &&
    crashedModules.contains(moduleMeta.location()))
{
    ... existing warning / previouslyCrashed handling ...
}
```

Important: even when an existing crash marker is ignored, `loadSingleModule()` must still use the normal `CrashedModulesLog::makeSnapshot()` around the actual load. In other words, an explicit test may ignore the *previous* marker, but the attempted load must still be crash-tracked. If that explicit load crashes, the marker must remain. If it fails normally, `loadSingleModule()` must return `false`; if it actually loads successfully, it may return `true`. Do not turn a skipped module into a successful result.

Also please do not add any new public API or change general CLI/exit-code semantics for this.

### 2. Preserve the old `moduleEnvironmentVars()` behaviour without reintroducing the large lock scope

Before this PR, `loadModuleMeta()` also filtered modules listed in `loading_crashed`, which meant environment variables from previously crashed/suppressed modules were not returned. Now that crash filtering was correctly removed from generic metadata discovery, `moduleEnvironmentVars()` would otherwise start returning environment variables from modules that normal startup will not load. That is an unintended behavioural change for this bug-fix release.

Please keep `loadModuleMeta()` as the current pure/lock-free metadata discovery function. Do **not** move crash filtering back into it.

Instead, add a small private helper in `gt_moduleloader.cpp`, preferably next to the metadata helpers, whose only purpose is to obtain metadata with the previous crash-filter semantics for read-only callers such as `moduleEnvironmentVars()`. The helper should:

1. call `loadModuleMeta()` **before** taking the lock;
2. acquire `ModuleLoadingLock`;
3. fail safely (log and return an empty result) if the lock cannot be acquired;
4. read `CrashedModulesLog().crashedModules()` while holding the lock;
5. filter those module locations from the already-discovered metadata;
6. return the filtered metadata and release the lock immediately.

Conceptually:

```cpp
ModuleMetaMap loadModuleMetaExcludingCrashed()
{
    auto metaData = loadModuleMeta();

    ModuleLoadingLock lock;
    if (!lock.locked())
    {
        gtError() << QObject::tr("Cannot acquire the module loading lock.");
        return {};
    }

    const auto crashedModules = CrashedModulesLog().crashedModules();
    return filterModules(metaData, [&](const ModuleMetaData& module) {
        return !crashedModules.contains(module.location());
    });
}
```

Then keep `moduleEnvironmentVars()` itself almost unchanged: ideally only replace its current `loadModuleMeta()` call with this helper. This is intentional: for a `2.0.X` bug fix we want the existing function body to remain easy to compare with the previous implementation.

Do **not** use this helper from `performLoading()`: actual module initialization still needs the lock to remain held across reading the crash state, `makeSnapshot()`, `QPluginLoader::instance()`, registration/checking and `onLoad()`. The helper is only for restoring the previous read-only filtering semantics without serializing metadata discovery.

### Scope constraints

Please keep the following current implementation choices unchanged unless required by the two corrections above:

- keep `ModuleLoadingLock` based on `QLockFile::tryLock(0)` with the short 10 ms retry;
- keep the existing stale-lock behaviour of `QLockFile`;
- keep metadata discovery lock-free;
- keep dependency ordering before acquiring the module-loading lock;
- keep one lock across the actual module-loading phase rather than locking once per module;
- keep the existing `GTlab.ini` / `loading_crashed` format;
- do not refactor unrelated module-loading code or change command return/exit semantics.

### Please re-test these cases

1. One `GTlabConsole --help` start remains around the current baseline.
2. Ten concurrent `GTlabConsole --help` starts remain close to the current ~1.2 s result and produce no false crash-marker warnings.
3. A module already present in `loading_crashed` is skipped during normal `load()`.
4. The same module is nevertheless actually attempted by `loadSingleModule()` and that function reports the real load result.
5. `moduleEnvironmentVars()` does not return environment variables from modules present in `loading_crashed`.

The objective is not to minimize the number of newly added lines. The objective is to minimize changes to previously existing `2.0.X` code: new small helpers are preferable if they let the old functions remain structurally unchanged and easy to reason about.

rainman110:
@rainman110 Addressed in 6ab7d113.

- `loadSingleModule()` now passes an internal `PreviousCrashPolicy::Ignore`, so it still attempts the requested module while retaining the normal crash snapshot around that attempt.
- `moduleEnvironmentVars()` now uses a small read-only helper that discovers metadata without a lock, briefly locks only to read and filter `loading_crashed`, then releases it.

Verified: normal startup skips a pre-marked module; explicit `load_module` instantiates and successfully loads that same module; ten concurrent `--help` starts still complete without false warnings in about 1.27 s.

rainman110:
@jensschmeink can you please review this? If you still have your auto opti setup, please try it out with this change!.

jensschmeink (approved):
Code LGTM and the basic function to avoid crashed modules is still working. Parallel execution problems could not be tested by me (no valid setup).

PR #1543, opened 2026-09-07
Update gitlab ci to 2.1 toolchain

### LINKED ISSUE #1542: [CHANGE] - Update gitlab pipeline to 2.1 toolchain usage
### Summary

We should switch to the 2.1 toolchain on master. This will use qt6 and the 2.1 devtools

### Impact on the codebase

-

### Current workarounds

_No response_

rainman110:
## Description

This PR changes the default build toolchain to the 2.1 images.
We keep 2.0 build jobs mainly to still support the Qt5 based Squish tests.

The 2.1 toolchain has a new cppcheck version, shows many more code issues. To not include too many changes here in this PR, I collected e current issues as a baseline and suppressed most of them. I have fixed critical issues like uninitialized variables, which have been exposed during cppcheck runs.

## How Has This Been Tested?

- CI
- locally In docker 

## Checklist:

## Summary by Sourcery

Move the CI pipeline to the 2.1 toolchain while preserving 2.0 compatibility coverage and stabilizing updated static analysis.

Bug Fixes:
- Initialize previously undefined state and reject null module upgrade routines to address issues identified by the updated static analysis.

Enhancements:
- Adopt the 2.1 toolchain as the default for Windows and Linux builds while retaining Linux 2.0 jobs for Qt5-based GUI tests.
- Update testing, coverage, packaging, and code-quality jobs to use the new toolchain configuration.
- Establish a cppcheck baseline and adapt analysis configuration for the newer cppcheck behavior.

Build:
- Update CI build and packaging configurations for the 2.1 toolchain and corresponding Windows compiler images.

CI:
- Retain dedicated 2.0 Linux build and test jobs for compatibility coverage while moving standard validation jobs to 2.1.

Chores:
- Make coverage checks informational in pull requests.


rainman110:
@mariusalexander do you have any opinions, how we should proceed with the new cppcheck findings? I suppressed them for now

src/gui/gt_extendedtaskdata.h
```diff
@@ -49,6 +49,9 @@ class GT_GUI_EXPORT GtExtendedTaskDataImpl : public GtTaskDataImpl
      */
     ~GtExtendedTaskDataImpl() override;
 
+    GtExtendedTaskDataImpl(const GtExtendedTaskDataImpl&) = delete;
+    GtExtendedTaskDataImpl& operator=(const GtExtendedTaskDataImpl&) = delete;
```
mariusalexander, line 53: Would it make sense to disallow copy (and move) in the base class(es) too? I am a bit rusty regarding inheritance and copy semantics, but I think it is common practice to disable copy by value to avoid "slicing"?

However, maybe the types are not used via the base classes and there is no risk in slicing, so I am unsure. But If we want to disallow copying, I think its best to make the copy (and move?) constructors/operators protected. That way they can be used by the derived classes internally.

src/gui/gt_extendedcalculatordata.h
```diff
@@ -48,6 +48,9 @@ class GT_GUI_EXPORT GtExtendedCalculatorDataImpl : public GtCalculatorDataImpl
      */
     ~GtExtendedCalculatorDataImpl() override;
 
+    GtExtendedCalculatorDataImpl(const GtExtendedCalculatorDataImpl&) = delete;
+    GtExtendedCalculatorDataImpl& operator=(const GtExtendedCalculatorDataImpl&) = delete;
```
mariusalexander, line 52: Same here regarding disallowing copy/move in base classes.

src/core/internal/gt_moduleupgrader.cpp
```diff
@@ -146,6 +146,13 @@ GtModuleUpgrader::registerModuleConverter(const QString &modId,
                                           GtVersionNumber target,
                                           ConverterFunction func)
 {
+    if (func == nullptr)
```
mariusalexander, line 149: Good catch :D

cppcheck.suppress
```diff
@@ -1,5 +1,7 @@
 # SPDX-FileCopyrightText: 2026 German Aerospace Center (DLR)
 # SPDX-License-Identifier: CC0-1.0
 
-passedByValue:*
 unmatchedSuppression:*
+missingInclude:*
+missingIncludeSystem:*
+normalCheckLevelMaxBranches:*
```
mariusalexander, line 7: what do these three settings do? They are disabled (suppressed), right?
Missing includes could be beneficial (though I guess the build would fail first).

mariusalexander (commented):
Better code checking is awesome, thanks for updating it! I agree, that we should not fix all the issues in this PR but our ultimate goal should be to slowly view & fix all suppressed issues. 

Though I have not found a way to view the warnings that are currently suppressed. Can you guide me?

src/gui/gt_extendedtaskdata.h
rainman110: I think yes. Better making it explicit

src/gui/gt_extendedcalculatordata.h
rainman110: Agree

cppcheck.suppress
rainman110: There were a ton of issues regarding missing excludes due to the way cppcheck was handling it. It was not checking, that an include was missing, but it could not find includes.

I decided that the compiler does a better job here and disabled it in cppcheck.

rainman110:
> Though I have not found a way to view the warnings that are currently suppressed. Can you guide me?

The suppressions are all stored in the baseline file 
cppcheck-baseline.suppress part of this pr. There are many shadowed variables warnings and others that are not particularly bad, but maybe not good practice.

And yes, we should fix these things from time to time, but not in this pr.

cppcheck.suppress
mariusalexander: Okay I see :+1:

rainman110:
@mariusalexander can you please check again?

mariusalexander (approved):
LGTM 👍

PR #1548, opened 2026-09-08
Add task memento diff and task state to run_task_from_memento

### LINKED ISSUE #1547: 💡 [REQUEST] - Add task memento diff and task state to run_task_from_memento
### Summary

Add additional output of task memento diff (the changed monitoring properties) and the task state, e.g. "FINISHED" as a json file.

Currently, report only the task state numeric and/or string of the enum GtProcessComponent::STATE to the json. Numeric value is more machine-relevant, but the enum key (e.g. "FINISHED") is more human-readable.
The json file can be extended/enriched with more information later.

### Improvements

Output gives a more complete picture of what has happened. Current diff of project (data) is already written to disk, additional output of two more files should not be an issue.

### Drawbacks

Makes the CLI command a little longer. Optionally default values for the filenames could be used.

### Current workarounds

_No response_

MattSchuff:
## Description
Adds optional commandline arguments for run_task_from_memento to output the task diff and task state
Task diff provides the changes in monitoring properties which occured during the task.
Task state is a JSON file which provides the resulting GtProcessComponent::STATE, as the numeric value and the string id of the enum value.
Implements #1547 

Actual names of the cli arguments are open for discussion
Actual content of the task state JSON is open for discussion.

## How Has This Been Tested?
Existing tests have been extended.

## Checklist:

## Summary by Sourcery

Extend run_task_from_memento to optionally write task changes and the resulting execution state alongside the project memento diff.

New Features:
- Add optional CLI outputs for task memento diffs and task execution state JSON.
- Expose the resulting task state through its enum string identifier.

Bug Fixes:
- Prevent project, task, and optional output files from overwriting one another.

Enhancements:
- Refactor memento execution option and path handling while preserving existing project diff generation.

Documentation:
- Update headless execution documentation with the additional memento output options.

Tests:
- Extend console execution tests to cover task diffs, task state output, output path validation, and finished, warning, and failed task states.

### DIFF
--- a/docs/user/headless/01_overview.rst
+++ b/docs/user/headless/01_overview.rst
@@ -23,9 +23,11 @@ process. The project input is the ``GtObjectGroup`` Memento produced by
        --project-memento project_memento.xml \
        --task-memento task.xml \
        --output-diff result.diff.xml \
+	   [--task-diff task.diff.xml] \
+	   [--task-state taskstate.json] \
        [--working-directory working_dir]
 
-The short option names are ``-p``, ``-t``, ``-o``, and ``-w``. If no working
+The short option names are ``-p``, ``-t``, ``-o``, ``-m``, ``-s`` and ``-w``. If no working
 directory is supplied, the directory containing the project Memento is used.
 During execution this directory is both the process working directory and the
 project path exposed by the execution context. Therefore existing calculators
--- a/src/batch/gt_consoleruntaskfrommemento.cpp
+++ b/src/batch/gt_consoleruntaskfrommemento.cpp
@@ -28,6 +28,9 @@
 #include <algorithm>
 #include <array>
 #include <memory>
+#include <qjsondocument.h>
+#include <qjsonobject.h>
+#include <qmetaobject.h>
 
 namespace
 {
@@ -39,6 +42,79 @@ namespace
         {
         }
     };
+
+    int
+    checkFiles(const QFileInfo projectFile,
+               const QFileInfo taskFile,
+               const QString outputProjectFile,
+               const QString outputTaskDiffFile,
+               const QString outputTaskStateFile,
+               const QString workingDirectory )
+    {
+
+        if (outputProjectFile == projectFile.absoluteFilePath() ||
+            outputProjectFile == taskFile.absoluteFilePath())
+        {
+            gtError() << QObject::tr(
+                "Output project diff must not overwrite an input Memento");
+            return 2;
+        }
+
+        if (outputTaskDiffFile == projectFile.absoluteFilePath() ||
+            outputTaskDiffFile == taskFile.absoluteFilePath())
+        {
+            gtError() << QObject::tr(
+                "Output task diff must not overwrite an input Memento");
+            return 2;
+        }
+
+        if (outputTaskStateFile == projectFile.absoluteFilePath() ||
+            outputTaskStateFile == taskFile.absoluteFilePath())
+        {
+            gtError() << QObject::tr(
+                "Output task state must not overwrite an input Memento");
+            return 2;
+        }
+
+
+        if (!outputTaskDiffFile.isEmpty() &&
+            QStringList{outputProjectFile, outputTaskStateFile}.contains(outputTaskDiffFile))
+        {
+            gtError() << QObject::tr(
+                "Output files cannot have the same name");
+            return 2;
+        }
+
+        if (!outputTaskStateFile.isEmpty() &&
+            QStringList{outputProjectFile, outputTaskDiffFile}.contains(outputTaskStateFile))
+        {
+            gtError() << QObject::tr(
+                "Output files cannot have the same name");
+            return 2;
+        }
+
+
+        if (!projectFile.isFile())
+        {
+            gtError() << QObject::tr("Project Memento does not exist: %1")
+            .arg(projectFile.absoluteFilePath());
+            return 3;
+        }
+        if (!taskFile.isFile())
+        {
+            gtError() << QObject::tr("Task Memento does not exist: %1")
+            .arg(taskFile.absoluteFilePath());
+            return 3;
+        }
+        if (!QFileInfo(workingDirectory).isDir())
+        {
+            gtError() << QObject::tr("Working directory does not exist: %1")
+            .arg(workingDirectory);
+            return 3;
+        }
+
+        return 0;
+    }
 
     std::unique_ptr<GtObjectGroup> restoreProjectData(QString const& fileName)
     {
@@ -161,6 +237,41 @@ namespace
         return true;
     }
 
+    bool writeTaskState(QString const& fileName, QJsonObject const& taskstate)
+    {
+        if (fileName.isEmpty())
+        {
+            return true;
+        }
+
+        QJsonDocument statusOutputDoc(taskstate);
+
+        QSaveFile file(fileName);
+        if (!file.open(QIODevice::WriteOnly))
+        {
+            gtError() << QObject::tr("Cannot write output task state json '%1': %2")
+                             .arg(fileName, file.errorString());
+            return false;
+        }
+
+        if (file.write(statusOutputDoc.toJson(QJsonDocument::Indented)) == -1)
+        {
+            gtError() << QObject::tr("Cannot write complete output task state json '%1': %2")
+                             .arg(fileName, file.errorString());
+            file.cancelWriting();
+            return false;
+        }
+
+        if (!file.commit())
+        {
+            gtError() << QObject::tr("Cannot publish output task state json '%1': %2")
+                             .arg(fileName, file.errorString());
+            return false;
+        }
+
+        return true;
+    }
+
     bool removeExistingOutput(QString const& fileName)
     {
         QFile outputFile(fileName);
@@ -186,11 +297,16 @@ gt::console::runTaskFromMementoOptions()
 {
     return {{{"project-memento", "p"}, "Serialized project-data Memento path"},
             {{"task-memento", "t"}, "Serialized task Memento path"},
-            {{"output-diff", "o"}, "Output Memento-Diff path"},
+            {{"output-diff", "o"}, "Output Project Memento-Diff path"},
+            {{"task-diff", "m"}, "Output Task Memento-Diff path"},
+            {{"task-state", "s"}, "Output Task State path"},
             {{"working-directory", "w"},
              "Execution working directory (defaults to project Memento "
              "directory)"}};
 }
+
+
+
 
 int
 gt::console::runTaskFromMemento(QStringList const& args)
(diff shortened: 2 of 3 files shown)


MattSchuff:
Docs have been extended. Original PR for introducing run_task_from_memento is not in changelog atm? Changelog left untouched

src/batch/gt_consoleruntaskfrommemento.cpp
```diff
@@ -40,6 +43,79 @@ namespace
...
+            outputProjectFile == taskFile.absoluteFilePath())
+        {
+            gtError() << QObject::tr(
+                "Output project diff must not overwrite an input Memento");
+            return 2;
+        }
+
+        if (outputTaskDiffFile == projectFile.absoluteFilePath() ||
+            outputTaskDiffFile == taskFile.absoluteFilePath())
+        {
```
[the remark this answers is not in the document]
MattSchuff: @sourcery-ai is addressed in [f92c2df](https://github.com/dlr-gtlab/gtlab-core/pull/1548/commits/f92c2df5b3459cb1582727f80f9f7120a1170c69)

src/batch/gt_consoleruntaskfrommemento.cpp
```diff
@@ -28,6 +28,9 @@
 #include <algorithm>
 #include <array>
 #include <memory>
+#include <qjsondocument.h>
+#include <qjsonobject.h>
+#include <qmetaobject.h>
```
mariusalexander: We normally use this style of inlcudes for Qt Classes. Also we sort the includes roughly in this order:
1. files from this module
2. files from gtlab
3. Qt/other thirdparty libraries
4. c++ std 

But no worries, just a hint for future work :) (and we sometimes forget this too)

```suggestion
#include <QJsonDocument>
#include <QJsonObject>
#include <QMetaObject>

#include <algorithm>
#include <array>
#include <memory>
```

docs/user/headless/01_overview.rst
```diff
@@ -23,9 +23,11 @@ process. The project input is the ``GtObjectGroup`` Memento produced by
        --project-memento project_memento.xml \
        --task-memento task.xml \
        --output-diff result.diff.xml \
+	   [--task-diff task.diff.xml] \
+	   [--task-state taskstate.json] \
        [--working-directory working_dir]
 
-The short option names are ``-p``, ``-t``, ``-o``, and ``-w``. If no working
+The short option names are ``-p``, ``-t``, ``-o``, ``-m``, ``-s`` and ``-w``. If no working
```
mariusalexander, line 30: We may want to revise the output options. I find `-m` and `-s` not quite intuitive and I can't think of another shortcut option. I would think a single `-o` option for the output *directory* would fit better (ideally I'd like to use `-op`, `-ot`, and `-os` for output project, task and state but I think that is not possible/standard behavior.

src/batch/gt_consoleruntaskfrommemento.cpp
```diff
@@ -40,6 +43,79 @@ namespace
...
+    checkFiles(const QFileInfo projectFile,
+               const QFileInfo taskFile,
+               const QString outputProjectFile,
+               const QString outputTaskDiffFile,
+               const QString outputTaskStateFile,
+               const QString workingDirectory )
+    {
+
+        if (outputProjectFile == projectFile.absoluteFilePath() ||
+            outputProjectFile == taskFile.absoluteFilePath())
```
mariusalexander: Maybe it is better to also normalize the outputs paths too using  `QFileInfo::canonicalFilePath()` (resolves symlinks) or just `QFileInfo::absoluteFilePath()` (does not resolve symlinks)

Further, I'd like to condense the logic here as well, as these checks are quite similar. Something along the lines of:

```cpp
// may need to check what happens if fileinfo is called with an empty filepath
QString normalize(QFileInfo const& fileInfo) { return fileInfo..canonicalFilePath(); } 

...

// normalize outputs and input paths
auto outputPaths = { normalize(outputProjectFile), normalize(outputTaskDiffFile), normalize(outputTaskStateFile) };
auto inputPaths = { normalize(projectFile), normalize(taskFile) };

for (QString const& outputPath : outputPaths)
{
    // check if output would overwrite an input file
    for (QString const& inputPath : inputPaths)
    {
        if (outputPath == inputPath)
        {
            gtError() << QObject::tr("Output file must not overwrite the input file '%1'!")
                            .arg(inputPath);
            return 2;
        }
    }

    // check if output would overwrite another output file
    auto count = std::accumulate(outputPaths.begin(), outputPaths.end(), unsigned{0}, [&outputPath ](unsigned count, QString const& nextOutput){ 
        return count + unsigned{outputPath == nextOutput};
    });
    if (!outputPath.isEmpty() && count > 1)
    {
        gtError() << QObject::tr("Output files must not overwrite each other (file: '%1')!")
                        .arg(outputPath);
        return 2;
    }
}
...
```

src/batch/gt_consoleruntaskfrommemento.cpp
```diff
@@ -186,12 +297,17 @@ gt::console::runTaskFromMementoOptions()
 {
     return {{{"project-memento", "p"}, "Serialized project-data Memento path"},
```
mariusalexander: I know this was there before and I should probably have noted it in the previous PR, but I'd like to reduce the redundant definitions of the arguments in this function and in the main "run" method. Maybe by using a list of QCommandLineOptions that is defined somewhere else. That way arguments and their text must only be updated once.

src/batch/gt_consoleruntaskfrommemento.cpp
```diff
@@ -186,12 +297,17 @@ gt::console::runTaskFromMementoOptions()
...
             {{"task-memento", "t"}, "Serialized task Memento path"},
-            {{"output-diff", "o"}, "Output Memento-Diff path"},
+            {{"output-diff", "o"}, "Output Project Memento-Diff path"},
+            {{"task-diff", "m"}, "Output Task Memento-Diff path"},
+            {{"task-state", "s"}, "Output Task State path"},
```
mariusalexander:
```suggestion
            {{"task-diff", "m"}, "Optional: Output Task Memento-Diff path"},
            {{"task-state", "s"}, "Optional: Output Task State path"},
```

src/batch/gt_consoleruntaskfrommemento.cpp
```diff
@@ -250,40 +372,36 @@ gt::console::runTaskFromMemento(QStringList const& args)
...
     {
-        gtError() << QObject::tr(
-            "Output diff must not overwrite an input Memento");
-        return 2;
+        outputTaskDiffFile0 = QFileInfo(parser.value(taskdiffOption)).absoluteFilePath();
     }
+    const QString outputTaskDiffFile = outputTaskDiffFile0;
 
-    if (!projectFile.isFile())
+    QString outputTaskStateFile0 = "";
```
mariusalexander: I'd just use `outputTaskStateFile` here directly instead of the `outputTaskStateFile0` helper

src/batch/gt_consoleruntaskfrommemento.cpp
```diff
@@ -335,6 +463,18 @@ gt::console::runTaskFromMemento(QStringList const& args)
     }
 
     const auto taskState = executionTask->currentState();
+
+    taskStateOutput["taskStateNumeric"] = static_cast<int>(taskState);
+    taskStateOutput["taskStateString"] =  QString::fromLatin1(
+        QMetaEnum::fromType<GtProcessComponent::STATE>().valueToKey(taskState)
+        );
```
mariusalexander: I'd use just one here

```suggestion
    taskStateOutput["taskState"] =  QString::fromLatin1(
        QMetaEnum::fromType<GtProcessComponent::STATE>().valueToKey(taskState)
    );
```

mariusalexander (commented):
Thanks for your work! 
A few ideas that came to my mind. I'd like to know @rainman110 opinion about these suggestions :)

mariusalexander:
> Original PR for introducing run_task_from_memento is not in changelog atm? Changelog left untouched

Good catch. I think it would be beneficial to add a short notice to the Chagelog.

docs/user/headless/01_overview.rst
MattSchuff: Agreed. I am also not totally happy about the naming of options. As mentioned in the PR at the top, naming should be discussed.

src/batch/gt_consoleruntaskfrommemento.cpp
MattSchuff: Good point. Agreed as a general idea. Question: do we want input and output files as a positional list (which file is which is purely defined by position), or should we put them in a struct (my suggestion for improvement here)?

src/batch/gt_consoleruntaskfrommemento.cpp
MattSchuff: Good point. I was a little lazy here, but would agree to put the QCommandLineParser into a seperate function to be called by the help option and the actual run.

src/batch/gt_consoleruntaskfrommemento.cpp
MattSchuff: I wanted to use the "const QString" style from above. The optional argument path should only by normalized to absolutePath when set, I'd change it to this:
`const QString outputTaskDiffFile = parser.value(taskdiffOption).isEmpty() ? QString() : QFileInfo(parser.value(taskdiffOption)).absoluteFilePath();`

However, maybe this becomes obsolete when putting all paths through the normalize() function suggested above.

src/batch/gt_consoleruntaskfrommemento.cpp
MattSchuff: done

src/batch/gt_consoleruntaskfrommemento.cpp
MattSchuff: obsolete, refactored whole section

src/batch/gt_consoleruntaskfrommemento.cpp
MattSchuff: refactored, see commit

src/batch/gt_consoleruntaskfrommemento.cpp
MattSchuff: added to refactoring commandline arguments section

MattSchuff:
@sourcery-ai please redo your review for latest commit

src/batch/gt_consoleruntaskfrommemento.cpp
```diff
@@ -335,6 +487,17 @@ gt::console::runTaskFromMemento(QStringList const& args)
     }
 
     const auto taskState = executionTask->currentState();
+
+    taskStateOutput["taskState"] =  QString::fromLatin1(
+        QMetaEnum::fromType<GtProcessComponent::STATE>().valueToKey(taskState)
+    );
```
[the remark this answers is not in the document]
MattSchuff: @sourcery-ai We revoked the idea to output the numeric value for now.

rainman110 (commented):
We are currently working on the generic worker/event architecture (#1529, #1530), which already provides an alternative path for task/calculator state changes and execution events. Both will also be written to files, but with a different format/protocol. Therefore, I would prefer not to introduce --task-state or --task-diff as additional file-based interfaces here.

Please defer both outputs to the task integration work, so they can use the shared result/event mechanisms instead of creating a parallel protocol. Feel free to give me a call and we can quickly align on the intended direction.

rainman110 (approved):
Even though it might not be the final design, let's merge this

src/batch/gt_consoleruntaskfrommemento.cpp
```diff
@@ -292,17 +349,18 @@ namespace
...
-             "Execution working directory (defaults to project Memento "
-             "directory)"}};
+    MementoExecutionOptions options;
+    QList<GtCommandLineOption> theList;
+
+    for(auto& option: options.list())
+    {
+        theList.append(GtCommandLineOption{option.names(), option.description()});
+    }
+    return theList;
```
mariusalexander, line 363: Won't that work too?
```suggestion
    MementoExecutionOptions options;
    return options.list();
```

mariusalexander:
Otherwise Im fine merging the changes :)

src/batch/gt_consoleruntaskfrommemento.cpp
MattSchuff: GtCommandLineOption (the returned list item by the function) is not derived from QCommandLineOption (the list item of `options`).
Technically it appears to be working, but I don't like these implicit conversions (for various reasons, mainly because bug-prone).
Bigger question to be discussed elsewhere: why does GtCommandLineOption exist and why is it not derived from QCommandLineOption. I found an inconsistent usage throughout other code.

mariusalexander: Thanks for the info, Ill create a new issue to discuss this. Was not aware of that.

PR #1560, opened 2026-09-18
Add a mapping of original to new UUID for copy function of GtObject

### LINKED ISSUE #1556: 💡 [REQUEST] - Introduce a reliable way to get a mapping of old and new UUIDs when copying a GtObject and its children
### Summary

When copying a tree of GtObject, there is currently no reliable way to tell which object is the copy of another (mapping of UUID from existing to new or vice versa).
This is bothersome when some operations have to be performed "on top", e.g. transferring property connections.

New feature should allow to get a mapping or similar feature of the copied objects' UUIDs.

### Improvements

Reliable way to tell which object is the copy of another.

### Drawbacks

Maybe API/ABI changes, depending on implementation.

### Current workarounds

Copying a GtTask uses a way by searching for the objectName and other features to transfer the property connections. But this is is only implicit and might be ambiguous when two calculators have the same name. It is also buggy when using and not reliable.

MattSchuff:
## Description
GtObject::copy() gets an overload function to track UUIDs when copying.
The overloaded function receives a new type GtObjectUUIDMap which is filled with pairs of existing object's UUID mapped to the value of the copied object's UUID.
Internally GtObjectIO receives a private function toMementoImpl() which receives the uuid map. The public API toMemento() calls this private function with nullptr for the mapping.

GtObjectIO has a friend class GtObject so that GtObject can now directly call the new private function. This avoids routing the GtObject::copy() through GtObject::toMemento() which created a temporary GtObjectIO. The temporary GtObjectIO is now created inside the copy function directly.

Resolves issue #1556 

## How Has This Been Tested?
Unit tests have been added.

## Checklist:

## Summary by Sourcery

Enable GtObject copies to expose the correspondence between original and copied object UUIDs.

New Features:
- Add an overload of GtObject::copy() that records mappings from original object UUIDs to the UUIDs assigned to copied objects and descendants.
- Introduce GtObjectUUIDMap for querying and managing copied-object UUID mappings.

Enhancements:
- Refactor object memento generation so copying can collect UUID mappings while preserving existing serialization behavior.

Documentation:
- Document the UUID-aware object copying behavior and mapping usage.

Tests:
- Add unit tests covering UUID map operations and mapped copies of nested object trees.


MattSchuff:
cppcheck fails for code that was not touched with this PR

src/dataprocessor/gt_object.h
```diff
@@ -193,6 +199,21 @@ class GT_DATAMODEL_EXPORT GtObject : public QObject
...
+     * Copies the object and its children, recording the new UUIDs.
+     * @param uuidMap Maps original UUIDs to the new UUIDs of the copied objects,
+     * including this object and all copied descendants. Existing entries are
+     * replaced. The map is empty if operation fails. Original UUIDs must be
+     * unique in the tree.
+     * @return a copy of the object.
+     *
+     * Use copy->getObjectByUuid(uuidMap.value(original->uuid()) to and
+     * original object's copy. Property references (e.g. Object Link) are
+     * copied unchanged.
```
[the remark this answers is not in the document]
MattSchuff: done in [d94160c](https://github.com/dlr-gtlab/gtlab-core/pull/1560/commits/d94160ce50734a0d5db0189d17f086194c3e1334)

src/dataprocessor/gt_object.h
```diff
@@ -161,9 +163,13 @@ class GT_DATAMODEL_EXPORT GtObject : public QObject
     /**
      * @brief Creates a memento of the internal object state.
      * @param clone
+     * @param uuidMap Maps original UUIDs to the new UUIDs of the copied objects,
+     * including this object and all copied descendants. Existing entries are
+     * replaced. The map is empty if operation fails. Original UUIDs must be
+     * unique in the tree.
```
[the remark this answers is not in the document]
MattSchuff: done in [d94160c](https://github.com/dlr-gtlab/gtlab-core/pull/1560/commits/d94160ce50734a0d5db0189d17f086194c3e1334)

src/dataprocessor/gt_objectio.cpp
```diff
@@ -223,11 +223,16 @@ GtObjectIO::setFactory(GtAbstractObjectFactory* factory)
...
+}
+
+GtObjectMemento
+GtObjectIO::toMemento(const GtObject* o, bool clone, QHash<QString, QString>* uuidMap)
 {
     // global object element
     GtObjectMemento memento;
 
-
     // class name
```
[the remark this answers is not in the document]
MattSchuff: done in [d94160c](https://github.com/dlr-gtlab/gtlab-core/pull/1560/commits/d94160ce50734a0d5db0189d17f086194c3e1334)

rainman110:
I think the general idea is fine, but I would like to clean up the API a bit before merging this.

What I would prefer:

 - Please introduce a small strong type for the UUID mapping instead of exposing a raw `QHash<QString, QString>`. The hash is an implementation detail and from the type alone it is not really clear what is mapped in which direction. A small `GtObjectUUIDMap` would make the API much more obvious.
 - Keep the existing `GtObject::copy()` as-is and add an overload like `copy(GtObjectUUIDMap& mapping)`. So no pointer/nullptr in the public API. If I don't want a mapping, I call `copy()`. If I want one, I pass the mapping.
 - I would try to keep the existing public `toMemento()` / `GtObjectMemento` API unchanged. For me the UUID tracking is an implementation detail of `copy()`, not something that should leak through several public APIs.
 - Internally it is totally fine to use a nullable pointer to the mapping. The nullptr idea actually came from me originally, but I meant it for exactly this purpose: if no mapping was requested, we should not build one just to throw it away afterwards.

I still think collecting the mapping where the new UUID is actually generated is the right approach. I just don't think the mechanism for doing that should become public API.

One option could be to introduce an internal/private `toMementoImpl(...)` which additionally takes the mapping pointer. The existing public `toMemento(...)` can then simply call this implementation with nullptr, while the copy implementation can use the internal variant with the mapping.

Something along these lines conceptually:

```
toMemento(...)
{
    return toMementoImpl(..., nullptr);
}

toMementoImpl(..., GtObjectCopyUuidMapping* mapping)
{
    // create UUID and update mapping if requested
}
```

So basically: keep the current public API as much as possible, expose the mapping only at `GtObject::copy()`, and keep the actual tracking mechanism under the hood.

rainman110 (changes requested):
@MattSchuff Please see my comments above.

In addition, don't change copyright years  of existing files. They should be kept at the date where the copyright was initially stated.

MattSchuff:
@rainman110 

All done.

Added `GtObjectUUIDMap` as a new type.

`GtObjectIO ` now has `GtObject` as a friend class.

Implemented `GtObjectIO::toMementoImpl(...)`.

Made the `copyCloneHelper` function now a private function of `GtObject`. That was necessary to call the private function of GtObjectIO.
Eliminated the use of `GtObjectMemento` constructor in `copyClone()` (former `copyCloneHelper()`), instead using directly the GtObjectIO. This way `GtObjectMemento` API does not need to know about `GtObjectUUIDMap` altogether.

MattSchuff:
@sourcery-ai review

src/dataprocessor/gt_objectuuidmap.h
```diff
@@ -0,0 +1,40 @@
...
+
+#include "gt_datamodel_exports.h"
+
+#include <QString>
+#include <QHash>
+
+class GT_DATAMODEL_EXPORT GtObjectUUIDMap
+{
+public:
+    QString newUuid(const QString& originalUuid) const;
```
rainman110: Maybe rename to `copiedUuid`

src/dataprocessor/gt_objectuuidmap.h
```diff
@@ -0,0 +1,40 @@
...
+
+#ifndef GT_OBJECTUUIDMAP_H
+#define GT_OBJECTUUIDMAP_H
+
+#include "gt_datamodel_exports.h"
+
+#include <QString>
+#include <QHash>
+
+class GT_DATAMODEL_EXPORT GtObjectUUIDMap
```
rainman110, line 24: Please add docstrings to the class (what does it do, what is the purpose) and to all methods

src/dataprocessor/gt_objectuuidmap.h
```diff
@@ -0,0 +1,40 @@
+/* GTlab - Gas Turbine laboratory
+ *
+ * SPDX-License-Identifier: MPL-2.0+
+ * SPDX-FileCopyrightText: 2026 German Aerospace Center (DLR)
+ * Source File: gt_objectuuidmap.h
```
rainman110: Remove line Source File and Created On

This is all in the git, we don't need to write it here anymore.

tests/unittests/datamodel/test_copy_gtobject_with_uuidmap.cpp
```diff
@@ -0,0 +1,265 @@
+/* GTlab - Gas Turbine laboratory
+ *
+ * SPDX-License-Identifier: MPL-2.0+
+ * SPDX-FileCopyrightText: 2026 German Aerospace Center (DLR)
+ * Source File: test_copy_gt_object_with_uuidmap.cpp
```
rainman110: Remove Line Source File and Created On

src/dataprocessor/gt_object.h
```diff
@@ -13,6 +13,7 @@
 
 #include "gt_datamodel_exports.h"
 #include "gt_typetraits.h"
+#include "gt_objectuuidmap.h"
```
rainman110: Prefer forward declaration of GtObjectUUIDMap to include.

src/dataprocessor/gt_object.h
```diff
@@ -798,6 +814,23 @@ protected slots:
...
+     * @brief Private backend function for serializing an object
+     * @param clone If true, UUIDs are preserved. If false, new UUIDs are
+     * generated.
+     * @param uuidMap Maps original UUIDs to the new UUIDs of the copied objects,
+     * including this object and all copied descendants. Existing entries are
+     * replaced. The map is empty if operation fails. Original UUIDs must be
+     * unique in the tree.
+     * @return
+     */
+   // GtObjectMemento toMementoImpl(bool clone = true, GtObjectUUIDMap* uuidMap = nullptr) const;
```
rainman110: There is still this outcommented code block

src/dataprocessor/gt_object.cpp
```diff
@@ -316,7 +332,7 @@ GtObject::setUuid(const QString& val)
 void
 GtObject::newUuid(bool renewChildUUIDs)
 {
-    pimpl->uuid = QUuid::createUuid().toString();
+    setUuid(QUuid::createUuid().toString());
```
rainman110: This change is unrelated, right? so think of reverting it.

tests/unittests/datamodel/test_copy_gtobject_with_uuidmap.cpp
```diff
@@ -0,0 +1,265 @@
...
+                }
+            }
+
+        }
+    };
+
+    GtObjectGroup* objgroup;
+};
+
+TEST_F(TestCopyGtObjectWithUuidMapping, testFillingAndClearingOfTestGroup)
```
rainman110: The tests look a bit too large for what we actually need to verify here.

I would remove the tests that mainly test the test setup itself (`testFillingAndClearingOfTestGroup`, `testAppendingToTestGroup`) and keep the tests focused on the actual copy behavior.

For me the relevant cases are basically:

* `copy()` still creates new UUIDs for the object and its children.
* `copy(uuidMap)` contains the root and all descendants and maps every original UUID to the UUID of the corresponding copied object.
* An already filled map is cleared/replaced by `copy(uuidMap)`.
* The mapping has the expected size.

The current `copyWithMapping` test can probably also be simplified quite a bit. We don't need to check all object names again if the purpose of the test is the UUID mapping.

Also please remove the debug output / unused includes and make sure the created objects are cleaned up properly.

tests/unittests/datamodel/test_copy_gtobject_with_uuidmap.h
```diff
@@ -0,0 +1,27 @@
+/* GTlab - Gas Turbine laboratory
+ *
+ * SPDX-License-Identifier: MPL-2.0+
+ * SPDX-FileCopyrightText: 2026 German Aerospace Center (DLR)
+ * Source File: test_copy_gt_object_with_uuidmap.h
+ *
+ *  Created on: 18.09.2026
```
rainman110: Same as above

src/dataprocessor/gt_object.h
```diff
@@ -193,6 +194,21 @@ class GT_DATAMODEL_EXPORT GtObject : public QObject
...
+    /**
+     * @brief copy
+     * Copies the object and its children, recording the new UUIDs.
+     * @param uuidMap Maps original UUIDs to the new UUIDs of the copied objects,
+     * including this object and all copied descendants. Existing entries are
+     * replaced. The map is empty if operation fails. Original UUIDs must be
+     * unique in the tree.
+     * @return a copy of the object.
+     *
+     * Use copy->getObjectByUuid(uuidMap.value(original->uuid())) to obtain the
```
rainman110: uuidMap.value() does not exists anymore. You are currently using `newUUID`. Please change it to

`copy->getObjectByUuid(uuidMap.copiedUuid(original->uuid()))`

See comment in UUIDMap

rainman110 (changes requested):
Thanks for the quick rework. It looks much better now.

We only need a small cleanup, but conceptually, I like the implementation now.

src/dataprocessor/gt_objectuuidmap.h
MattSchuff: done

src/dataprocessor/gt_objectuuidmap.h
MattSchuff: done

tests/unittests/datamodel/test_copy_gtobject_with_uuidmap.cpp
MattSchuff: done

src/dataprocessor/gt_object.h
MattSchuff: removed

tests/unittests/datamodel/test_copy_gtobject_with_uuidmap.h
MattSchuff: done

src/dataprocessor/gt_object.h
MattSchuff: done

tests/unittests/datamodel/test_copy_gtobject_with_uuidmap.cpp
MattSchuff: > I would remove the tests that mainly test the test setup itself (`testFillingAndClearingOfTestGroup`, `testAppendingToTestGroup`) and keep the tests focused on the actual copy behavior.

I like testing my own test structure, but I am also happy to remove it here ;-) 

>     * `copy()` still creates new UUIDs for the object and its children.

already available `copyWithoutMapping`

>     * `copy(uuidMap)` contains the root and all descendants and maps every original UUID to the UUID of the corresponding copied object.

already available `copyWithMapping` 

>     * An already filled map is cleared/replaced by `copy(uuidMap)`.

added

>     * The mapping has the expected size.

added

> The current `copyWithMapping` test can probably also be simplified quite a bit. We don't need to check all object names again if the purpose of the test is the UUID mapping.

cleaned up 

> Also please remove the debug output / unused includes and make sure the created objects are cleaned up properly.

Objects in the test group should now be cleaned up by `TearDown()` of the `TestCopyGtObjectWithUuidMapping` and adding `delete objgroup`.
Added `delete newobj` to end of test functions

src/dataprocessor/gt_object.h
MattSchuff: done

src/dataprocessor/gt_object.cpp
MattSchuff: Yes, was an edit on the side project that I found makes sense. Happy to revert.

MattSchuff:
@sourcery-ai review

tests/unittests/datamodel/test_copy_gtobject_with_uuidmap.cpp
```diff
@@ -0,0 +1,164 @@
...
+        ASSERT_TRUE(uuidMap->containsOriginalUuid(orig->uuid()));
+        ASSERT_EQ(uuidMap->copiedUuid(orig->uuid()), copied->uuid() );
+    }
+
+    auto childrenOrig = orig->findDirectChildren();
+    auto childrenCopied = copied->findDirectChildren();
+
+    for (int i=0; i<childrenOrig.size(); i++)
+    {
+        checkObjectsUuidChange( childrenOrig.at(i), childrenCopied.at(i));
```
rainman110, line 122: The `uuidMap` gets lost in the recursion here. So `copyWithMapping` currently only checks the mapping for the root object, not for the children. Just pass `uuidMap` through here as well.

tests/unittests/datamodel/test_copy_gtobject_with_uuidmap.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+#include "gtest/gtest.h"
+
+#include "gt_objectfactory.h"
+#include "gt_objectgroup.h"
+#include "gt_objectuuidmap.h"
+
+#include "test_copy_gtobject_with_uuidmap.h"
+
+#include <QDebug>
```
rainman110, line 16: `QDebug` / `printObjects()` are still left over but unused. Can go.

rainman110 (changes requested):
Looks good overall now. Just two small things left from my side. Also please update the PR description, it still describes the old QHash/pointer API and the old Memento/IO changes.

src/dataprocessor/gt_objectio.h
```diff
@@ -133,6 +137,20 @@ class GtObjectIO
     /// Pointer to current object factory
```
rainman110, line 137: Please forward declare `GtObjectUUIDMap` here instead of relying on the transitive include through `GtObjectMemento` / `GtObject`.

rainman110 (commented):
One last cleanup thing: please also update the PR description. It still talks about a `QHash*` API and public Memento/IO overloads, which is no longer how this is implemented.

rainman110:
One more small thing: I think the new public API should also get a short note in the developer docs.

`docs/dev/basics/data_modelling.rst` looks like the right place. Maybe add a small `Copying Objects` section around the Objects/Ownership part, before Properties.

It doesn't need to be much, just explain that `GtObject::copy()` creates new UUIDs, `clone()` preserves them, and that `copy(GtObjectUUIDMap&)` can be used if the caller needs the original -> copied UUID mapping for the complete copied hierarchy. A tiny usage example with `copiedUuid()` would be useful.

rainman110:
@MattSchuff Can you please also rebase on the current master?  I changed the way cppcheck works. Hopefully, we don't see any issue then here anymore.

rainman110:
> @MattSchuff Can you please also rebase on the current master? I changed the way cppcheck works. Hopefully, we don't see any issue then here anymore.

@MattSchuff  I've just done that for you. Make sure, that you  update you local branches accordingly

src/dataprocessor/gt_object.cpp
```diff
@@ -181,26 +182,27 @@ GtObject::revertDiff(GtObjectMementoDiff& diff)
     return GtObjectIO::revertDiff(diff, this);
 }
 
-namespace {
-GtObject* copyCloneHelper(const GtObject* toCopy,
-                          GtAbstractObjectFactory* fac,
-                          bool clone)
+GtObject*
+GtObject::copyClone(bool clone, GtObjectUUIDMap* uuidMap) const
```
rainman110, line 186: According to cppcheck, the argument `clone` shadows the `clone()` method, so please rename the arg in the cpp file (only!).

tests/unittests/datamodel/test_copy_gtobject_with_uuidmap.cpp
MattSchuff: Thanks for spotting that!

tests/unittests/datamodel/test_copy_gtobject_with_uuidmap.cpp
MattSchuff: removed

src/dataprocessor/gt_objectio.h
MattSchuff: good catch, done

src/dataprocessor/gt_object.cpp
MattSchuff: renamed to cloneObject

MattSchuff:
> One more small thing: I think the new public API should also get a short note in the developer docs.
> 
> `docs/dev/basics/data_modelling.rst` looks like the right place. Maybe add a small `Copying Objects` section around the Objects/Ownership part, before Properties.
> 
> It doesn't need to be much, just explain that `GtObject::copy()` creates new UUIDs, `clone()` preserves them, and that `copy(GtObjectUUIDMap&)` can be used if the caller needs the original -> copied UUID mapping for the complete copied hierarchy. A tiny usage example with `copiedUuid()` would be useful.

done

MattSchuff:
> ## [Codecov](https://app.codecov.io/gh/dlr-gtlab/gtlab-core/pull/1560?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=dlr-gtlab) Report
> 
> ❌ Patch coverage is `94.11765%` with `2 lines` in your changes missing coverage. Please review.
> [Files with missing lines](https://app.codecov.io/gh/dlr-gtlab/gtlab-core/pull/1560?dropdown=coverage&src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=dlr-gtlab) 	Patch % 	Lines
> [src/dataprocessor/gt_object.cpp](https://app.codecov.io/gh/dlr-gtlab/gtlab-core/pull/1560?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=dlr-gtlab#diff-c3JjL2RhdGFwcm9jZXNzb3IvZ3Rfb2JqZWN0LmNwcA==) 	93.33% 	[1 Missing ⚠️ ](https://app.codecov.io/gh/dlr-gtlab/gtlab-core/pull/1560?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=dlr-gtlab)
> [src/dataprocessor/gt_objectio.cpp](https://app.codecov.io/gh/dlr-gtlab/gtlab-core/pull/1560?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=dlr-gtlab#diff-c3JjL2RhdGFwcm9jZXNzb3IvZ3Rfb2JqZWN0aW8uY3Bw) 	85.71% 	[1 Missing ⚠️ ](https://app.codecov.io/gh/dlr-gtlab/gtlab-core/pull/1560?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=dlr-gtlab)
> 
> 📢 Thoughts on this report? [Let us know!](https://github.com/codecov/feedback/issues/255)

The missing code is theoretically unreachable.
`GtObjectIO::toMementoImpl` has a check if the argument `const GtObject* o` is null.
This is correct to not run into segfault here:
https://github.com/dlr-gtlab/gtlab-core/pull/1560/changes#diff-c252b1051464dec9c80e58c1206de3a0948d723a4e221e7ccdc30d9f18392cf4R237-R240
But in my test setup I cannot achieve sending a nullptr to `GtObjectIO::toMementoImpl` (but the check for a nullptr is still valid here IMHO), as the only user is within GtObject::copyClone where the usage is `objectIO.toMementoImpl(this, ....)`
Although it is currently not possible, future implementations might run into segfaulting here.

The other missing line is the uuidMap.clear() when the self-copying of GtObject fails in `GtObject::copy(GtObjectUUIDMap&)`. However, in current implementation the self-copying never fails (or?).

rainman110:
Don't worry. A patch coverage of 95% is awesome. Often, the error paths are hard or nearly impossible to trigger.

Good job 👍

rainman110 (approved):
Thanks for the work. Looks good to me now

rainman110:
One note: I have rebased the branch to gain a linear history. You need to update your stacked branch when this is merged to master.

